PR #4, opened 2019-09-05
Update fluentd config for new qserv logging format

andy-slac:
This has to be deployed when PDAC starts running new qserv version, coordination with qserv people is needed.


cbanek:
This looks great, let me know  when you want me to roll it out and I'll merge it then and then deploy it

PR #77, opened 2020-01-27
Add auth application

brianv0:
Requires lsst-sqre/charts#41 before merging


services/authnz/values-int.yaml
```diff
@@ -0,0 +1,86 @@
+secrets:
```
cbanek: Since the authnz chart is technically a subchart (with the name authnz that you gave it in the requirements.yaml), you'll want to take this whole file, and put it under an authnz block, like:

```
authnz:
    ... (all the stuff)
```

Otherwise it will get confused and not read your values files

cbanek: Same for stable

science-platform-int/templates/authnz-application.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: auth-int
```
rra: Can we make this auth instead? That will require a redeploy in int, but it gets rid of a configuration difference.

science-platform-int/templates/authnz-application.yaml
```diff
@@ -0,0 +1,27 @@
...
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: tap
+  namespace: argocd
+  finalizers:
+  - resources-finalizer.argocd.argoproj.io
+spec:
+  destination:
+    namespace: auth-int
```
rra: Likewise, auth here.

services/authnz/values-int.yaml
```diff
@@ -0,0 +1,87 @@
...
+  oauth2_proxy_client_id: "cilogon:/client_id/6ca7b54ac075b65bccb9c885f9ba4a75"
+  
+  # Default capability required for user-based operations (like generating a token)
+  user_capability: "exec:user"
+  
+  # Default capability required for admin-based operations (like revoking a token, possibly)
+  admin_capability: "exec:admin"
+  
+  oauth2_proxy:
+    image: "lsstdm/oauth2_proxy:stable"
```
rra: Should use a version tag, not `stable`, so that it doesn't automatically upgrade.

services/authnz/values-int.yaml
```diff
@@ -0,0 +1,87 @@
...
+    # URLs for oauth2_proxy. This can be simpler, but we need all of them so we can have a
+    # special one for login_url - to include the LSST skin.
+    issuer_url: https://cilogon.org
+    login_url: https://cilogon.org/authorize/?skin=LSST
+    jwks_url: https://cilogon.org/oauth2/certs
+    redeem_url: https://cilogon.org/oauth2/token
+    issuer_key_ids: ["244B235F6B28E34108D101EAC7362C4E"]
+  
+  jwt_authorizer:
+    image: "lsstdm/jwt_authorizer:latest"
```
rra: Should use a version tag, not `latest`, so that it doesn't automatically upgrade.

services/authnz/values-stable.yaml
```diff
@@ -0,0 +1,87 @@
...
+  oauth2_proxy_client_id: "cilogon:/client_id/7ae419868b97e81644ced9886ffbcec"
+  
+  # Default capability required for user-based operations (like generating a token)
+  user_capability: "exec:user"
+  
+  # Default capability required for admin-based operations (like revoking a token, possibly)
+  admin_capability: "exec:admin"
+  
+  oauth2_proxy:
+    image: "lsstdm/oauth2_proxy:stable"
```
rra: Same comment here about version tag.

services/authnz/values-stable.yaml
```diff
@@ -0,0 +1,87 @@
...
+    # URLs for oauth2_proxy. This can be simpler, but we need all of them so we can have a
+    # special one for login_url - to include the LSST skin.
+    issuer_url: https://cilogon.org
+    login_url: https://cilogon.org/authorize/?skin=LSST
+    jwks_url: https://cilogon.org/oauth2/certs
+    redeem_url: https://cilogon.org/oauth2/token
+    issuer_key_ids: ["244B235F6B28E34108D101EAC7362C4E"]
+  
+  jwt_authorizer:
+    image: "lsstdm/jwt_authorizer:latest"
```
rra: Same comment here about version tag.

science-platform-int/templates/authnz-application.yaml
```diff
@@ -0,0 +1,27 @@
...
+metadata:
+  name: auth-int
+spec:
+  finalizers:
+    - kubernetes
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: tap
```
cbanek: This should be auth not tap

services/authnz/values-int.yaml
brianv0: I will tag and update

brianv0: I've tagged the release as 0.1. I've also created a branch to publish commits to master to docker hub. This is more important for work on -int probably.

For oauth2_proxy, our code was based off of the 3.2 release of the pusher branch, so I've tagged it as 3.2-lsst1. Depending on what kind of response we get on our PR (e.g. if we have to maintain a fork) we should work to update a release and tag policy for the package.

PR #96, opened 2020-02-26
[DM-23653] Create the vault-secrets-operator namespace

cbanek:
We need to make sure this namespace exists, and since this is a
standalone app, the namespace should be here with the rest of the
chart.


afausti:
Nice, I'm doing the same at ``lsst-sqre/argocd-efd/blob/master/apps/vault-secrets-operator/templates/vault-secrets-operator-ns.yaml``

PR #125, opened 2020-04-08
Tickets/dm 24312

athornton:
Pin jh_config to older version to match JH pin


services/nublado/values-stable.yaml
```diff
@@ -108,3 +108,52 @@ nublado:
     I79d2OwyjAqpf3SJKiBOpimK+1Vh/ZcKIPBVT+INfoe9XPcKnFV9csmpFXfgR5gP
     TwIDAQAB
     -----END PUBLIC KEY-----
+
+  jupyterhub_config: |
+    '''Runtime configuration for JupyterHub in the LSST environment.
+    '''
```
cbanek, line 114: Ah okay yes this makes sense.  We just don't use the newer imports here.  All good.

PR #126, opened 2020-04-13
Add LSP-internal postgres

athornton:


services/postgres/requirements.yaml
```diff
@@ -0,0 +1,4 @@
+dependencies:
+- name: postgres
+  version: ">=0.0.1"
```
rra, line 3: I've seen this done two ways: pinning the version and requiring a change to bump it (what I've been doing), and using this syntax so that new Helm charts are picked up automatically. Should we standardize on one approach or the other?

athornton: Until I have it working, I want to let it float.  Once it's working we should definitely pin it for stable, anyway.

PR #127, opened 2020-04-13
[DM-23776] Protect Nublado with Gafaelfawr

rra:
Deploy Gafaelfawr instead of authnz on nublado.lsst.codes and
switch to protecting nublado with it.  Deploy from the PR branch
for now while code review is in progress.

### DIFF
--- a/science-platform/templates/gafaelfawr-application.yaml
+++ b/science-platform/templates/gafaelfawr-application.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: auth
+  name: gafaelfawr
 spec:
   finalizers:
     - kubernetes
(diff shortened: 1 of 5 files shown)


science-platform/templates/gafaelfawr-application.yaml
```diff
@@ -1,27 +1,27 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: auth
+  name: gafaelfawr
```
cbanek, line 4: I'm already excited to misspell this every time until I learn to spell this new wonderful word.

services/gafaelfawr/values-nublado.yaml
```diff
@@ -0,0 +1,55 @@
...
+      read:tap/efd: Execute SELECT queries in the TAP interface on EFD datasets
+      read:tap/user: Execute SELECT queries in the TAP interface on your data
+      write:tap/user: Upload tables to your database workspace
+      read:tap/history: Read the history of your TAP queries.
+      read:workspace: Read project datasets from the file workspace
+      read:workspace/user: Read the data in your file workspace
+      write:workspace/user: Write data to your file workspace
+  
+    # Mapping of capabilities to groups that are part of the `isMemberOf` claim
+    group_mapping:
```
cbanek, line 46: are these the same as in the values.yaml?  if so I'd probably just leave it (same for capabilities) - otherwise I'd assume there's something different and start diffing to find it

rra: Oh, yes, good point. I've made this file much smaller.

PR #133, opened 2020-04-16
Add configuration for NCSA Teststand environment

athornton:


rra:
You need to add either gafaelfawr or authnz for authentication. Currently the configuration you have assumes authnz, so you'll want to include that chart and a new values file for the authnz service directory. (I can then move it to gafaelfawr later.)

athornton:
Think this is the necessary fields.

PR #142, opened 2020-04-24
Add Kueyen, parameterize create.sh, add helm-create to initialize helm.

athornton:


argo-cd/create.sh
```diff
@@ -1 +1,11 @@
...
+VALUES=$1
+if [ -z "${VALUES}" ]; then
+    VALUES="argo-cd"
+fi
+vfile="${VALUES}-values.yaml"
+if ! [ -e "${vfile}" ]; then
+    echo "No values file ${vfile}!" 1>&2
+    exit 2
+fi
+helm install argo/argo-cd --namespace argocd --values ${vfile} --name argocd
```
cbanek, line 11: Can we not add the needlessly complicated mechanism to pick a filename?

argo-cd/helm-create.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+helm init
+kubectl create namespace argocd
+kubectl create serviceaccount tiller --namespace kube-system
+kubectl create -f tiller-clusterrolebinding.yaml
+helm init --service-account tiller --upgrade
+./create.sh $1
```
cbanek, line 7: I didn't put these together because helm init frequently times out or takes time to start.  This script exists in the develop_gke folder as install_tiller.sh and the rbac file is in the develop_gke/tiller folder and takes that in.

argo-cd/kueyen-values.yaml
```diff
@@ -0,0 +1,32 @@
+redis:
```
cbanek, line 1: Is kueyen the name of the environment?  This is a word I've never seen

argo-cd/tiller-clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: tiller-clusterrolebinding
+subjects:
+- kind: ServiceAccount
+  name: tiller
+  namespace: kube-system
+roleRef:
```
cbanek, line 9: This is already checked in, don't double it up

cbanek:
Don't get too fancy trying to make all this perfect, I'm working on better bootstrapping, but pretty much all of this other than the values file already exists in this repo

argo-cd/helm-create.sh
cbanek: Also if you're installing argocd with helm, you shouldn't need to make a namespace here

argo-cd/create.sh
athornton: But this way you can just run `create.sh` with an argument like `kueyen` to select the values file, and if you just run it with no arguments, it behaves just like it used to.

athornton:
OK, I'll back everything but the values file out, but you need to install tiller and pals in things that aren't a GKE environment too.  The script to do so should not be hidden over in develop-gke

argo-cd/kueyen-values.yaml
athornton: Yes.  I don't know what it means either.

PR #148, opened 2020-04-27
[DM-24595] Convert int to Gafaelfawr

rra:
Replace authnz with Gafaelfawr on int.  Update the nginx ingress
annotations to point to the new URLs and remove the 403 handling
since we don't want to redirect those to the login handler.

### DIFF
--- a/services/tap/values-int.yaml
+++ b/services/tap/values-int.yaml
@@ -14,10 +14,11 @@ cadc-tap:
 
   ingress:
     authenticated_annotations:
+      nginx.ingress.kubernetes.io/auth-request-redirect: $request_uri
       nginx.ingress.kubernetes.io/auth-response-headers: X-Auth-Request-Uid, X-Auth-Request-Token
-      nginx.ingress.kubernetes.io/auth-url: https://lsst-lsp-int.ncsa.illinois.edu/auth?capability=read:tap&reissue_token=true
+      nginx.ingress.kubernetes.io/auth-signin: "https://lsst-lsp-int.ncsa.illinois.edu/login"
+      nginx.ingress.kubernetes.io/auth-url: "https://lsst-lsp-int.ncsa.illinois.edu/auth?scope=read:tap"
       nginx.ingress.kubernetes.io/configuration-snippet: |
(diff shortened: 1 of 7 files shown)


services/tap/values-int.yaml
```diff
@@ -14,10 +14,11 @@ cadc-tap:
 
   ingress:
     authenticated_annotations:
+      nginx.ingress.kubernetes.io/auth-request-redirect: $request_uri
       nginx.ingress.kubernetes.io/auth-response-headers: X-Auth-Request-Uid, X-Auth-Request-Token
-      nginx.ingress.kubernetes.io/auth-url: https://lsst-lsp-int.ncsa.illinois.edu/auth?capability=read:tap&reissue_token=true
+      nginx.ingress.kubernetes.io/auth-signin: "https://lsst-lsp-int.ncsa.illinois.edu/login"
+      nginx.ingress.kubernetes.io/auth-url: "https://lsst-lsp-int.ncsa.illinois.edu/auth?scope=read:tap"
       nginx.ingress.kubernetes.io/configuration-snippet: |
-        error_page 403 = @Basic401; # Basic401 configured in jwt-authorizer ingress rule
```
cbanek, line 20: what happened to this part?  We had problems without this line I think because the auth proxy would send me a redirect, which isn't part of the TAP flow.  Is this going to send me a redirect or is this just something that has been deprecated?

rra: Gafaelfawr standardizes on using 401 if the person isn't logged in and 403 if they don't have access. oauth2_proxy was ambiguous about this and sometimes used 403 if they weren't logged in, so both had to have the same behavior. That's undesirable since you end up in redirect loops if the user can authenticate but legitimately doesn't have access (which returns a 403). So now we let ingress-nginx redirect to the signin URL on 401, but on 403 just return the 403 straight to the user.

Does that make sense for TAP?

cbanek: That sounds good - one of the things was the 403 was actually a redirect back to login, which TAP couldn't tell apart from an actual result to the query, which can also issue a redirect.

So 401 and 403 is good, redirects bad.  Maybe this was also part of the fixing the redirect loops.  I think we're good here :)

PR #158, opened 2020-04-29
[DM-24597] Add JWT public key for int and NCSA test stand

rra:
The chart has the value for int, but duplicate it in the int values
file since I otherwise keep forgetting that it needs to be set for
all other environments.


rra:
Stable's already set -- this isn't a new requirement with Gafaelfawr, so you added it for stable back when we added authnz.

PR #170, opened 2020-05-04
[DM-24662] Set default Gafaelfawr IdP to NCSA

rra:
This worked well on nublado.lsst.codes.  The IdP selection screen
is still shown, but NCSA is selected as the default.  Update the
other Gafaelfawr environments to have the same configuration.


rra:
We've decided not to do this for the time being because there's no way to just set the default, only constrain the list of IdPs. We may ask for this as a feature from CILogon in the future.

PR #187, opened 2020-05-13
Grant base access to the lsst-sqre friends team

rra:


rra:
It uses what GitHub calls the "slug" rather than the name, which is the canonicalized version and I believe is forced to lowercase.

PR #199, opened 2020-06-01
Fix values file naming for new layout

rra:
For nublado.lsst.codes, some services were using values-nublado
and some services were using values-gke.  The new templating
standardizes on values-gke.  Rename the other values files
accordingly.


cbanek:
I'm honestly not sure - there is also a nublado environment and a gke environment which was just a general google environment.  Seems like maybe we want both?  Or maybe we should deprecate the gke environment for bleed?  Let's talk at standup tomorrow on this one

rra:
Oh, I didn't see there was a separate nublado environment. I proposed this because nublado.lsst.codes Argo CD is showing out of date, and if I updated it it would break because it would look for `values-gke.yaml` files in places that don't have them (at least if I'm reading the output correctly).

I was assuming the goal was to have nublado.lsst.codes and bleed have the same configuration, but now that I think about it, maybe that doesn't work because the client identity for GitHub auth isn't the same. So maybe instead we need more values files?

rra:
This was fixed a different way.

PR #203, opened 2020-06-03
[DM-25185] Fix yaml files

cbanek:
I thought helm would expand out these keys with the dots, like it
does on the command line.  Apparently it will just silently set
keys with dots.


cbanek:
Third time lucky?

PR #209, opened 2020-06-11
Update dependencies

rra:
- Update vault-secrets-operator Helm chart from 1.4.0 to 1.6.0
- Update argo Helm chart from 0.6.5 to 0.9.5


rra:
We're going to do this with Renovate, which will generate separate PRs for each dependency update. The argo update is riskier.

PR #237, opened 2020-06-25
[DM-25487] Manage nginx-ingress on Tucson teststand

rra:
Take over management of nginx-ingress on the Tucson teststand,
and switch from the wildcard certificate to the Let's Encrypt
certificate.


athornton:
LGTM...although Josh doesn't want us using the default cert at all, but one step at a time.

PR #241, opened 2020-06-27
[DM-25703] Add documentation framework

rra:
Add a very basic operational documentation framework based on
Roundtable.  Add operational documentation for vault-secrets-operator.


rra:
My inclination would be to avoid the "lsp-deploy" name in the URL as well because I'm about 95% sure that we'll end up needing to rename it at some point. Also, it feels to me like keeping the operational documentation here is mostly a matter of convenience, but it's not limited to just the things in the lsp-deploy repository. It's operational documentation for the entirety of the Science Platform (which probably but not definitely will be the Rubin Science Platform at some point).

Not sure what that implies for URLs. science-platform-ops.lsst.io is probably the least ambiguous, but might be a bit long.

jonathansick:
Ok, sounds good. I like the `science-platform-ops.lsst.io` URL. Maybe we can push to rename the lsp-deploy repo sooner than later to keep things consistent?

rra:
Yeah, let's talk about that in stand-up today. I think that's a good idea.

PR #249, opened 2020-06-30
[DM-25412] Script to generate secrets and push to vault

cbanek:
generate_secrets.py will ask for the values and generate random
values as well as generate RSA keys.  These are written into a
directory called secrets, which push_secrets.sh will then use
to write all the keys and their values to vault.  You must
run the push_secrets.sh script with the VAULT_TOKEN appropriately
set for the environment you are trying to setup.

### DIFF
--- a/installer/generate_secrets.py
+++ b/installer/generate_secrets.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+from base64 import b64encode
+from cryptography.fernet import Fernet
+from cryptography.hazmat.primitives.asymmetric import rsa
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import serialization
+import json
+import os
+import secrets
+
+
+def input_field(component, name, description):
+  prompt_string = f"[{component} {name}] ({description}): "
+  return input(prompt_string)
+
+def generate_tap_secrets():
+  fname = input_field("TAP", "google_creds.json",
+    "file containing google service account credentials")
+
+  with open(fname, 'r') as f:
+    return {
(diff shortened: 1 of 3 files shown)


installer/generate_secrets.py
```diff
@@ -0,0 +1,76 @@
...
+
+def generate_tap_secrets():
+  fname = input_field("TAP", "google_creds.json",
+    "file containing google service account credentials")
+
+  with open(fname, 'r') as f:
+    return {
+      "slack_webhook_url": input_field("TAP", "slack_webhook_url",
+        "slack webhook url for querymonkey"),
+      "google_creds.json": b64encode(f.read().encode("utf-8")).decode("utf-8")
```
rra, line 23: Nit: `utf-8` will always be the default so you can just use `.encode()` and `.decode()`.

installer/generate_secrets.py
```diff
@@ -0,0 +1,76 @@
...
+  key_bytes = key.private_bytes(
+    serialization.Encoding.PEM,
+    serialization.PrivateFormat.PKCS8,
+    serialization.NoEncryption()
+  )
+
+  return {
+    "cilogon-client-secret": input_field("gafaelfawr", "cilogon-client-secret",
+                                         "CILogon client secret"),
+    "session-secret": "HqgO-VmEqNFZrbzWNOfzFb3wWeAnxV3tTlBY1b_n_1c=",
```
rra, line 55: We shouldn't use the same secret for every environment.

```
from cryptography.fernet import Fernet
# ....
    "session-secret": Fernet.generate_key().encode(),
```

PR #263, opened 2020-07-07
Update version of Gafaelfawr

rra:
Picks up the change to include exec:admin in the token list.


rra:
Merging the Renovate version of this instead.

PR #288, opened 2020-07-23
[DM-26078] Add Redis password secret, bump Gafaelfawr on bleed

rra:
Add a Redis password secret for Gafaelfawr to the installer script.
Move bleed to a ticket branch so that I can test Redis password
authentication.

### DIFF
--- a/services/gafaelfawr/values-bleed.yaml
+++ b/services/gafaelfawr/values-bleed.yaml
@@ -1,5 +1,8 @@
 gafaelfawr:
   host: bleed.lsst.codes
+  image:
+    tag: "tickets-DM-26078"
+    pullPolicy: "Always"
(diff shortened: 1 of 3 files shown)


services/gafaelfawr/values-bleed.yaml
```diff
@@ -1,5 +1,8 @@
 gafaelfawr:
   host: bleed.lsst.codes
+  image:
+    tag: "tickets-DM-26078"
```
cbanek, line 4: So this is just for testing this version on bleed, and will go back to master later?

rra: Yes, indeed. As soon as the next Gafaelfawr is released (probably with the OpenID Connect support I'm working on now).

PR #343, opened 2020-09-09
[DM-25963] Add SQuaSH applications to lsp-deploy

afausti:
This PR adds the "squash-sandbox" environment to lsp-deploy enabling the following apps under the science-platform parent app

- nginx-ingress
- cert-issuer
- cert-manager
- vault-secrets-operator
- gafaelfawr
- squash-api
- chronograf
- influxdb
- kapacitor 

we took the opportunity to test the Chronograf, InfluxDB and Kapacitor Helm charts maintained by InfluxData and configure ingress rules to use a path base routing (`/squash-api`, `/chronograf` and `/influxdb`). We enable Gafaelfawr to authenticate Chronograf and to issue tokens for InfluxDB.
 
We tested that by deploying the above apps using Argo CD and this branch. Chronograf can be reached for example at https://squash-sandbox.lsst.codes/chronograf

### DIFF
--- a/services/chronograf/values-squash-sandbox.yaml
+++ b/services/chronograf/values-squash-sandbox.yaml
@@ -72,6 +72,7 @@ chronograf:
   env:
     HOST_PAGE_DISABLED: true
     BASE_PATH: /chronograf
+    GENERIC_NAME: "CILogon"
     GENERIC_AUTH_URL: https://squash-sandbox.lsst.codes/auth/openid/login
     GENERIC_TOKEN_URL: https://squash-sandbox.lsst.codes/auth/openid/token
     USE_ID_TOKEN: 1
--- a/services/chronograf/values-squash-sandbox.yaml
+++ b/services/chronograf/values-squash-sandbox.yaml
@@ -66,8 +66,7 @@ chronograf:
     hostname: squash-sandbox.lsst.codes
     annotations:
       kubernetes.io/ingress.class: "nginx"
-      # kubernetes.io/tls-acme: "true"
-    path: "/chronograf"
+    path: "/chronograf/"
 
   ## OAuth Settings for OAuth Providers
   ## More information -> https://github.com/influxdata/chronograf/blob/master/docs/auth.md
@@ -93,6 +92,7 @@ chronograf:
   ## Extra environment variables that will be passed onto deployment pods
   env:
     HOST_PAGE_DISABLED: true
+    BASE_PATH: /chronograf
 
   ## The name of a secret in the same kubernetes namespace which contain values to be added to the environment
   ## This can be useful for auth tokens, etc
--- a/services/chronograf/values-squash-sandbox.yaml
+++ b/services/chronograf/values-squash-sandbox.yaml
@@ -61,8 +61,7 @@ chronograf:
   ##
   ingress:
     enabled: true
-    tls: true
-    secretName: tls-certs
+    tls: false
     hostname: squash-sandbox.lsst.codes
     annotations:
       kubernetes.io/ingress.class: "nginx"
--- a/services/chronograf/values-squash-sandbox.yaml
+++ b/services/chronograf/values-squash-sandbox.yaml
@@ -68,32 +68,21 @@ chronograf:
       kubernetes.io/ingress.class: "nginx"
     path: "/chronograf/"
 
-  ## OAuth Settings for OAuth Providers
-  ## More information -> https://github.com/influxdata/chronograf/blob/master/docs/auth.md
-  ##
-  oauth:
-    # Need to set to true to use any of the oauth options
-    enabled: false
-    generic:
-      enabled: false
-      # client_id: CHANGE_ME
-      # client_secret: CHANGE_ME
-      api_key: ""
-      scopes: ""
-      # eg. http://chronograf.foobar.com
-      public_url: ""
-      auth_url: ""
-      token_url: ""
-      api_url: ""
-      # optional
-      # name: "generic"
-      # domains: ""
-
   ## Extra environment variables that will be passed onto deployment pods
   env:
     HOST_PAGE_DISABLED: true
     BASE_PATH: /chronograf
+    GENERIC_AUTH_URL: https://squash-sandbox.lsst.codes/auth/openid/login
+    GENERIC_TOKEN_URL: https://squash-sandbox.lsst.codes/auth/openid/token
+    USE_ID_TOKEN: 1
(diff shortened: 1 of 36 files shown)


services/chronograf/values-squash-sandbox.yaml
```diff
@@ -0,0 +1,88 @@
+chronograf:
+  ## Image Settings
+  ##
+  image:
+    repository: "chronograf"
+    tag: 1.8.6
+    pullPolicy: IfNotPresent
```
rra, line 7: Does upstream release a new version of the Helm chart for each version of the software? If so, it may be easier to use their defaults to pick up new versions. This way, we have to remember to upgrade the underlying application as well as the Helm chart, and Renovate may warn us when there's a new application available. (It can sometimes figure this out but not always.)

services/chronograf/values-squash-sandbox.yaml
```diff
@@ -0,0 +1,88 @@
...
+  affinity: {}
+
+  ## Configure the ingress object to hook into existing infastructure
+  ## ref : http://kubernetes.io/docs/user-guide/ingress/
+  ## OPTIONALLY you can set .Values.ingress.secretName to set which secret to use
+  ##
+  ingress:
+    enabled: true
+    tls: true
+    secretName: tls-certs
```
rra, line 65: I think you need the cluster-issuer configuration here as well.

In practice, this is a little odd since multiple ingresses that share the same hostname can't have separate TLS certificates, so one of these is going to win and the rest will lose. What we have currently done for lsp-deploy is make nublado win and not configure TLS for anything else. You don't have nublado in this cluster so that doesn't work. I think this works since all ingresses that define TLS obtain essentially identical certs, but be aware that one of them will win (based on age of the resource) and the rest will be ignored.

services/influxdb/values-squash-sandbox.yaml
```diff
@@ -0,0 +1,257 @@
+influxdb:
+  ## influxdb image version
+  ## ref: https://hub.docker.com/r/library/influxdb/tags/
+  image:
+    repository: "influxdb"
+    tag: "1.8.2-alpine"
+    pullPolicy: IfNotPresent
```
rra, line 7: Same question here about whether the upstream Helm chart will handle this for us.

services/kapacitor/values-squash-sandbox.yaml
```diff
@@ -0,0 +1,73 @@
+kapacitor:
+  ## influxdb image version
+  ## ref: https://hub.docker.com/r/library/influxdb/tags/
+  ##
+  image:
+    repository: "kapacitor"
+    tag: "1.5.6-alpine"
```
rra, line 7: Same question here about whether the upstream Helm chart will handle this for us.

services/squash-api/values-squash-sandbox.yaml
```diff
@@ -0,0 +1,91 @@
+squash-api:
+  replicaCount: 1
+
+  image:
+    repository: lsstsqre/squash-restful-api
+    tag: latest
```
rra, line 6: This is a little scary in the long run since it means uploading a new Docker image will cause the API to be upgraded when it is restarted (which could occur for various reasons automatically and unpredictably in the middle of the night). I think it would be better in the long run to pin the version.

services/squash-api/values-squash-sandbox.yaml
```diff
@@ -0,0 +1,91 @@
...
+
+  service:
+    type: ClusterIP
+    port: 8080
+
+  ingress:
+    enabled: true
+    annotations:
+      kubernetes.io/ingress.class: "nginx"
+      nginx.ingress.kubernetes.io/rewrite-target: /
```
rra, line 72: Does this annotation do anything? I wonder if you could drop it for simplicity.

rra (approved):
Just a few nits and questions. Overall looks good to me.

Could you add the new apps to all the other values files in the science-platform directory and explicitly set `enabled` to `false` as well, just for clarity? We've been following the convention of listing all the apps in each of the values files.

services/chronograf/values-squash-sandbox.yaml
afausti: Unfortunately they don't, the InfluxData Helm charts are behind.  Right now they have  Chronograf 1.8.0, InfluxDB 1.8.0-alpine and Kapacitor 1.5.4-alpine.

services/influxdb/values-squash-sandbox.yaml
afausti: See above.

services/kapacitor/values-squash-sandbox.yaml
afausti: See above.

services/squash-api/values-squash-sandbox.yaml
afausti: Agreed. I didn't touch the SQuaSH API code base for a while but I'll soon change this.  In fact I'll take your previous suggestion and remove this from the yaml and use the default version set in the chart, which is `latest` right now but will change soon.

services/squash-api/values-squash-sandbox.yaml
afausti: I couldn't reach the Flask app without it, I think I need it because I have the path base route to `/squash-api` and the app expects `/` I had a similar issue with the [InfluxDB ingress configuration](https://github.com/lsst-sqre/lsp-deploy/pull/343/commits/ae8cbe67cbdf4436aefb2d336a126a6219eecfaf).

services/chronograf/values-squash-sandbox.yaml
afausti: Good catch, let me review that I think I can set `tls: false` since I'm using path base routing now.

afausti: That's right, I can use`tls: false` here.

afausti: I'm amazed. I've disable tls for `influxdb` and the `squash-api` as well. I don't need the `tls-certs` secret in each app namespace anymore, things are much simpler with path base routing.

afausti: Ok, too early to say that. I just confirmed what you said. I need at least one service creating the `tls-certs` even if it is created on the app namespace. Since all services share the same base hostname this works. So I think I'll choose to configure tls in the `squash-api`.

afausti:
Adrressed review comments and updated brach. I think I'm ready to merge these changes.

PR #346, opened 2020-09-10
[DM-25487] Revert "Split infrastructure services into new parent app"

rra:
This reverts commit a2f9e921f9b83d0a8248289c2783594c51a81f7a.

The creation of the new app is not safe because syncing `science-platform` will delete these apps.
Revert to the scheme that seemed to mostly be working except for a failure of nublado to sync that
we haven't yet tracked down.


cbanek (approved):
LGTM!

PR #359, opened 2020-09-23
update hub/wfdispatcher on int

athornton:


cbanek (approved):
This should be fine, but you might want to rebuild prepuller and increment that just in case?

PR #360, opened 2020-09-24
prepuller is fine on nublado, updating int

athornton:


athornton:
going to self-merge.  It ran fine overnight at nublado.lsst.codes.

PR #367, opened 2020-09-28
Enable workflow-ui and add ingress annotations

athornton:


services/nublado/values-nublado.yaml
```diff
@@ -4,11 +4,7 @@ nublado:
   oauth_provider: 'jwt'
 
   hub:
-    image: 'lsstsqre/sciplat-hub:latest'
-
-  wf:
-    no_verify_signature: 'true'
-    no_verify_audience: 'true'
+    image: 'lsstsqre/sciplat-hub:exp_0.32.1'
```
cbanek, line 7: Do you want to check this line in?

services/wf/Chart.yaml
```diff
@@ -1,3 +1,3 @@
 apiVersion: v1
 name: wf
-version: 0.0.1
+version: 0.0.2
```
cbanek, line 3: I don't think you need to increment this version

cbanek (approved):

services/wf/Chart.yaml
athornton: I do want to check that line in, but it's going to change soon anyway.  I'll deincrement Chart.yaml.

PR #368, opened 2020-09-28
bump hub version on nublado

athornton:


athornton:
Imma self-merge--I'm only changing the version of hub on nublado.

PR #379, opened 2020-10-05
replicaCount 2 for int

athornton:


SimonKrughoff (commented):
Do we want to go to the released firefly and test it before we up the replica count?

SimonKrughoff (approved):
Adam reminds me that we don't have a new image for the released firefly image, so we may as well just go ahead and do this

PR #381, opened 2020-10-10
[DM-27044] Add minikube environment

cbanek:
Add an environment that uses minikube for helping development.

### DIFF
--- a/services/gafaelfawr/values-minikube.yaml
+++ b/services/gafaelfawr/values-minikube.yaml
@@ -14,7 +14,7 @@ gafaelfawr:
 
   # Use CILogon authentication.
   cilogon:
-    client_id: "cilogon:/client_id/4fb3456c6d99009d975450af19bee94"
+    client_id: "cilogon:/client_id/74e865cd71a3a327096d36081166b739"
     redirect_url: "https://minikube.lsst.codes:31337/oauth2/callback"
     login_params:
       skin: "LSST"
--- a/services/gafaelfawr/values-minikube.yaml
+++ b/services/gafaelfawr/values-minikube.yaml
@@ -15,6 +15,6 @@ gafaelfawr:
   # Use CILogon authentication.
   cilogon:
     client_id: "cilogon:/client_id/74e865cd71a3a327096d36081166b739"
-    redirect_url: "https://minikube.lsst.codes:31337/oauth2/callback"
+    redirect_url: "https://minikube.lsst.codes:31337/login"
     login_params:
       skin: "LSST"
--- a/services/gafaelfawr/values-minikube.yaml
+++ b/services/gafaelfawr/values-minikube.yaml
@@ -8,10 +8,6 @@ gafaelfawr:
   issuer:
     exp_minutes: 43200  # 30 days
 
-  # Disabled but kept so that the client ID is easily accessible.
-  # github:
-  #   client_id: a19e79298a352f3e5650
-
   # Use CILogon authentication.
   cilogon:
     client_id: "cilogon:/client_id/74e865cd71a3a327096d36081166b739"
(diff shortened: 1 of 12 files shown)


services/gafaelfawr/values-minikube.yaml
```diff
@@ -0,0 +1,20 @@
...
+  vault_secrets_path: "secret/k8s_operator/minikube.lsst.codes/gafaelfawr"
+
+  # Session length and token expiration (in minutes).
+  issuer:
+    exp_minutes: 43200  # 30 days
+
+  # Disabled but kept so that the client ID is easily accessible.
+  # github:
+  #   client_id: a19e79298a352f3e5650
+
```
rra, line 14: Should delete this -- this was specific to whatever environment the file was cloned from.

services/nublado/values-minikube.yaml
```diff
@@ -0,0 +1,99 @@
+nublado:
```
rra, line 1: I'm confused -- this seems to be a configuration for the existing nublado, not for your new work, but we won't ever run the existing nublado in this environment, will we?

services/nublado/values-minikube.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+  mountpoints: |
+    [
+      { "disabled": true,
+        "mountpoint": "/example -- if it doesn't start with '/' we add it",
+        "fileserver-host": "omit to use Hub settings",
+        "fileserver-export": "defaults to '/exports<mountpoint>'",
+        "mode": "rw or ro, defaults to 'ro'",
+        "options": "mount options, e.g. 'local_lock=all'"
+      },
```
rra, line 75: If we're keeping this file, this should be a comment.

services/nublado/values-minikube.yaml
cbanek: Ah this is a good point, but then I remember the TLS for cert manager is somehow knotted around nublado I think due to the cert being put in that namespace.  So we needed this I think to get the cert to be requested.

services/nublado/values-minikube.yaml
cbanek: Sure I can do that, it's just in everywhere, because that's how it was done

rra: Yeah, understood this is just copy and paste. I think we can avoid configuring nublado 1 for the new environment at all, hopefully.

cbanek: Yeah, I tried to set up the fileserver and the networking to get the NFS mounting to work last week, but it was a hellish doomscape of problems.  The kubernetes host didn't originally know the DNS of private services, so I fixed that (apparently k8s has a long standing policy of not touching host's resolv.conf), and still I couldn't get the NFS services to mount.  Empty config works though, which is nice so that other people who still want to run nublado 1 (I'm thinking bootcamp and other people until we get nublado 2 done) can spin up a dev instance.  It might also help Adam with his dev since you don't have to push docker images across the internet before you test them.

services/gafaelfawr/values-minikube.yaml
athornton: Well, yes and no.  The _idea_ of keeping the client ID around in the file is not a bad one, as long as we don't keep the secret in the file.  But it should definitely be *our* client ID if we keep it.

services/nublado/values-minikube.yaml
athornton: That would be nice.  The internal fileserver has very likely bitrotted.

athornton (approved):
Minor comments, no issues.

services/nublado/values-minikube.yaml
cbanek: Okay just made this an empty list and that should be fine.

PR #398, opened 2020-10-29
Add pull user to deployments

athornton:
I've already updated vault with the password in all environments.


r-owen (approved):
Looks OK to me (though I am no expert on these files).

PR #401, opened 2020-11-04
Add kueyen environment

athornton:
It's a copy of base, with "base-lsp" replaced by "kueyen-lsp" and the comcam mount disabled.


afausti (approved):
LGTM

PR #406, opened 2020-11-09
DM-27474: Add exposure log entries for postgres

r-owen:


athornton (changes requested):
I'd prefer it if you'd only add these at the instances where we're going to need them, because if the password secret is missing but the other two are present, postgres won't start, and I don't want to put an unused secret in a bunch of environments in vault.

r-owen:
OK. I've pruned it as per your suggestion on Slack and adding Minikube and bleed as per Angelo's suggestion.

athornton (approved):

PR #407, opened 2020-11-09
Don't default the science-platform part to an environment name

cbanek:
This might be confusing, as Adam will likely attest to.


athornton (approved):
Looks wonderful to me.

PR #411, opened 2020-11-12
[DM-26507] Teach Argo CD about itself

rra:
- Add an app for Argo CD itself in the hope that it can upgrade itself properly
- Remove old scripts and Kubernetes resources made obsolete by the new installer
- Rename the Argo Workflows application configuration to use our standard naming scheme


cbanek (approved):
Seems reasonable

PR #420, opened 2020-11-17
[DM-27636] Fix nginx-ingress chart dependency

cbanek:
Okay so the previous chart was marked as deprecated and moved to
a new helm chart repo.  So first we update the requirements.yaml
to point to that repository, and the new URL.  The version numbers
have also changed, so pick the latest one there that is published.

Now since the name of the chart has changed from nginx-ingress to
ingress-nginx, this means that the name in the requirements.yaml
changes as well.  And since this is used as the section in the
values files, all of those have to change to match that name.
Otherwise that configuration section will be ignored (which is
all of it).

### DIFF
--- a/installer/argo-cd-values.yaml
+++ b/installer/argo-cd-values.yaml
@@ -20,3 +20,5 @@ server:
         name: lsst-sqre
       - url: https://ricoberger.github.io/helm-charts/
         name: ricoberger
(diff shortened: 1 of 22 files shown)


athornton (approved):
LGTM

rra (approved):
LGTM, hopefully everything is backwards-compatible

PR #424, opened 2020-11-19
Enable comcam mount at summit

SimonKrughoff:


athornton (approved):
If that's the export path, LGTM

PR #430, opened 2020-11-21
[DM-27441] Add cachemachine and nublado2 to science-platform

cbanek:
Also adding values for bleed and minikube, which are the only
enabled environments.

### DIFF
--- a/services/nublado2/values-bleed.yaml
+++ b/services/nublado2/values-bleed.yaml
@@ -1,0 +1,49 @@
+nublado2:
+  jupyterhub:
+    debug:
+      enabled: true
+
+    hub:
+      image:
+        name: lsstsqre/nublado2
+        tag: "1.0.0"
+
+    singleuser:
+      storage:
+        type: none
+
+    auth:
+      state:
+        cryptoKey: "080c6d0eb8d9f84ef5f0860bea71dba6ab2e1be712c7bc8c2eaf92321c9fd968"
+
+    proxy:
+      secretToken: "3854245f5b48412ee9b7a2c985da6acdc146f80cc4f10eb175f0efa24530399b"
+      service:
+        type: ClusterIP
+
+    ingress:
+      hosts: ["bleed.lsst.codes"]
+      annotations:
+        nginx.ingress.kubernetes.io/auth-signin: "https://bleed.lsst.codes:31337/login"
+        nginx.ingress.kubernetes.io/auth-url: "http://gafaelfawr-service.gafaelfawr.svc.cluster.local:8080/auth?scope=exec:notebook"
+
+  config:
+    options_form:
+      images_url: "http://cachemachine.cachemachine.svc.cluster.local/cachemachine/jupyter/available"
+      images:
+        - name: Weekly 42
+          image: registry.hub.docker.com/lsstsqre/sciplat-lab:w_2020_42
+      sizes:
+        - name: Tiny
+          cpu: .5
+          ram: 1536M
+        - name: Small
+          cpu: 1
+          ram: 3072M
+        - name: Medium
+          cpu: 2
+          ram: 6144M
+        - name: Large
+          cpu: 4
+          ram: 12288M
+    user_resources: {}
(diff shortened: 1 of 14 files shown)


services/nublado2/values-bleed.yaml
```diff
@@ -0,0 +1,49 @@
...
+    singleuser:
+      storage:
+        type: none
+
+    auth:
+      state:
+        cryptoKey: "080c6d0eb8d9f84ef5f0860bea71dba6ab2e1be712c7bc8c2eaf92321c9fd968"
+
+    proxy:
+      secretToken: "3854245f5b48412ee9b7a2c985da6acdc146f80cc4f10eb175f0efa24530399b"
```
rra, line 20: These are secrets, which is okay as a temporary stopgap for minikube, but bleed is on the Internet. I think they need to be real secrets before we add them to non-minikube environments.

services/nublado2/values-bleed.yaml
```diff
@@ -0,0 +1,49 @@
...
+      service:
+        type: ClusterIP
+
+    ingress:
+      hosts: ["bleed.lsst.codes"]
+      annotations:
+        nginx.ingress.kubernetes.io/auth-signin: "https://bleed.lsst.codes:31337/login"
+        nginx.ingress.kubernetes.io/auth-url: "http://gafaelfawr-service.gafaelfawr.svc.cluster.local:8080/auth?scope=exec:notebook"
```
rra, line 28: The `auth-url` must be the external URL, not the internal one.

services/nublado2/values-bleed.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+    proxy:
+      secretToken: "3854245f5b48412ee9b7a2c985da6acdc146f80cc4f10eb175f0efa24530399b"
+      service:
+        type: ClusterIP
+
+    ingress:
+      hosts: ["bleed.lsst.codes"]
+      annotations:
+        nginx.ingress.kubernetes.io/auth-signin: "https://bleed.lsst.codes:31337/login"
```
rra, line 27: No port on bleed.

services/nublado2/values-bleed.yaml
```diff
@@ -0,0 +1,49 @@
...
+        nginx.ingress.kubernetes.io/auth-url: "http://gafaelfawr-service.gafaelfawr.svc.cluster.local:8080/auth?scope=exec:notebook"
+
+  config:
+    options_form:
+      images_url: "http://cachemachine.cachemachine.svc.cluster.local/cachemachine/jupyter/available"
```
rra, line 32: This should be in the chart rather than lsp-deploy, I think, because it's not going to vary by environment.

As I learned the hard way with Gafaelfawr, the Chilean clusters don't use `.svc.cluster.local` as their domain name. Omitting that seems to work.

services/nublado2/values-bleed.yaml
rra: Ack, ignore this, this is fine. I had completely forgotten our previous conversation about this and that you'd tested it and it works fine. (In fact, I probably should stop exposing that route to the Internet since there's no reason for people to use it directly.) Same note as below about `.svc.cluster.local`, though.

services/nublado2/values-bleed.yaml
cbanek: UGH!  That's really annoying.  Thanks for mentioning this.  You'll probably mention it a few more times before I forget.  Will do on adding it to the chart.

services/nublado2/values-bleed.yaml
cbanek: Thanks!

services/nublado2/values-bleed.yaml
cbanek: Ah yes that's right, I was working on adding cachemachine secrets generation to the generate secrets.py, but I couldn't finish that before I left, and was tempted to just do it without that and add it later, but I'll try to add that in.  My problem got around ingested a dockerconfigjson file as a secret... it seems to work fine with the vault command line, but putting it in a json document doesn't seem to parse... I might have to work on that one for a bit.

cbanek:
Okay this one should be ready for a second look.

rra (approved):

PR #450, opened 2020-12-01
more frequent prepulling on int/stable/nublado; remove unused key

athornton:


services/nublado/values-int.yaml
```diff
@@ -21,6 +21,9 @@ nublado:
     persistent_home: false
     image: 'lsstsqre/sciplat-hub:0.32.5'
 
+  prepuller:
+    schedule: "27 01,07,13,19 * * *"
```
SimonKrughoff, line 25: Looks like this needs to go in the other prepuller block later

services/nublado/values-stable.yaml
```diff
@@ -22,6 +22,9 @@ nublado:
     persistent_home: false
     image: 'lsstsqre/sciplat-hub:0.32.5'
 
+  prepuller:
+    schedule: "57 01,07,13,19 * * *"
```
SimonKrughoff, line 26: Same as in the int values file

SimonKrughoff (approved):
Every six hours seems fine to me, obviously the linting needs to pass

PR #451, opened 2020-12-02
Fix exposurelog values for the NTS environment

afausti:
- Set correct values for nfs_path_1, butler_uri_1 and nfs_server at NCSA Test Stand


r-owen (approved):
This looks great. Thank you for figuring out how to mount that volume.

PR #452, opened 2020-12-02
Add IDF Dev environment

athornton:


services/nublado/values-idfdev.yaml
```diff
@@ -0,0 +1,103 @@
...
+        "mountpoint": "/example -- if it doesn't start with '/' we add it",
+        "fileserver-host": "omit to use Hub settings",
+        "fileserver-export": "defaults to '/exports<mountpoint>'",
+        "mode": "rw or ro, defaults to 'ro'",
+        "options": "mount options, e.g. 'local_lock=all'"
+      },
+      {
+       "mountpoint": "/home",
+        "mode": "rw",
+        "fileserver-host": "FIXME",
```
afausti: Do you have a value for the fileserver-host?

afausti (commented):
Could not spot any problem, only the values for  `fileserver-host` which are still FIXME

services/nublado/values-idfdev.yaml
athornton: I think you're looking at only the first commit.

10.210.11.194

PR #463, opened 2020-12-07
Set values file when installing vault-secrets-operator

afausti:
- Now that vault-secrets-operator is parametrized for each environment the install.sh script needs to set the --values option.


rra (approved):
Good catch!

PR #465, opened 2020-12-08
[DM-27441] science-platform: add in nublado2

cbanek:
Forgot this when doing the last PR, and that caused a syncing error
since the revision couldn't be sourced.


athornton (commented):
LGTM

SimonKrughoff (approved):
Looks good

PR #472, opened 2020-12-09
Seems to be working.

athornton:


tools/multipathsecret/multipathsecret/vaultconfig.py
```diff
@@ -0,0 +1,125 @@
...
+    rendered-to-memory JSON document representing all the vault paths
+    and tokens, and the list of vault paths to not update.
+
+    '''
+    vault_address: str = os.getenv('VAULT_ADDR')
+    secret: Optional[Dict[str, str]] = None
+    enclaves: List[Enclave] = field(default_factory=list)
+    skip_list: List[str] = field(default_factory=list)
+
+    def __init__(self,
```
rra, line 41: Dataclasses shouldn't have an `__init__` method since this breaks the dataclass magic. If you need to run initialization code in a dataclass, use `__post_init__` (or just make this a regular class). In this case, this doesn't look very dataclass-y since you have a whole host of methods defined, so this should probably just be a regular class.

tools/multipathsecret/multipathsecret/standalone.py
```diff
@@ -0,0 +1,38 @@
+import click
+from .vaultconfig import VaultConfig
+
+
+@click.group()
+@click.option("--vault-address", "-a")
+@click.option("--secret-name", "-n", required=True)
+@click.option("--vault-file", "-v", required=True)
+@click.option("--omit", "-o", multiple=True)
+@click.option("--dry-run", "-x", is_flag=True)
```
rra, line 10: Documentation strings for these that explain what the parameters are for would be very helpful.

tools/multipathsecret/multipathsecret/standalone.py
```diff
@@ -0,0 +1,38 @@
...
+    ctx.obj['vault_config'] = VaultConfig(vault_address=vault_address,
+                                          vault_file=vault_file,
+                                          skip_list=omit)
+    ctx.obj['options'] = {'secret_name': secret_name,
+                          'dry_run': dry_run}
+
+
+@cli.command()
+@click.pass_context
+@click.option("--secret-file", "-s", required=True)
```
rra, line 23: Same here for documentation (including what the syntax of this file should be).

tools/multipathsecret/multipathsecret/vaultconfig.py
```diff
@@ -0,0 +1,125 @@
...
+            enclave = Enclave(name=name, read=read_k, write=write_k)
+            self.enclaves.append(enclave)
+
+    def load_secret(self, secret_file: str) -> None:
+        with open(secret_file, "r") as f:
+            self.secret = json.load(f)
+
+    def _get_write_key_for_enclave(self, enclave: Enclave) -> str:
+        '''Given a Vault path, return its write key.
+        '''
```
rra, line 68: Nit: This function takes an Enclave with the keys already loaded, not a Vault path.

tools/multipathsecret/multipathsecret/vaultconfig.py
```diff
@@ -0,0 +1,125 @@
...
+    '''Contains the vault address (a URL represented as a string), the
+    secret to be added, which is a dict mapping strings to strings
+    (but can be None, if you're deleting a secret from a path), the
+    rendered-to-memory JSON document representing all the vault paths
+    and tokens, and the list of vault paths to not update.
+
+    '''
+    vault_address: str = os.getenv('VAULT_ADDR')
+    secret: Optional[Dict[str, str]] = None
+    enclaves: List[Enclave] = field(default_factory=list)
```
rra, line 38: Nit: Given how it's used, I think this is crying out to be a `Dict[str, Enclave]` mapping vault paths to their enclaves.

tools/multipathsecret/multipathsecret/vaultconfig.py
```diff
@@ -0,0 +1,125 @@
...
+                   dry_run: bool = False) -> None:
+        self._change_secret(verb="add", enclave=enclave,
+                            secret_name=secret_name, dry_run=dry_run)
+
+    def remove_secret(self, enclave: Enclave, secret_name: str,
+                      dry_run: bool = False) -> None:
+        self._change_secret(verb="remove", enclave=enclave,
+                            secret_name=secret_name, dry_run=dry_run)
+
+    def _change_secret(self, verb: str, enclave: Enclave, secret_name: str,
```
rra, line 104: Nit: Use an enum for verb and you can avoid writing the code to check that the argument is valid.

athornton: Done

tools/multipathsecret/multipathsecret/vaultconfig.py
athornton: Done

tools/multipathsecret/multipathsecret/standalone.py
athornton: Done

tools/multipathsecret/multipathsecret/standalone.py
athornton: Done

tools/multipathsecret/multipathsecret/vaultconfig.py
athornton: Done

athornton:
I am closing this one and moving the tool under lsstvaultutils.

PR #474, opened 2020-12-09
DM-27844-1: exposurelog: fix nfs_path_1 for summit

r-owen:


r-owen:
Thanks to @jhoblitt-sqre for the correct path. I tested it by editing the manifest directly (that's what is currently running).

afausti (approved):
Glad it worked at the summit too.

PR #489, opened 2020-12-15
Add pull secrets across nublado.lsst.codes

athornton:


rra (commented):
Could you pin the version of pull-secret everywhere? I want to avoid more places where we float all of our environments as soon as we merge a commit to charts.

This change doesn't seem to do anything. It adds a pull secret, but nothing uses that secret. Is that coming in a subsequent diff?

The expansion of the values files seems fine. I think there are ways to avoid that and would like to clean this up at some point, but I don't think that point needs to be now.

athornton:
I will pin the versions.  This creates the secrets but nothing uses them.  That will be a future PR.

rra (approved):

PR #492, opened 2020-12-16
[DM-28043] Fix up minikube networking

cbanek:
Okay, so first off, this came about by not being able to get to the
gafaelfawr token endpoint, due to the annotations being wrong (no
port), and no way to configure it.

Realizing I'm digging a hole rather than making a pile, I found a
way to configure the nginx-ingress to use the host networking on
80 and 443 (which is actually the VM's host).  This means we can
get rid of port 31337 (picked because NodePorts can only have
ephemeral range).

Using the host network also allows the pod to contact itself,
something it couldn't do previously since it wasn't attached to
the host networking stack.

One downside is that I haven't gotten the admissions webhook
service working yet, and this means that you won't get a warning
if you make an editing goof in your ingress yaml.  But only for
minikube.  I think this can be fixed later, but I'm not sure how,
it has to do with binding that to a nodeport or something.

We will also need to notify CILogon to change the redirect URL
to not redirect to 31337 but instead just 443.

### DIFF
--- a/services/nublado/values-minikube.yaml
+++ b/services/nublado/values-minikube.yaml
@@ -11,7 +11,7 @@ nublado:
       annotations:
         nginx.ingress.kubernetes.io/auth-method: GET
         nginx.ingress.kubernetes.io/auth-response-headers: X-Auth-Request-Token
-        nginx.ingress.kubernetes.io/auth-signin: "https://minikube.lsst.codes:31337/login"
+        nginx.ingress.kubernetes.io/auth-signin: "https://minikube.lsst.codes/login"
         nginx.ingress.kubernetes.io/auth-url: "http://gafaelfawr-service.gafaelfawr.svc.cluster.local:8080/auth?scope=exec:notebook"
         nginx.ingress.kubernetes.io/configuration-snippet: |
           proxy_set_header X-Forwarded-Proto https;
@@ -52,7 +52,7 @@ nublado:
       annotations:
         nginx.ingress.kubernetes.io/auth-request-redirect: $request_uri
         nginx.ingress.kubernetes.io/auth-response-headers: X-Auth-Request-Token
-        nginx.ingress.kubernetes.io/auth-signin: "https://minikube.lsst.codes:31337/login"
+        nginx.ingress.kubernetes.io/auth-signin: "https://minikube.lsst.codes/login"
         nginx.ingress.kubernetes.io/auth-url: "http://gafaelfawr-service.gafaelfawr.svc.cluster.local:8080/auth?scope=exec:notebook"
         nginx.ingress.kubernetes.io/configuration-snippet: |
           proxy_set_header X-Forwarded-Proto https;
--- a/installer/minikube/coredns.yaml
+++ b/installer/minikube/coredns.yaml
@@ -10,7 +10,6 @@ data:
         health {
            lameduck 5s
         }
-        rewrite name minikube.lsst.codes nginx-ingress-controller.nginx-ingress.svc.cluster.local
         ready
         kubernetes cluster.local in-addr.arpa ip6.arpa {
            pods insecure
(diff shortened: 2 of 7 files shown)


services/nublado/values-minikube.yaml
```diff
@@ -11,8 +11,8 @@ nublado:
       annotations:
         nginx.ingress.kubernetes.io/auth-method: GET
         nginx.ingress.kubernetes.io/auth-response-headers: X-Auth-Request-Token
-        nginx.ingress.kubernetes.io/auth-signin: "https://minikube.lsst.codes:31337/login"
-        nginx.ingress.kubernetes.io/auth-url: "http://gafaelfawr-service.gafaelfawr.svc.cluster.local:8080/auth?scope=exec:notebook"
+        nginx.ingress.kubernetes.io/auth-signin: "https://minikube.lsst.codes/login"
+        nginx.ingress.kubernetes.io/auth-url: "https://minikube.lsst.codes/auth?scope=exec:notebook"
```
rra, line 15: I think you should leave the auth-url pointing to the cluster service. I want to go change all the rest of them to match; there's no need for that request to go through the ingress. (Same note in several other places.)

rra (approved):
Thanks for doing this! I love removing that divergence.

services/nublado/values-minikube.yaml
cbanek: Okay that seems fine - one thing though is that I'll need to make a PR to the gafaelfawr chart as there's no way to configure it like that with the values files.  Or should I just change the default in the chart to point to the service?  Are we ready for that?

rra: I don't think I understand. I'm only advocating not changing gafaelfawr-service.gafaelfawr in auth-url back to minikube.lsst.codes in this diff. No changes to the chart should be required, or changing any files you're not already touching. I will clean up the rest of the annotations elsewhere and make them consistent in another diff.

auth-signin does have to stay minikube.lsst.codes.

cbanek: https://github.com/lsst-sqre/charts/blob/7d167bed2d2aed177d351c708c5e11c2ae3aaee8/charts/gafaelfawr/templates/ingress-token.yaml#L9

This is where I had some trouble, as there's no way to set it to gafaelfawr-service.gafaelfawr here (as you said, the signin url should still be minikube.lsst.codes) but they are generated from the same thing.  So I think we'll need some configuration or that line should be pointing to gafaelfawr-service

rra: I agree that the current Gafaelfawr chart won't work if minikube is configured to use a non-standard port. That will continue to be the case (because of auth-signin) even when I make a sweep and change the other auth-url references. Apologies for not having thought of that earlier; that's a good catch.

I'm lost as to what that has to do with leaving auth-url set to gafaelfawr-service here, though. The auth-uri you have now works and I'm fairly sure wasn't part of your problem, and I eventually want to change every other auth-uri to match (including in the Gafaelfawr chart). The auth-uri is only used by the nginx-ingress service, which is running inside the cluster, so it doesn't need to use the ingress name. That said, it doesn't really matter; if you want to change it to minikube here, I can change it back to gafaelfawr-service later along with everything else.

cbanek: Sorry my first problem that precipitated this was actually not being able to get to the gafaelfawr token endpoint (and not able to configure it to be so) on minikube.  Totally fine leaving all the auth-url as they are.  But then to get to the token endpoint on minikube we'll need to change the gafaelfawr chart.  I can make a separate PR for the chart or you can handle it when you do your auth-url sweep

rra: I think we're just talking past each other. :) Please go ahead and merge this PR as is. It will work fine. I'll handle auth-url separately later.

PR #495, opened 2020-12-16
Add pull-secret to parent chart.

athornton:


rra (approved):

rra:
You may want to use `pull_secret` rather than `pull-secret` for the values parameter.

PR #497, opened 2020-12-17
[DM-28050] Installer improvements

cbanek:
Use helm3.
Install argocd using the services directory version to eliminate a double sync.
Clean up the shell script.
Make a CI job.
Get rid of the URL of the repo in the deployment (call git in the installer to set it).  This will let us rename this repo more easily, as well as enables forks to be fully qualified test and deploy instances of themselves.
Flatten out revision now that people can fork.  Have a whole environment look at a specific branch to keep things a little more simpler mentally.


cbanek:
The CI run didn't seem to work from a fork although it worked in that fork, secrets are the same on both sides.

PR #498, opened 2020-12-17
[DM-28050] Installer improvements

cbanek:
Use helm3.
Install argocd using the services directory version to eliminate a double sync.
Clean up the shell script.
Make a CI job.
Get rid of the URL of the repo in the deployment (call git in the installer to set it). This will let us rename this repo more easily, as well as enables forks to be fully qualified test and deploy instances of themselves.
Flatten out revision now that people can fork. Have a whole environment look at a specific branch to keep things a little more simpler mentally.


afausti (commented):
Great work on simplifying and making the installer more consistent. I've made a small suggestion on using the `helm dependency update` command instead of `helm dependency build`. I've created a new cluster on GKE and I'm trying to run the new installer. For some reason it get's stuck when syncing the science-platform apps, I'm testing the installer with the `squash-sandbox` environment. I'll reach to you so we can inspect the cluster and perhaps learn from that.

installer/install.sh
```diff
@@ -69,15 +69,12 @@ argocd app sync science-platform \
   --port-forward-namespace argocd
 
 echo "Syncing critical early applications"
-argocd app sync nginx-ingress \
-  --port-forward \
-  --port-forward-namespace argocd || true
```
afausti, line 84: I see that you removed the `nginx-ingress` from the critical early applications here.  That's getting the installer  stuck when syncing the science platform apps. If I cancel the installer, sync the the `nginx-ingress` app manually and run the installer again it goes on and finishes the installation.

afausti (commented):
I'll start over from a fresh cluster and try to confirm that `nginx-ingress` must be in the list of critical early applications.

installer/install.sh
```diff
@@ -1,42 +1,40 @@
...
+kubectl create ns vault-secrets-operator || true
 kubectl create secret generic vault-secrets-operator \
   --namespace vault-secrets-operator \
   --from-literal=VAULT_TOKEN=$VAULT_TOKEN \
   --from-literal=VAULT_TOKEN_LEASE_DURATION=31536000 \
   --dry-run -o yaml | kubectl apply -f -
 
+echo "Update / install vault-secrets-operator..."
+# ArgoCD depends on pull-secret, which depends on vault-secrets-operator.
+helm dependency build ../services/vault-secrets-operator
```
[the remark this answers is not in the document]
cbanek: Sounds good thanks!  Done!

installer/install.sh
```diff
@@ -1,42 +1,40 @@
...
+helm upgrade vault-secrets-operator ../services/vault-secrets-operator \
+  --install \
+  --values ../services/vault-secrets-operator/values-$ENVIRONMENT.yaml \
+  --create-namespace \
+  --namespace vault-secrets-operator \
+  --timeout 15m \
+  --wait
+
+echo "Update / install argocd using helm3..."
+helm dependency build ../services/argocd
```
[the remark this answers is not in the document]
cbanek: Updated!  Thanks!

installer/install.sh
cbanek: Ah fantastic, okay, I was wondering that.  I thought it was tied to the cert manager stuff, and since using DNS it shouldn't need that anymore.  I just ran on GKE the rogue-two setup and that doesn't hang - but that doesn't include chronograph and some of the other components too.  Maybe those require some ingress to do something.  I'm going to try to deploy that one (but not mess with the DNS) and see if I can repro, and then add this back and see if it works.

Thanks so much!

cbanek: Okay added this back in.

afausti (approved):
Awesome! just did another installation on a fresh cluster and it finished without issues.

PR #500, opened 2020-12-17
Add pull secrets across environments

athornton:


rra (commented):
This blindly adds pull-secret to a bunch of charts that don't and will never support it, such as Argo CD (for which you should set `global.imagePullSecrets` instead). I'd rather we not do that, because I think it's going to be quite confusing when we look at it later and have forgotten the history. Can you instead add this only to the charts where we've plumbed this setting through and it does something?

athornton:
closing in favor of new branch; got a little wound round the axle there.

PR #508, opened 2020-12-21
DM-28135: exposurelog: add site_id and disable nts deployment

r-owen:
Add site_id to all existing exposurelog values (bleed, nts, and summit).
Disable deployment at nts because it doesn't have a suitable butler registry.


SimonKrughoff (approved):
Looks good

PR #509, opened 2020-12-22
Add idfint

athornton:


athornton:
Didn't sanitize it on disk correctly first.

PR #513, opened 2020-12-23
Try to fix the stable repo problem (and add science-platform-values f…

athornton:
…or int, which I missed)

### DIFF
--- a/science-platform/values-idfint.yaml
+++ b/science-platform/values-idfint.yaml
@@ -1,0 +1,40 @@
+environment: idfint
+
+argo:
+  enabled: true
+cert_issuer:
+  enabled: true
+cert_manager:
+  enabled: true
+chronograf:
+  enabled: false
+exposurelog:
+  enabled: false
+gafaelfawr:
+  enabled: true
+influxdb:
+  enabled: false
+kapacitor:
+  enabled: false
+landing_page:
+  enabled: true
+logging:
+  enabled: false
+mobu:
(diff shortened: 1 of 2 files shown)


science-platform/values-idfint.yaml
```diff
@@ -0,0 +1,40 @@
...
+logging:
+  enabled: false
+mobu:
+  enabled: true
+nginx_ingress:
+  enabled: true
+nublado:
+  enabled: true
+obstap:
+  enabled: true
```
SimonKrughoff, line 30: Not sure we want this true right now.  Are we sure it's working?

science-platform/values-idfint.yaml
```diff
@@ -0,0 +1,40 @@
...
+portal:
+  enabled: true
+postgres:
+  enabled: true
+squash_api:
+  enabled: false
+tap:
+  enabled: true
```
SimonKrughoff, line 38: Not sure about this either.  With no qserv this can't work, but maybe we just want the endpoint up so we can do a health check?  If there is actually a qserv and I'm just ill informed, ignore me.

SimonKrughoff (approved):
Just a couple of questions about services that depend on qserv

science-platform/values-idfint.yaml
athornton: I mean, maybe?  I just copied dev.  I presume that there will eventually be qserv in each?

PR #517, opened 2021-01-04
[DM-28198] Portal: use new way to specify hostpath

cbanek:
This is a slightly different way to configure it based on the
helm chart changes on the same ticket.


athornton (approved):
LGTM

PR #519, opened 2021-01-05
[DM-27979] Rename nginx-ingress to ingress-nginx

rra:
Make the name of the chart match the upstream subchart, since
otherwise Helm adds nginx-ingress- as a prefix to the name of
every resource, resulting in obnoxiously long resource names and
some confusion.


cbanek:
That looks good to me, and while it's a tricky change, I feel better knowing that minikube did a test deploy.  Approved!

cbanek (approved):

PR #526, opened 2021-01-07
[DM-28256] Vault TLS certificate loading

cbanek:
Load the certificate from a vault secret.

Disable for all environments other than minikube for now, although
forks of this repo may use this feature.  This means that instead
of using let's encrypt, minikube will use the certificate stored
in vault, not stressing our rate limits.


athornton (approved):
LGTM

rra (approved):
Looks good to me too. That was simpler than I feared it would be!

PR #533, opened 2021-01-15
[DM-27983] Flesh out cert-issuer documentation

rra:
Add documentation for setting up Route 53 for a new zone, and be
clearer about the steps for setting up cert-issuer for a new cluster.
Reference the cert-issuer documentation from cert-manager.


athornton (approved):
Yes, much clearer.

PR #539, opened 2021-01-22
[DM-28463] Enable cert-manager and ingress-nginx at base and summit

rra:
For now, we'll be managing cert-manager and ingress-nginx in the
antu and andes clusters.  Enable them and add the required values
files for ingress-nginx.

Try to pin the IP addresses for base, summit, and tucson-teststand
since we hopefully should be able to request the same IP address as
we're currently using in those clusters.

### DIFF
--- a/services/ingress-nginx/values-base.yaml
+++ b/services/ingress-nginx/values-base.yaml
@@ -9,11 +9,13 @@ ingress-nginx:
       use-forwarded-headers: "true"
     service:
       externalTrafficPolicy: Local
-      loadBalancerIP: "140.252.34.101"
+      loadBalancerIP: "139.229.146.150"
+    podLabels:
+      hub.jupyter.org/network-access-proxy-http: "true"
 
 vault_certificate:
   enabled: false
 
 pull-secret:
   enabled: true
-  path: secret/k8s_operator/tucson-teststand.lsst.codes/pull-secret
+  path: secret/k8s_operator/base-lsp.lsst.codes/pull-secret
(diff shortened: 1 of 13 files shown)


services/ingress-nginx/values-base.yaml
```diff
@@ -0,0 +1,21 @@
...
+      loadBalancerIP: "139.229.146.150"
+    podLabels:
+      hub.jupyter.org/network-access-proxy-http: "true"
+
+vault_certificate:
+  enabled: false
+
+pull-secret:
+  enabled: true
+  path: secret/k8s_operator/bleed.lsst.codes/pull-secret
```
cbanek, line 21: bleed / base?

services/ingress-nginx/values-summit.yaml
```diff
@@ -0,0 +1,21 @@
...
+      loadBalancerIP: "139.229.160.150"
+    podLabels:
+      hub.jupyter.org/network-access-proxy-http: "true"
+
+vault_certificate:
+  enabled: false
+
+pull-secret:
+  enabled: true
+  path: secret/k8s_operator/bleed.lsst.codes/pull-secret
```
cbanek, line 21: bleed / summit

cbanek (changes requested):

services/ingress-nginx/values-base.yaml
rra: Oh, good, catch, thank you! Fixed both of those.

athornton (approved):
Good catch by @cbanek ; lgtm now

cbanek (approved):

PR #564, opened 2021-02-05
[DM-28675] Add a start of an architecture overview

rra:
Document the very basics about the architecture of lsp-deploy, and
add some more detailed documentation about versioning and the
layering of Helm charts.


cbanek (approved):
LGTM - thanks!

PR #584, opened 2021-02-19
[DM-28839] Refactors to generate secrets, 1password, and argocd password

cbanek:


installer/generate_secrets.py
```diff
@@ -1,38 +1,50 @@
...
+    fname = input(prompt_string)
+
+    if fname:
+        with open(fname, "r") as f:
+            s[component][name] = f.read()
+
 
+def set_generated_secret(s, component, name, new_value):
+    regenerate = True
+    if regenerate:
```
athornton, line 40: Do you want to pass regenerate as an argument to the method?  If not, why bother with the if?

installer/install.sh
```diff
@@ -1,9 +1,12 @@
-#!/bin/bash -ex
+#!/bin/bash -e
 USAGE="Usage: ./install.sh ENVIRONMENT VAULT_TOKEN"
 ENVIRONMENT=${1:?$USAGE}
-VAULT_TOKEN=${2:?$USAGE}
-GIT_URL=`git config --get remote.origin.url`
+export VAULT_TOKEN=${2:?$USAGE}
+export VAULT_ADDR=https://vault.lsst.codes
+VAULT_PATH_PREFIX=`yq -r .vault_path_prefix ../science-platform/values-$ENVIRONMENT.yaml`
+ARGOCD_PASSWORD=`vault kv get --field=argocd.admin.plaintext_password $VAULT_PATH_PREFIX/enclave`
```
athornton, line 7: Not sure I like literal "enclave"; in lsstvaultutils I used that as the term for "an autonomously managed space in vault".   Not sure what would be better.  "_secret_stash" ?

athornton (approved):
Looks fine.  One question about a thing I don't understand and one nomenclatural thing.

installer/generate_secrets.py
cbanek: Ah yes, I forgot about that, I want to pass that in as an argument to the whole program, will do

installer/install.sh
cbanek: any better idea for a name?  Maybe Russ?  I honestly don't care but secret stash sounds like a different secret stash I have.

rra: What if we just call it `installer`?

rra (approved):

.github/workflows/minikube.yaml
```diff
@@ -18,9 +18,14 @@ jobs:
         run: kubectl get nodes
       - name: Download installer dependencies
         run: |
+          curl -sSL -o /tmp/vault.zip https://releases.hashicorp.com/vault/1.6.2/vault_1.6.2_linux_amd64.zip
+          unzip /tmp/vault.zip
+          sudo mv vault /usr/local/bin/vault
```
athornton, line 23: You could do this with `install -m0755 vault /usr/local/bin` but I guess that wouldn't do the remove, so, eh, it's fine.

installer/install.sh
```diff
@@ -1,9 +1,12 @@
-#!/bin/bash -ex
+#!/bin/bash -e
```
athornton, line 1: Probably better for production use; we can always set -x explicitly before running.

installer/vault_key.py
```diff
@@ -0,0 +1,30 @@
...
+import json
+import pprint
+
+from onepassword import OnePassword
+
+
+class VaultKeyRetriever:
+    def __init__(self):
+        self.op = OnePassword()
+        vault_keys_doc = self.op.get_item(uuid="dg5afgiadsffeklfr6jykqymeu")
```
athornton, line 12: This is...just a magic constant that points to an already-extant document?  Should it be a parameter?  I feel like it should be settable (maybe as an init arg?) so that Dominique et al, as they develop their own vaults, don't have to modify the code?  Or maybe I don't understand what's going on here.

athornton (approved):
I like it.  A couple of minor questions.

installer/generate_secrets.py
```diff
@@ -1,233 +1,297 @@
...
+
+            # If we don't find anything, pull the password from a login item.
+            if not secret_value:
+                for f in doc["details"]["fields"]:
+                    if f["designation"] == "password":
+                        secret_value = f["value"]
+
+            if self.environment in environments:
+                self.op_secrets[key] = secret_value
+            elif not environments and self.op_secrets.get(key, None) == None:
```
rra, line 269: Minor nit: Prefer `is None` to `== None` in Python because some things that are not `None` can compare equal to it.

rra (approved):
Some minor notes but overall looks good to me.

installer/generate_secrets.py
rra: Looking at this again, I think you just want the last part to be `and key not in self.op_secrets`.

.github/workflows/minikube.yaml
cbanek: Yeah, kind of a bummer, maybe later on we can wrap that into a setup script that would make the virtualenv later on

installer/vault_key.py
cbanek: Sounds good, I forgot about that!  Will do that.

installer/generate_secrets.py
cbanek: yeah, you're right, I'll do that

PR #601, opened 2021-03-03
[DM-28993] Pin mobu and cachemachine charts and remove tags

rra:
Rather than pinning the Docker image tag, pin the version of the
chart for mobu and cachemachine.  The chart now pins the Docker
image.  This works correctly with Renovate's automatic PR
generation: it will now create PRs for new Docker image versions
in the charts repo and then create PRs for new chart versions in
the Phalanx repo.

### DIFF
--- a/services/cachemachine/Chart.yaml
+++ b/services/cachemachine/Chart.yaml
@@ -3,5 +3,5 @@ name: cachemachine
 version: 1.0.0
 dependencies:
 - name: cachemachine
-  version: ">=0.1.0"
+  version: 0.1.5
   repository: https://lsst-sqre.github.io/charts/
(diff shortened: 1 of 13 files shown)


cbanek (approved):
So just to be clear, this will make it so that all the different environments are running the same version, which comes from the chart values.yaml, which we will need to increment on releasing a new version.  And we increment the chart version.  We can still pin an environment to an older version though?  Just want to make sure I understand what the new flow is

rra:
It means that all environments will always run the same chart version (well, modulo deciding not to sync Argo CD or rolling back). By default, all environments will use the Docker image version configured in the charts repo, which we'll bump for each new release. Any environment can override and pin the Docker image version in its own `values.yaml` file.

cbanek:
Got it!  Sounds good to me!

PR #610, opened 2021-03-11
Add moneypenny flag and values for idfdev

athornton:


services/moneypenny/values-idfdev.yaml
```diff
@@ -0,0 +1,41 @@
+moneypenny:
+  host: "data-dev.lsst.cloud"
+
+  ingress:
+    enabled: true
+    hosts:
+      - host: data-dev.lsst.cloud
+        paths: ["/moneypenny"]
+    annotations:
+      nginx.ingress.kubernetes.io/auth-response-headers: X-Auth-Request-Uid, X-Auth-Request-Token
```
rra, line 10: I am very dubious that you're using this, given that this is synthesized data from your fake token and you're passing the UID you care about as the request body. I think you want to delete this.

services/moneypenny/values-idfdev.yaml
```diff
@@ -0,0 +1,41 @@
...
+      nginx.ingress.kubernetes.io/auth-url: "https://data-dev.lsst.cloud/auth?scope=exec:admin"
+      nginx.ingress.kubernetes.io/configuration-snippet: |
+        auth_request_set $auth_token $upstream_http_x_auth_request_token;
+        proxy_set_header Authorization "Bearer $auth_token";
```
rra, line 14: You should not need (or want) any of this. I'm hoping nothing needs any of this and that it's just a workaround for some old bug that we can drop.

athornton: Removed.

rra (approved):

PR #629, opened 2021-03-18
[DM-29170] Improve TLS documentation

rra:
Add some architectural discussion of how TLS is configured.  Stress
that only a commercial certificate or Route 53 with the DNS solver
are supported.  Be clearer about the prerequisites for the Route 53
approach.  Add a link to more information on creating glue records.


athornton (approved):
This adds a lot of clarity, I think.  I like it.

PR #636, opened 2021-03-22
Add configmaps for passwd/group/shadow/gshadow

athornton:


rra (approved):
This is okay for testing to see if this fixes the problem, but should move to the chart if it works (since presumably we'll want to do this the same way everywhere).

athornton:
I agree, but we can't do that until we merge the nublado2 moneypenny changes anyway.

PR #637, opened 2021-03-22
Add CM names

athornton:


athornton:
Self-merging: forgot I needed a name field in volume definitions

PR #638, opened 2021-03-22
Reflect running dev values in checked-in file

athornton:


athornton:
I feel like this is pretty safe to self-merge.  Not much that can go wrong here assuming the actual tagged build is good.

PR #639, opened 2021-03-24
Add cms and /tmp to all nublado2 instances

athornton:


cbanek (approved):
This looks good for now, but we should probably pull all the group and shadow stuff into the chart to make this smaller eventually.  But we can do that a bit later.

rra (approved):
Fine with me on the security side (seconding Christine's suggestion for later refactoring).

athornton:
The problem with pulling that all into the chart is that it is not a minor refactoring, because (as far as I can tell) the only place you can put this is in extraVolumes in singleuser.storage, and that's a list, which is not additive in configs.

So you'd have to pull the common parts out and then glue them into the user pod's volumes at spawn time.  This is certainly doable, but it is not trivial.

PR #644, opened 2021-03-25
[DM-29085] Add documentation of Gafaelfawr storage

rra:


athornton (approved):
This is excellent and likely something I'm going to yoink for Nublado2.

PR #652, opened 2021-03-30
Use test CILogon to verify their update.

ktlim:
CILogon has announced an update to their service on 2021-04-13.  Test it on -int prior to then by using their test service at test.cilogon.org.


athornton (approved):
LGTM

PR #653, opened 2021-04-01
DM-29432: Add squareone service

jonathansick:
This adds the squareone service and leaves it disabled everywhere except `idfint` (where it replaces the `landing-page` service).


cbanek (approved):
Looks good maybe pull default image version tag into the chart's values.yaml and only provide it where we are overriding it

PR #673, opened 2021-04-13
[DM-29647] Clean up nublado2 values files

cbanek:


athornton (approved):
The only place this gets tricky is whether the Kubespawner mounts or does not mount volumes into init containers.  I think it doesn't, and we don't want it to, but we'll deal with it either way as we need to.

PR #682, opened 2021-04-15
add summit mounts

athornton:


ktlim (approved):
Matches what I checked with Steve.  Thanks!

PR #692, opened 2021-04-21
Once more try reenabling summit mounts

athornton:


ktlim (approved):
What a saga.  Looks correct now.

PR #701, opened 2021-04-23
[DM-28120] Convert remaining environments to new stack

rra:
Convert all remaining environments to Gafaelfawr 2.0, nublado2, and
(where applicable) squareone instead of landing-page.  Remove some
stray configurations for kueyen, an environment that doesn't exist.

These nublado2 configurations for T&S clusters are not final since
they don't yet contain DDS/Multis configuration, pending discussion
on how best to do that.


rra:
This isn't ready for deployment as-is because we need to sort out DDN/Multus configuration, but it should be a good starting point. It at least hopefully copies the mount point configuration from the current nublado configs.

cbanek (approved):

PR #708, opened 2021-04-26
[DM-29799] Create nublado-users depending on nublado2 or nublado

cbanek:
Otherwise if we disable nublado, we'll stop creating this.


rra (approved):

cbanek:
Okay looks like the minikube deploy went well enough, so let's try this out on data-dev

PR #711, opened 2021-04-27
Add butler_secret_path

athornton:


cbanek:
Did we add the butler secrets to the vault repo or the generate secrets script?

athornton:
> Did we add the butler secrets to the vault repo or the generate secrets script?

Vault repo.  The Butler stuff is external to RSP, so it doesn't belong in generate secrets.

athornton:
Well crap it's also the wrong ticket number.

I'll work out the secret adding stuff and resubmit under the right one.

PR #712, opened 2021-04-27
Tickets/DM-29639B

athornton:
Add butler secret path to deploys and secret generation

### DIFF
--- a/installer/generate_secrets.py
+++ b/installer/generate_secrets.py
@@ -250,9 +250,6 @@ class SecretGenerator:
         self.input_file(
             "butler-secret", "postgres-credentials.txt", "Postgres credentials for butler"
             )
-        self.input_file(
-            "butler-secret", "butler-hmac-idf-creds.json", "Butler credentials (HMAC JSON format)"
-            )
 
     def _ingress_nginx(self):
         self.input_file("ingress-nginx", "tls.key", "Certificate private key")
--- a/installer/generate_secrets.py
+++ b/installer/generate_secrets.py
@@ -26,6 +26,7 @@ class SecretGenerator:
 
     def generate(self):
         self._pull_secret()
+        self._butler_secret()
         self._postgres()
         self._log()
         self._tap()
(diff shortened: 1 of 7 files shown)


cbanek (approved):
LGTM - Thanks for adding that secrets part, don't forget to stash that in the 1password RSP-Vault

services/nublado2/values-int.yaml
```diff
@@ -16,6 +16,7 @@ nublado2:
 
   config:
     base_url: "https://lsst-lsp-int.ncsa.illinois.edu"
+    butler_secret_path: "secret/k8s_operator/lsst-lsp-int.ncsa.illinois.edu/butler-secret"
```
rra, line 19: I don't think these credentials are meaningful for NCSA, are they? The PostgreSQL server at least wouldn't be reachable.

services/nublado2/values-minikube.yaml
```diff
@@ -11,6 +11,7 @@ nublado2:
 
   config:
     base_url: "https://minikube.lsst.codes"
+    butler_secret_path: "secret/k8s_operator/minikube.lsst.codes/butler-secret"
```
rra, line 14: Same here, I don't think there's any way to use them in minikube.

installer/generate_secrets.py
```diff
@@ -242,6 +243,17 @@ def _pull_secret(self):
...
 
+    def _butler_secret(self):
+        self.input_file(
+            "butler-secret", "aws-credentials.ini", "AWS credentials for butler"
+            )
+        self.input_file(
+            "butler-secret", "postgres-credentials.txt", "Postgres credentials for butler"
+            )
+        self.input_file(
+            "butler-secret", "butler-hmac-idf-creds.json", "Butler credentials (HMAC JSON format)"
```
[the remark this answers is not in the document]
athornton: I don't quite understand it, but Hsin-Fang knows.  I don't know that we need to support it, but it's just a reformatting of the aws-credentials.ini.  From Google's perspective it's a legacy key format for external access to the S3 bucket.  I'll drop it from the secret since we don't directly reference it and we can rederive it from the aws-credentials.ini.

services/nublado2/values-int.yaml
athornton: They are not meaningful.  However, rather than put a bunch of conditional logic in the resource manager, I'm just putting dummy secrets in all environments.

PR #730, opened 2021-05-05
Bump versions with auth fix

athornton:


athornton:
I'm just gonna merge; no reason Russ will know any better what the version numbers containing the auth fix are.

PR #782, opened 2021-05-25
DM-30300: exposurelog: change bleed deployment to base

r-owen:


services/exposurelog/values-base.yaml
```diff
@@ -1,25 +1,27 @@
...
+      - host: base.lsst.codes
         paths:
           - /exposurelog
     tls: []
     #  - secretName: chart-example-tls
     #    hosts:
     #      - chart-example.local
 
-vault_path: secret/k8s_operator/bleed.lsst.codes/postgres
+vault_path: secret/k8s_operator/base.lsst.codes/postgres
```
afausti, line 23: The correct value for the base ``vault_path`` is ``secret/k8s_operator/base-lsp.lsst.codes/postgres``

services/exposurelog/values-base.yaml
```diff
@@ -1,25 +1,27 @@
...
 
 pull-secret:
   enabled: true
-  path: secret/k8s_operator/bleed.lsst.codes/pull-secret
+  path: secret/k8s_operator/base.lsst.codes/pull-secret
```
afausti, line 27: Same as above.

afausti (commented):
Small fix needed on the `vault_path` for the base environment.  You also need to enable `exposurelog` in `phalanx/science-platform/values-base.yaml`.  That should do it.

services/exposurelog/values-base.yaml
r-owen: How about this entry:
```
    hosts:
      - host: base.lsst.codes
```
should that be base-lsp.lsst.codes as well?

r-owen: Looking at the summit file clearly it should be. I'll fix all three instances of base.lsst.codes.

afausti (approved):
Looks good!

PR #801, opened 2021-06-09
[DM-30667] Restructure Phalanx documentation

rra:
- Move operations documents to a subdirectory
- Add a new section for app developers and add a guide to updating their version
- Move the TLS setup details to the bootstrapping documentation
- Document that Squareone handles the TLS ingress and add this to the bootstrapping notes


jonathansick:
I like this set up :+1: One stretch goal to consider is migrating content from the README into the Sphinx docs and maybe re-envisioning the README as a launchpad from the GitHub homepage into the sphinx docs and into the Argo CD UIs of each deployment.

rra:
Good idea. Done!

frossie (approved):
Yeah I liked Jonathan's suggestion. 

Definitely an improvement, let's live this structure for a bit and see how it goes. 👍

PR #814, opened 2021-06-22
Allow the ops-panda team to access data-int notebook on IDF

hsinfang:


frossie (approved):
Okay I discussed this with Hsinfang and she is clear that only notebook access is needed here.

PR #833, opened 2021-06-28
Turn off debug for most Hub instances

athornton:


rra (approved):
Yeah, looking at the logs, I think this is a good idea. It will also cut back on the number of secrets we're currently leaking into logs.

athornton:
Not sure why I never finished this one up, but let's do that now....

PR #843, opened 2021-06-30
[DM-30904] Update TAP application for 0.2.0 chart

rra:
Depend on 0.2.0 and remove the now-unnecessary override of the
address of the TAP schema database.


rra:
The first CI run will fail because the chart isn't published yet. Will retry.

cbanek (approved):

PR #865, opened 2021-07-13
Default banner message that will be useful sometimes

frossie:


services/squareone/values-idfprod.yaml
```diff
@@ -12,9 +12,10 @@ squareone:
     - name: "pull-secret"
   config:
     siteName: "Rubin Science Platform"
-    # broadcastMarkdown: |
-    #   Write a markdown-formatted broadcast notification.
-
+        siteName: "Rubin Science Platform @ data-int"
+        # broadcastMarkdown: |
+        #    Patch Thursday **today** 3pm-5pm Pacific / 22:00-00:00 UT [(*what is RSP Patch Thursday?)*](https://ls.st/9vz)
+
```
jonathansick, line 18: The linters are reporting an indentation inconsistency (our YAML standard is 2 spaces for an indent). It also looks like the `siteName` config from data-int got copied over? I’m also moving the italics for consistency:

```suggestion
  # broadcastMarkdown: |
  #    Patch Thursday **today** 3pm-5pm Pacific / 22:00-00:00 UT [*(what is RSP Patch Thursday?)*](https://ls.st/9vz)]

```

jonathansick (changes requested):
Good idea. I’ve left a suggestion that resolves the issue the linters are picking up.

One last thing: is the ls.st link useful here? I’m hesitant to use ls.st for production stuff because it doesn’t have all the reliability engineering we need. There also isn’t any need to save characters here.

jonathansick (approved):

jonathansick:
LGTM

PR #881, opened 2021-07-20
Only sync critical early apps if they are enabled in the environment

afausti:


installer/install.sh
```diff
@@ -68,34 +68,38 @@ argocd app sync science-platform \
...
 
 echo "Syncing critical early applications"
-argocd app sync ingress-nginx \
-  --port-forward \
-  --port-forward-namespace argocd
+[ `yq -r .ingress_nginx.enabled ../science-platform/values-$ENVIRONMENT.yaml` == "true" ] && \
+  echo "Syncing ingress-nginx..." && \
+  argocd app sync ingress-nginx \
+    --port-forward \
+    --port-forward-namespace argocd
```
rra, line 75: Because this uses `bash -e`, I think this will be considered a failure if the first expression resolves to false and will then abort the run. I think you'll want to instead write these as if statements, which don't cause a problem with `-e`:
```
if [ $(yq -r .ingress_nginx.enabled ../science-platform/values-$ENVIRONMENT.yaml) == 'true' ]; then
    echo "Syncing ingress-nginx..."
    argocd app sync ingress-nginx \
        --port-forward
        --port-forward-namespace argocd
fi
```
and similarly below. (Using `$()` instead of backticks is optional; I just prefer it because they nest and the syntax is less weird.)

afausti: Thanks, Russ. I have tested this with `bash -e` and it worked fine for the `nts` environment, where some of the critical early apps are disabled. I also prefer `$()` over backticks but I decided to follow the style currently in use. Let me know if you still prefer if statements though, I'm happy to change that.

afausti: Anyways, that's a small change and I agree it improves the script I'll go ahead and fix it.

rra (approved):
Looks good to me. Thanks!

installer/install.sh
rra: Oh, huh, I'm surprised... hm, maybe this doesn't work the way that I thought it did. I can confirm that indeed you're right and a failed `&&` condition doesn't terminate the script with `set -e` in bash.

Oh, indeed, and this is documented, and I was wrong. Sorry about that!

> The shell does not exit if the  command  that  fails is [...] part of any command executed in a && or || list except the command following the final && or ||

PR #885, opened 2021-07-23
DM-31191: Improve secrets script and docs for new apps

jonathansick:
- Enable the update_secrets.sh / generate_secrets.sh scripts in Phalanx to create Vault secrets from 1Password secrets for new applications that aren't already in Vault.
- Add documentation to Phalanx for how to create secrets for applications. Preview the draft at: https://phalanx.lsst.io/v/DM-31191/app-guide/add-a-onepassword-secret.html


docs/app-guide/add-a-onepassword-secret.rst
```diff
@@ -0,0 +1,116 @@
+###########################################
+Add a secret with 1Password and VaultSecret
+###########################################
+
+Whenever possible, static secrets for applications are stored in a 1Password vault before being automatically synced to the Vault service itself and ultimately to Kubernetes ``Secret`` resources via :ref:`vault-secrets-operator`.
```
rra, line 5: I think it may be worth clarifying here that it's not really "whenever possible," it's any secret that for one reason or another we can't randomly generate for new installations. So it's any secret for external systems (Slack webhooks, AWS credentials, Google credentials) or that we need to share with other people (Argo CD admin passwords). We prefer to generate random secrets for anything we can (such as in-cluster Redis passwords) and those aren't stored in 1Password.

docs/app-guide/add-a-onepassword-secret.rst
```diff
@@ -0,0 +1,116 @@
...
+Part 1. Open the 1Password vault
+================================
+
+In one password, access the **LSST IT** 1Password team and open the vault named ``RSP-Vault``.
+Items in this vault are synced into Kubernetes ``Secret`` resources.
+
+Part 2. Create a Secret Note
+============================
+
+Each item in a Kubernetes ``Secret`` corresponds to the contents of a secret note in 1Password.
```
rra, line 19: Either a secure note (1Password calls them that, not secret note) or a login can be used (I prefer logins). Both work the same way.

docs/app-guide/add-a-onepassword-secret.rst
```diff
@@ -0,0 +1,116 @@
...
+(Many secret notes can combined into a single Kubernetes ``Secret`` by configuring the ``VaultSecret``).
+
+- The title of the secret note should be formatted as:
+
+  .. code-block:: text
+
+     {{service}} {{env}} {{secret name}}
```
rra, line 26: I'm not sure anything follows this convention? For instance, your semaphore secrets don't have the secret name in the title; they use the human-readable phrase "GitHub App private key", which I prefer.

docs/app-guide/add-a-onepassword-secret.rst
```diff
@@ -0,0 +1,116 @@
...
+
+  This format is a convention and isn't tied into the automation.
+  The ``env`` can be omitted if the secret applies to all environments.
+
+- Set the **contents** of the secret note to the secret value.
```
rra, line 31: When using a login, set the password field.

docs/app-guide/add-a-onepassword-secret.rst
```diff
@@ -0,0 +1,116 @@
...
+
+  .. code-block:: text
+
+     {{service}} {{secret name}}
+
+  This field provides part of a Vault path for the secret value, which in turn is used by :ref:`vault-secrets-operator` resources to create Kubernetes secrets.
+
+- Add a metadata field labeled ``environment``. The value of that field should be the **hostname** of the RSP environment that this secret applies to (e.g. ``data.lsst.cloud``, not the Phalanx name ``idfprod``).
+
+  If the secret applies to multiple environments, add additional ``environment`` metadata fields.
```
rra, line 43:
```suggestion
  If the secret applies to multiple environments, add an additional ``environment`` metadata field for each additional environment for which that secret applies.
```

rra (approved):
Had some comments on the documentation. Everything else looks good to me once those are taken into account. Thank you for writing this!

docs/app-guide/add-a-onepassword-secret.rst
jonathansick: Sounds good. It sounded like we weren't always using 1password and some folks were still using Vault directly, so I wasn't sure on the policy line.

docs/app-guide/add-a-onepassword-secret.rst
jonathansick: So when using a login item, only the password field is used? The username part is optional/ignored? I couldn't figure out how this worked, so I didn't want to document it.

docs/app-guide/add-a-onepassword-secret.rst
rra: Oh, right, the non-us deployments are not using 1Password. My point was more that we don't want to put the ephemeral secrets in 1Password; they should live only in Vault.

docs/app-guide/add-a-onepassword-secret.rst
rra: Correct, when using a login item only the password is used and the username is ignored (and can be used for documentation purposes).

PR #888, opened 2021-07-23
Improve doc and add new environments

athornton:


athornton:
Just gonna self-merge.

PR #894, opened 2021-07-27
[DM-31083] Add mobu autostart configuration

rra:
- Update for new mobu chart
- Add autostart configuration


athornton (approved):
These are fine, but for IDF you may want to start with:

https://github.com/athornton/system-test branch "prod".  We're now calling it system-test instead of notebook-demo, and my fork contains some fixes we need that Simon should be able to merge once he's back.

PR #908, opened 2021-08-03
[DM-31201] Reconfigure the monkeys for NCSA

rra:
Run one firefighter monkey on int and five on stable, using a
similar configuration as IDF but use notebook-demo and the
robo-simon branch.


athornton (approved):
LGTM

PR #918, opened 2021-08-05
typo

michitaro:
just a typo


rra:
Thank you for noticing and fixing that!

PR #921, opened 2021-08-06
keep IDF pinned at w_25

athornton:


athornton:
nm, setting recommended back to 25 globally

PR #925, opened 2021-08-09
Set up shared workarea for Firefly

athornton:


rra:
Pretty sure using `hostPath` on GKE is not going to work.

rra (approved):

PR #927, opened 2021-08-09
add shared-workarea

athornton:


rra (approved):

services/portal/values-idfprod.yaml
```diff
@@ -3,8 +3,8 @@ firefly:
   replicaCount: 4
 
   volumes:
-    firefly_workarea_nfs:
-      path: /share1/home/firefly/workarea
+    firefly_shared_workarea_nfs:
+      path: /share1/home/firefly/shared-workarea
```
gpdf, line 7: Here you're using Google Cloud resources to create a persistent NFS filesystem that's visible from all the pods, under that pathname, right?

services/portal/values-int.yaml
```diff
@@ -47,7 +47,7 @@ firefly:
     runAsGroup: 102
 
   volumes:
-    firefly_workarea_hostpath: /sui/firefly/workarea
+    firefly_shared_workarea_hostpath: /sui/firefly/workarea
```
gpdf, line 50: This instance of `workarea` on the RHS is part of the "real" GPFS pathname of the filesystem - something that already physically exists on GPFS at NCSA and therefore (at least for now) should not be changed.  Is that correct?

Same for `values-stable.yaml` below.

gpdf (commented):
Looks reasonable to me

services/portal/values-int.yaml
athornton: That is correct.  We would need sysadmin assistance at NCSA to change it...so let's not, at the moment.

services/portal/values-idfprod.yaml
athornton: Correct.  I did that by hand, but we will want to roll it into our FS provisioning automation.

PR #931, opened 2021-08-10
Set IDF int/prod portal to 4x30GB

athornton:


robyww (approved):
looks good.

athornton:
Merged, deployed in int, looks fine, will deploy to prod during usual maint window.

PR #960, opened 2021-09-01
[DM-31490] Update configuration for mobu 4.0.0

rra:
The next release of mobu will not require specifying all the values
to submit to the options form and will default to a Large image.
It also supports testing the latest weekly instead of the recommended
image, so use that on IDF int to test the latest weekly image before
we promote it.

### DIFF
--- a/services/mobu/Chart.yaml
+++ b/services/mobu/Chart.yaml
@@ -3,7 +3,7 @@ name: mobu
 version: 1.0.0
 dependencies:
   - name: mobu
-    version: ">=1.0.0"
+    version: ">=3.0.0"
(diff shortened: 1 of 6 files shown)


athornton (approved):
The only issue I see is really a design decision, and I commented on it a little on the mobu side: now if cachemachine is ill, all the monkeys will start failing all the time, right?

Given that that's definitely user-impacting behavior, that's probably the right thing to do.

athornton:
(obv. fix "services-expanded" to get helm happy)

rra:
I think it only moves the failure, since the spawner is going to fail if cachemachine is down anyway.

PR #963, opened 2021-09-02
[DM-31621] Run workflows on PRs, always run minikube

rra:
Run the workflows on PRs rather than pushes so that they will run
for external submissions.  Redo how minikube is run using an action
to filter on changed paths so that the job will always run (and thus
can be a blocking job for merges), but will quickly do nothing if the
paths changed do not affect minikube.


.github/workflows/ci.yaml
```diff
@@ -44,3 +43,64 @@ jobs:
...
+          filters: |
+            minikube:
+              - ".github/**"
+              - "installer/**"
+              - "science-platform/Chart.yaml"
+              - "science-platform/templates/**"
+              - "science-platform/values-minikube.yaml"
+              - "services/*/Chart.yaml"
+              - "services/*/templates/**"
+              - "services/*/values-minikube.yaml"
```
athornton, line 73: Oh thank goodness.

YES YES YES

athornton (approved):
Ship it.

PR #967, opened 2021-09-03
Add quotas for idf-prod

athornton:


SimonKrughoff (approved):
I don't know if I'd be able to identify an issue if there were one, but it all looks fairly reasonable to me

PR #972, opened 2021-09-07
Pin w_33 at NCSA and IDF

athornton:


SimonKrughoff (approved):
Looks good, but what about idfdev?

athornton:
I was going to let idfdev float.  I can change it if you want.

PR #991, opened 2021-09-23
[DM-31876] Fix auth-uri setting for portal

cbanek:
Add the delegate_to and delegate_scope params to the auth uri,
which were missing and causing trouble.


rra (approved):
Looks great, sorry about that.

PR #993, opened 2021-09-27
Add IDDS config

athornton:


rra (commented):
Per the ticket, I think this was only for data-int. At the least, it doesn't make sense to add to the T&S clusters and presumably not the NCSA clusters as well.

When can we stop duplicating the list of resources from the chart in the configuration for data-int and data? This pattern is going to bite us someday.

athornton:
I've removed the env vars from everything except IDF int.  I've left the configmap in prod because it's going to be mounted, because that comes from the chart.

I have some ideas for how to do this better when we do the work to allow arbitrary resource creation at spawn time (which upstream JupyterHub has indicated interest in).

rra (approved):

PR #999, opened 2021-09-29
tickets/DM-31959 Changes to allow JH 1.4 to spawn, and inject PANDA variables in idf-int

athornton:


services/nublado2/values-idfdev.yaml
```diff
@@ -9,6 +9,7 @@ nublado2:
       hosts: ["data-dev.lsst.cloud"]
       annotations:
         nginx.ingress.kubernetes.io/auth-signin: "https://data-dev.lsst.cloud/login"
+      pathSuffix: "*"
```
rra, line 12: I think this is no longer needed, right?

rra (approved):

services/nublado2/values-idfdev.yaml
athornton: Oh, right, thanks.

PR #1010, opened 2021-10-07
tickets/DM-31595 Nail down external IPs for IDF int/prod

athornton:


rra (approved):
While this still needs to be fixed in idf_deploy as well, this is clearly correct and will be desirable regardless. Thanks!

PR #1031, opened 2021-10-29
DM-32356: Add alert-stream-broker application

spenczar:
This application runs a Kafka broker for the alert stream.

Everything is left with the defaults in https://github.com/lsst-sqre/charts/tree/master/charts/alert-stream-broker.


afausti (approved):
That looks all right.

PR #1032, opened 2021-10-29
PREOPS-858: Deploy plot-navigator on data-int.

ctslater:


services/plot-navigator/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v2
+name: plot-navigator
+version: 1.0.0
+dependencies:
+- name: plot-navigator
+  version: ">=1.0.0"
```
spenczar, line 6: This looks a little odd to me. I'd think you'd want to pin a specific version when deploying in phalanx, but maybe this is fine?

science-platform/values-minikube.yaml
```diff
@@ -36,6 +36,8 @@ nublado2:
   enabled: true
 obstap:
   enabled: true
+plot_navigator:
+  enabled: false
```
ctslater, line 40: I'm not sure if I'm supposed to turn this on for minikube? I don't know if it's for CI or for demo purposes.

cbanek: it probably won't work since I don't think the butler parts work there right now, but it was useful just for testing and such.

cbanek (approved):

PR #1038, opened 2021-11-02
Grant exec:notebook and read:tap to lsst-ops:panda

rra:
Folks in that group are using int to test things that may be useful
for having Firefly and TAP for.


gpdf (approved):
Exactly as discussed; please proceed!

PR #1040, opened 2021-11-05
[DM-32477] Remove Elasticsearch logging service

rra:
We've decided to stop running our own ELK stack and rely on Log
Explorer on GKE environments and regular Kubernetes logs on the
other environments.  Remove the logging application and its
configuration and documentation.


installer/generate_secrets.py
```diff
@@ -133,54 +132,6 @@ def _tap(self):
...
-            },
-            "admin": {
-                "hash": admin[1],
-                "reserved": True,
-                "backend_roles": ["admin"],
-                "description": "Admin user",
-            },
-            "kibana": {
-                "hash": kibana[1],
-                "backend_roles": ["kibanauser"],
```
cbanek, line 161: aaah, that looks so good in red.

cbanek (approved):

PR #1046, opened 2021-11-10
tickets/DM-30329-1

mareuter:
TTS: Add NFS mounts from archivers for nublado2.


athornton (approved):
I don't have any insight into whether this is really where the filesystems are or where they should be mounted, but assuming you do, yeah, this looks fine.

PR #1094, opened 2021-12-08
Automatically reconcile VaultSecrets with Vault's state

spenczar:
This describes what I had to do to update the alert-stream-broker's VaultSecret's child Secret.

---

Update:

After discussion, this now goes beyond just documentation. It also changes the configuration of Vault Secrets Operator in each environment, enabling automatic reconciliation.

### DIFF
--- a/docs/service-guide/update-a-onepassword-secret.rst
+++ b/docs/service-guide/update-a-onepassword-secret.rst
@@ -1,0 +1,30 @@
+#####################################################
+Updating a secret stored in 1Password and VaultSecret
+#####################################################
+
+Secrets that are stored in 1Password are synchronized into Vault using the `installer/generate_secrets.py <https://github.com/lsst-sqre/phalanx/blob/master/installer/generate_secrets.py>`__ script.
+Once they are in Vault, they are accessible to the Vault Secrets Operator, which responds to creation of any ``VaultSecret`` resources in Kubernetes by grabbing the current value of the secret data in Vault.
+
+The Vault Secrets Operator doesn't independently detect if there are changes to data in Vault, though.
+It only provides a snapshot of the data in Vault at the time the ``VaultSecret`` was created.
+
+So, if you want to make any changes to a ``VaultSecret``'s data, you'll need to:
+
+1. Make the changes in 1Password
+2. Run the `installer/update_secrets.sh <https://github.com/lsst-sqre/phalanx/blob/master/installer/update_secrets.sh>`__ script, as described in :doc:`/service-guide/add-a-onepassword-secret`.
+3. Delete the ``VaultSecret`` in the ArgoCD UI.
+4. Sync your application in ArgoCD, recreating the ``VaultSecret``, and triggering the Vault Secrets Operator to incorporate your changes.
+
+
+Deleting the ``VaultSecret``
+========================
+
+You can delete the ``VaultSecret`` by first navigating to your Application in ArgoCD.
+
+Find the ``VaultSecret`` resource in your application, and click on it.
+You should see a "delete" button in the top right of the dialog that pops up.
+
+Click "delete", and go through with it; a "foreground delete" is sufficient.
+Be sure not to navigate away from this page until you see the resource get deleted in the UI, since ArgoCD will delete it and the underlying Secret with background calls.
+
+Once done, you can Sync your application in order to recreate the VaultSecret.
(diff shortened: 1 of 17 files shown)


docs/service-guide/update-a-onepassword-secret.rst
```diff
@@ -0,0 +1,30 @@
...
+So, if you want to make any changes to a ``VaultSecret``'s data, you'll need to:
+
+1. Make the changes in 1Password
+2. Run the `installer/update_secrets.sh <https://github.com/lsst-sqre/phalanx/blob/master/installer/update_secrets.sh>`__ script, as described in :doc:`/service-guide/add-a-onepassword-secret`.
+3. Delete the ``VaultSecret`` in the ArgoCD UI.
+4. Sync your application in ArgoCD, recreating the ``VaultSecret``, and triggering the Vault Secrets Operator to incorporate your changes.
+
+
+Deleting the ``VaultSecret``
+========================
```
jonathansick, line 20: Could you extend the underline out just a bit?

docs/service-guide/update-a-onepassword-secret.rst
```diff
@@ -0,0 +1,30 @@
+#####################################################
+Updating a secret stored in 1Password and VaultSecret
+#####################################################
+
+Secrets that are stored in 1Password are synchronized into Vault using the `installer/generate_secrets.py <https://github.com/lsst-sqre/phalanx/blob/master/installer/generate_secrets.py>`__ script.
+Once they are in Vault, they are accessible to the Vault Secrets Operator, which responds to creation of any ``VaultSecret`` resources in Kubernetes by grabbing the current value of the secret data in Vault.
+
+The Vault Secrets Operator doesn't independently detect if there are changes to data in Vault, though.
+It only provides a snapshot of the data in Vault at the time the ``VaultSecret`` was created.
```
jonathansick, line 9: Maybe this is something we could fix instead of documenting this (or using this doc as a fall-back). In the vault-secrets-operator doc, they talk about two ways for secrets to get updated:

1. Use a `spec.version` field to specify the version of the Vault secret you want to use. I guess in that case we'd need to use the vault CLI or have a wrapper that reports on the different versions of a secret in vault.
2. Have vault-secrets-operator be configured to update secrets via the `VAULT_RECONCILIATION_TIME` environment variable configuration.

> The spec.type and spec.keys fields are handled in the same way for both versions of the KV secret engine. The spec.version field is only processed, when the secret is saved under a KVv2 secret engine. If you specified the VAULT_RECONCILIATION_TIME environment variable with a value greater than 0 every secret is reconciled after the given time. This means, when you do not specify spec.version, the Kubernetes secret will be automatically updated if the Vault secret changes.

I notice in [our configuration](https://github.com/lsst-sqre/phalanx/blob/master/services/vault-secrets-operator/values-base.yaml) we haven't set this, so I wonder if the value is zero?

@rra What do you think? Should we change vault-secrets-operator to automatically reconcile with Vault?

jonathansick (commented):
This looks good as a doc, but I think it'd be great to have a say from @rra or the others on whether we should make Vault reconciliation automatic (see the inline comment)

docs/service-guide/update-a-onepassword-secret.rst
spenczar: Oh, I like this _much_ more. I would very much prefer to have `VAULT_RECONCILIATION_TIME=60` or something, and then document that it takes up to 60 seconds for changes to be reflected into Kubernetes Secrets.

rra: Yeah, let's set `VAULT_RECONCILIATION_TIME`. Spencer, do you want to add that to this diff? Looks like it's just a matter of adding `vault.reconciliationTime` to the `values-*.yaml` files for all the environments. 60 seconds is a bit aggressive, but hopefully the API call is cheap, so we can start there and change it if there's some problem.

docs/service-guide/update-a-onepassword-secret.rst
```diff
@@ -0,0 +1,30 @@
...
+
+So, if you want to make any changes to a ``VaultSecret``'s data, you'll need to:
+
+1. Make the changes in 1Password
+2. Run the `installer/update_secrets.sh <https://github.com/lsst-sqre/phalanx/blob/master/installer/update_secrets.sh>`__ script, as described in :doc:`/service-guide/add-a-onepassword-secret`.
+3. Delete the ``VaultSecret`` in the ArgoCD UI.
+4. Sync your application in ArgoCD, recreating the ``VaultSecret``, and triggering the Vault Secrets Operator to incorporate your changes.
+
+
+Deleting the ``VaultSecret``
```
rra, line 19: I think you can drop this section entirely; I don't think there's any reason why one would want to do this outside of deleting the whole application or syncing a new version.

jonathansick:
oh, thank you for doing this. I was just about to do the update. You can merge this, and then let's sync on idf-dev. If that doesn't blow up vault-secrets-operator, we can sync on idf-int and you'll get the benefit for your work. We only update idf-prod during the Thursday maintenance. See https://sqr-056.lsst.io/#process

So if you feel able to sync, go for it, otherwise we can do it.

PR #1099, opened 2021-12-09
[DM-32856] Add sherlock to science platform

cbanek:


athornton (approved):
Looks simple enough.  Think it's fine.

PR #1146, opened 2022-01-12
[DM-33239] Remove Argo Workflows

rra:
mobu can now do everything we were using Argo Workflows to do, and
the version we're currently installing is no longer supported and
doesn't work with Kubernetes 1.22.  Remove it from all environments.


athornton (approved):
LGTM

PR #1160, opened 2022-01-20
Use new read-alertdb scope for alertdb

spenczar:
This will be used by the alert database.

Just using some of the traditional GH teams for groups backing it.


rra (approved):
Looks reasonable to me.

PR #1161, opened 2022-01-21
[DM-33166] Add Sasquatch to Phalanx

afausti:
In this PR we start by adding the sasquatch service to phalanx using the sasquatch chart as a dependency.

As a proof of concept we move the sasquatch chart to phalanx. This greatly simplifies the development workflow,  because there is no need to create and release a separate chart for every phalanx service anymore. I particularly like the fact that the default values and the environment-specific values are next to each other in this setup.

Some charts used only by sasquatch like `kafka-connect-manager` can also be removed from the charts repo and be added as a sub chart here. The secrets needed by these sub charts can be provided by sasquatch now via `VaultSecret` resource. Other charts like the `strimzi-registry-operator` can still live in the charts repository (or even in the application repository) and made generally available. 

This also suggests eliminating the `pull-secrets` chart entirely in favor of creating the `VaultSecret` resource in phanlax.

We can run helm-docs for the phalanx service now.


services/sasquatch/charts/kafka-connect-manager/templates/jdbc_sink.yaml
```diff
@@ -0,0 +1,67 @@
...
+  selector:
+    matchLabels:
+      app: kafka-connect-manager
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app: kafka-connect-manager
+        app.kubernetes.io/instance: {{ .Release.Name }}
+    spec:
```
rra, line 44: Security policies here too.

services/sasquatch/charts/kafka-connect-manager/templates/mirrormaker2.yaml
```diff
@@ -0,0 +1,90 @@
...
+  selector:
+    matchLabels:
+      app: kafka-connect-manager
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app: kafka-connect-manager
+        app.kubernetes.io/instance: {{ .Release.Name }}
+    spec:
```
rra, line 62: Security policies.

services/sasquatch/charts/kafka-connect-manager/templates/s3_sink.yaml
```diff
@@ -0,0 +1,88 @@
...
+  selector:
+    matchLabels:
+      app: kafka-connect-manager
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app: kafka-connect-manager
+        app.kubernetes.io/instance: {{ .Release.Name }}
+    spec:
```
rra, line 55: Security policies.

services/sasquatch/charts/kafka-connect-manager/templates/s3_sink.yaml
```diff
@@ -0,0 +1,88 @@
...
+        - kafkaconnect
+        - create
+        - s3-sink
+        - --show-status
+        - /etc/s3-sink/s3-sink-config.json
+        env:
+          - name: AWS_ACCESS_KEY_ID
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.s3Sink.awsSecret | quote }}
```
rra, line 70: Now that everything is in Phalanx, we know the secret name already since we're going to create it with a `VaultSecret`, so this no longer needs to be a configuration parameter, I think.

services/sasquatch/charts/kafka-connect-manager/templates/s3_sink.yaml
```diff
@@ -0,0 +1,88 @@
...
+                key: aws_access_key_id
+          - name: AWS_SECRET_ACCESS_KEY
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.s3Sink.awsSecret | quote }}
```
rra, line 75: Likewise here.

services/sasquatch/charts/kafka-connect-manager/values.yaml
```diff
@@ -0,0 +1,160 @@
+# Default values for kafka-connect-manager.
+# See also https://kafka-connect-manager.lsst.io
+image:
+  repository: lsstsqre/kafkaconnect
+  tag: 0.9.3
+  pullPolicy: Always
```
rra, line 6: Missing helm-docs comments here.

services/sasquatch/charts/kafka-connect-manager/README.md.gotmpl
```diff
@@ -0,0 +1,7 @@
+{{ template "chart.header" . }}
```
rra, line 1: Do you need these templates? I've never used them before.

services/sasquatch/charts/strimzi-kafka/templates/connect.yaml
```diff
@@ -0,0 +1,30 @@
...
+      key: user.key
+  config:
+    group.id: {{ .Values.cluster.name }}-connect
+    offset.storage.topic: {{ .Values.cluster.name }}-connect-offsets
+    config.storage.topic: {{ .Values.cluster.name }}-connect-configs
+    status.storage.topic: {{ .Values.cluster.name }}-connect-status
+    # -1 means it will use the default replication factor configured in the broker
+    config.storage.replication.factor: -1
+    offset.storage.replication.factor: -1
+    status.storage.replication.factor: -1
```
rra, line 30: I think I'd find this easier to read if it were written in normal YAML syntax, e.g.:
```
offset:
  storage:
    topic: {{ .Values.cluster.name }}-connect-offets
    replication:
      factor: -1
# ...
```

services/sasquatch/charts/strimzi-kafka/values.yaml
```diff
@@ -0,0 +1,45 @@
...
+    # -- Name of a StorageClass to use when requesting persistent volumes.
+    storageClassName: ""
+  # -- Configuration overrides for the Kafka server.
+  config:
+    # -- Number of minutes for a consumer group's offsets to be retained.
+    offsets.retention.minutes: 10080
+    # -- Number of days for a topic's data to be retained.
+    log.retention.hours: 168
+    # -- Maximum retained number of bytes for a topic's data.
+    log.retention.bytes: "644245094400"
```
rra, line 23: Do you want this much nested hierarchy? In other words, why not `offsetsRetentionMinutes`, etc.?

services/sasquatch/templates/vault-secret.yaml
```diff
@@ -0,0 +1,16 @@
...
+  path: {{ .Values.vaultSecretsPath }}/sasquatch
+  type: Opaque
+---
+apiVersion: ricoberger.de/v1alpha1
+kind: VaultSecret
+metadata:
+  name: pull-secret
+spec:
+  path: {{ .Values.vaultSecretsPath }}/pull-secret
+  type: kubernetes.io/dockerconfigjson
```
rra, line 16: Ugh, I'm not sure what to think about this. I would rather all the references go through the pull-secret chart for consistency, but I can see how the configuration is a bit more annoying. I think the right solution is to use global variables to affect the pull-secret chart, though, so I think I would still lean towards using the pull-secret chart instead of replicating its contents here.

services/sasquatch/values-idfdev.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+chronograf:
+  ingress:
+    enabled: true
+    hostname: data-dev.lsst.cloud
+
+  env:
+    GENERIC_NAME: "CILogon"
+    GENERIC_AUTH_URL: https://data-dev.lsst.cloud/auth/openid/login
+    GENERIC_TOKEN_URL: https://data-dev.lsst.cloud/auth/openid/token
```
rra, line 21: For the record, this will use GitHub, not CILogon, on data-dev, at least for now. CILogon may be correct in the long run, though.

services/sasquatch/values.yaml
```diff
@@ -0,0 +1,100 @@
+# Default values for Sasquatch.
```
rra, line 1: This is missing a lot of helm-docs comments, and it would be nice to generate a `README.md` for the services/sasquatch directory as well in this new style.

rra (approved):
A bunch of fairly minor comments, but the overall structure looks good. Feel free to commit this before all the comments are addressed if it makes it easier to test. Thank you for working on this! I think I like the new structure a lot better.

services/sasquatch/charts/kafka-connect-manager/templates/s3_sink.yaml
afausti: Yes, thanks! that makes sense.

services/sasquatch/charts/strimzi-kafka/templates/connect.yaml
afausti: Oh, those are Kafka configuration options, it didn't occur to me I could use normal YAML syntax here, I'll try.

services/sasquatch/charts/strimzi-kafka/values.yaml
afausti: Yeah, those are Kafka configuration options, I cannot change that. What is under the config section overrides the default kafka configuration which is written in a `ConfigMap`.

services/sasquatch/templates/vault-secret.yaml
afausti: Oh, I thought you would like the idea of eliminating the pull-secret chart entirely :)

services/sasquatch/values-idfdev.yaml
afausti: Yep, I notice this after I deployed sasquatch on data-dev. I'll ask tomorrow during stand-up about which environments I could use to test CILogon.

services/sasquatch/templates/vault-secret.yaml
rra: Hm, actually, maybe that's not a bad idea. Maybe let's talk about this in the next stand-up or co-work.

PR #1162, opened 2022-01-22
[DM-33241] Update for new version of TAP chart

rra:


rra:
Tests will fail until https://github.com/lsst-sqre/charts/pull/606 has been merged.

PR #1193, opened 2022-02-03
DM-33534: Add production-tools to data-int

ctslater:


services/production-tools/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if .Values.ingress.enabled -}}
```
rra, line 1: You can remove this conditional as well.

services/production-tools/templates/secrets.yaml
```diff
@@ -0,0 +1,8 @@
+---
```
rra, line 1: Conventionally, we call this file `vault-secrets.yaml`. You'll need to install the pull-secret here as well.

services/production-tools/templates/secrets.yaml
```diff
@@ -0,0 +1,8 @@
...
+apiVersion: ricoberger.de/v1alpha1
+kind: VaultSecret
+metadata:
+  name: butler-secret
+spec:
+  path: {{ .Values.butler_secret_path }}
```
rra, line 7: This shouldn't be a separate configuration parameter and instead should use `global.vaultSecretsPath` (see how Gafaelfawr does this).

services/production-tools/templates/service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "production-tools.fullname" . }}
+  labels:
+    {{- include "production-tools.labels" . | nindent 4 }}
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+    - port: {{ .Values.service.port }}
```
rra, line 10: Hard-code these rather than making them configurable.

services/production-tools/values-idfint.yaml
```diff
@@ -0,0 +1,13 @@
+production-tools:
+  butler_secret_path: "secret/k8s_operator/data-int.lsst.cloud/butler-secret"
+  environment:
+    BUTLER_URI: "s3://butler-us-central1-panda-dev/dc2/butler-external.yaml"
+    PGPASSFILE: "/home/worker/.lsst/postgres-credentials.txt"
+    AWS_SHARED_CREDENTIALS_FILE: "/home/worker/.lsst/aws-credentials.ini"
+    S3_ENDPOINT_URL: "https://storage.googleapis.com"
```
rra, line 7: None of these vary by environment and should therefore be hard-coded in the deployment so that you don't have to repeat them in every values file.

services/production-tools/values-idfint.yaml
```diff
@@ -0,0 +1,13 @@
+production-tools:
```
rra, line 1: Since the chart is directly in Phalanx, you should delete this level and put the settings at the top level.

services/production-tools/values-idfint.yaml
```diff
@@ -0,0 +1,13 @@
...
+  butler_secret_path: "secret/k8s_operator/data-int.lsst.cloud/butler-secret"
```
rra, line 2: You can drop this setting given the above changes.

services/production-tools/values-idfint.yaml
```diff
@@ -0,0 +1,13 @@
...
+  environment:
+    BUTLER_URI: "s3://butler-us-central1-panda-dev/dc2/butler-external.yaml"
+    PGPASSFILE: "/home/worker/.lsst/postgres-credentials.txt"
+    AWS_SHARED_CREDENTIALS_FILE: "/home/worker/.lsst/aws-credentials.ini"
+    S3_ENDPOINT_URL: "https://storage.googleapis.com"
+    LOG_BUCKET: "drp-us-central1-logging"
+    LOG_PREFIX: "Panda-RubinLog"
+  ingress:
+    host: "data-int.lsst.cloud"
+    path: "/production-tools"
```
rra, line 12: These two will go away with the above changes.

services/production-tools/values-idfint.yaml
```diff
@@ -0,0 +1,13 @@
...
+    gafaelfawrAuthQuery: "scope=exec:portal&delegate_to=productiontools"
```
rra, line 13: Do you need delegated Gafaelfawr tokens for something? That's only required if you need to make authenticated calls on behalf of the user to other services. (If you do need this, use `production-tools` to match the name of the service.)

Also, this should be set in `values.yaml`, not here, since it won't vary by environment.

services/production-tools/values.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+# -- Number of web deployment pods to start
+replicaCount: 1
+
+image:
+  # -- Image to use in the production-tools deployment
+  repository: lsstdm/production_tools
+
+  # -- Pull policy for the production-tools image
+  pullPolicy: Always
```
rra, line 13: This should be `IfNotPresent` by default.

services/production-tools/values.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+  # -- Overrides the image tag whose default is the chart appVersion.
+  tag: ""
+
+# -- Secret names to use for all Docker pulls
+imagePullSecrets: []
```
rra, line 19: Can delete this.

services/production-tools/values.yaml
```diff
@@ -0,0 +1,87 @@
...
+nameOverride: ""
+
+# -- Override the full name for resources (includes the release name)
+fullnameOverride: ""
+
+# -- Annotations for the production-tools deployment pod
+podAnnotations: {}
+
+# Path to the vault secret.
+butler_secret_path: ""
```
rra, line 31: Can delete this.

services/production-tools/values.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+# Environment variables passed to container
+environment: {}
+
+service:
+  # -- Type of service to create
+  type: ClusterIP
+
+  # -- Port of the service to create and map to the ingress
+  port: 8080
```
rra, line 41: Can delete this.

services/production-tools/values.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+ingress:
+  # -- Create an ingress resource
+  enabled: true
+
+  # -- Hostname of the deployment to run behind
+  host: ""
+
```
rra, line 49: Can delete this.

services/production-tools/values.yaml
```diff
@@ -0,0 +1,87 @@
...
+  gafaelfawrAuthQuery: ""
+
+  # -- Additional annotations for the ingress rule
+  annotations: {}
+
+  # -- Path type for the ingress rule
+  pathType: ImplementationSpecific
+
+  # -- URL path to dispatch to the production-tools deployment pod
+  path: "/production-tools"
```
rra, line 61: Can delete this. Also the `pathType` when you hard-code it should be `Prefix`, not `ImplementationSpecific`.

rra (commented):
In addition to those changes, you'll want to add the `global` block to the bottom of `values.yaml` so that all parameters are documented (see Gafaelfawr for an example), copy over `README.md.gotmpl` from another service like Gafaelfawr, and then generate the documentation with `helm-docs -c services/production-tools`. You can get `helm-docs` from https://github.com/norwoodj/helm-docs

ctslater:
Thank a lot @rra, that's super helpful. It's much nicer with that boilerplate code removed. Pushed a new commit with all the changes.

services/production-tools/templates/deployment.yaml
```diff
@@ -0,0 +1,110 @@
...
+            - "/bin/bash"
+            - "-c"
+            - |
+              cp -RL /home/worker/secrets-raw/* /home/worker/.lsst/
+              chown worker:worker /home/worker/.lsst/*
+              chmod 0400 /home/worker/.lsst/*
+          securityContext:
+            runAsNonRoot: false
+            runAsUser: 0
+            runAsGroup: 0
```
rra, line 54: I realized in a subsequent iteration of the vo-cutouts chart that you can delete the `chown` line (but keep the other two) and use the standard `securityContext` section of:
```
          securityContext:
            allowPrivilegeEscalation: false
            capabilities:
              drop:
                - "all"
```
The init container doesn't have to run as root because the files will be mounted world-readable. It can run as the regular user and the act of copying them will fix the ownership, which will make PostgreSQL happy (along with the `chmod`).

services/production-tools/templates/vault-secrets.yaml
```diff
@@ -0,0 +1,18 @@
+---
+apiVersion: ricoberger.de/v1alpha1
+kind: VaultSecret
+metadata:
```
rra, line 4: Nit: For consistency, also add the labels stanza here similar to pull-secret below.

science-platform/templates/production-tools-application.yaml
```diff
@@ -0,0 +1,27 @@
...
+  destination:
+    namespace: production-tools
+    server: https://kubernetes.default.svc
+  project: default
+  source:
+    path: services/production-tools
+    repoURL: {{ .Values.repoURL }}
+    targetRevision: {{ .Values.revision }}
+    helm:
+      valueFiles:
```
rra, line 24: Before `valueFiles` you'll need to add this stanza to pass down the globals:
```
      parameters:
        - name: "global.host"
          value: {{ .Values.fqdn | quote }}
        - name: "global.baseUrl"
          value: "https://{{ .Values.fqdn }}"
        - name: "global.vaultSecretsPath"
          value: {{ .Values.vault_path_prefix | quote }}
```

rra (commented):
Almost there, just a few more minor changes.

rra (approved):
Looks good!

ctslater:
Thanks @rra!

PR #1205, opened 2022-02-11
DM-33661: narrativelog: add support

r-owen:
including configuration for summit, base, and Tucson test stand.

Also tweak a few exposurelog settings.

### DIFF
--- a/services/exposurelog/values-roe.yaml
+++ b/services/exposurelog/values-roe.yaml
@@ -1,8 +1,13 @@
+# WARNING: this is a "playground" deployment
+# using exposurelog's built-in test butler registries.
 exposurelog:
   pull_secret: pull-secret
 
-  site_id: roe
-  # Use the built-in toy registry, so set neither butler_uri_1 nor butler_uri_2
+  site_id: test
+  # Use the test butler registries.
+  # Note: exposurelog's Dockerfile copies the test repos to the top of the container
+  butler_uri_1: LSSTCam
+  butler_uri_2: LATISS
 
   ingress:
     enabled: true
--- a/science-platform/templates/narrativelog-application.yaml
+++ b/science-platform/templates/narrativelog-application.yaml
@@ -1,0 +1,26 @@
+{{- if .Values.narrativelog.enabled -}}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: narrativelog
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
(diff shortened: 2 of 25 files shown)


setup.cfg
```diff
@@ -0,0 +1,3 @@
+[flake8]
```
athornton, line 1: Pretty sure we use line-length 79; not sure we even use flake8.  Take a look at the settings for https://github.com/lsst-sqre/mobu (in pyproject.toml)--I think those are canonical.

athornton (changes requested):
Bring the black/flake8 settings into line with other SQuaRE projects and I think we're good.  The narrativelog stuff looks fine, and I'm glad the linter caught useless f-strings and unused imports and stuff.

setup.cfg
jonathansick: Our "standard" configurations aren't in mobu — they're in the template: https://github.com/lsst/templates/tree/main/project_templates/fastapi_safir_app/example

Mobu has a few extra configs for its specific use cases.

r-owen:
I have moved the reformat to a different PR that I'll assign to @jonathansick once this one is merged and I can rebase it. So now the only changes are due to narrativelog. Please have another look.

r-owen:
As to the failed github action...I'm wondering if this needs the actual narrativelog service to pass. That's waiting for review so has not yet been tagged...

athornton (approved):
LGTM as long as you're sure that changing site_id isn't going to break roe.

jonathansick:
We've got to finish https://github.com/lsst-sqre/charts/pull/646 first and publish that chart so the helm CI test can try to deploy narrativelog in minikube and validate that the chart works.

services/exposurelog/values-roe.yaml
```diff
@@ -1,8 +1,13 @@
+# WARNING: this is a "playground" deployment
+# using exposurelog's built-in test butler registries.
 exposurelog:
   pull_secret: pull-secret
 
-  site_id: roe
```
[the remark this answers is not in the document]
r-owen: Good question. This deployment is a sandbox (it is using the built-in test registries) and so it should use site_id=test. That said, it's not a big deal. It only controls the site_id field for new messages.

setup.cfg
r-owen: I pulled it out. I hope somebody will fix this because it's a nuisance to edit files when your editor is set to reformat them and there is no format configuration in the package.

PR #1207, opened 2022-02-14
Reformat files with black

r-owen:


jonathansick (changes requested):
Let's hold off on this PR and instead we'll bring in our standard linting and formatting toolchain to go along with this widespread reformatting of Phalanx's infrastructure.

rra:
This has now been implemented via pre-commit in master.

PR #1208, opened 2022-02-15
remove NTS from Phalanx.  RIP 2022-02-11

athornton:


afausti (approved):
Nice! I think it removes every trace of the NTS environment in Phalanx.

athornton:
Now, I will revoke the keys for the enclave, and delete the data from the enclave.

PR #1215, opened 2022-02-18
Document database addition (and do a step I missed last time)

athornton:


docs/ops/postgres/add-database.rst
```diff
@@ -0,0 +1,136 @@
...
+push them into Vault.  That, however, would require that you restart
+everything with randomly-generated passwords, and that's a fairly
+disruptive operation, so you probably are better off manually injecting
+just your new password.
+
+* Consult ``1Password`` and retrieve the appropriate vault write token for
+  the instance you're working with from ``vault_keys.json``.
+* Set up your environment: ``export VAULT_ADDR=vault.lsst.codes ; export
+  VAULT_FORMAT=json ; export VAULT_TOKEN=<retrieved-token>``
+* Run ``vault kv get secret/k8s_operator/<instance>/postgres >
```
afausti, line 105: I think you can simplify these steps by suggesting simply `vault kv patch secret/k8s_operator/<instance>/postgres exposurelog_password=<password>`

afausti (commented):
It looks good. I like the detailed documentation. I just made a suggestion on the section about manually adding a secret to Vault.

docs/ops/postgres/add-database.rst
athornton: Oh!  Yes, that would be simpler than what I've been doing.

athornton: Done.

PR #1238, opened 2022-03-01
[DM-33804] Add Telegraf to Sasquatch

afausti:
- Add InfluxData Telegraf chart as a dependency to Sasquatch
- Configure the Telegraf Prometheus input plugin to collect [JupyterHub metrics](https://jupyterhub.readthedocs.io/en/stable/reference/metrics.html) and write them to InfluxDB.


athornton (approved):
Seems fine to me.  I was a little worried about `hub.nublado2` but the Proxy API service isn't exposed externally (and should not be), so that makes sense.

PR #1242, opened 2022-03-03
tickets/DM-33827: Enable Flux in IDF dev

athornton:
I'm sure this is conceptually correct.  Whether it's `flux-enabled`, `flux-enable`, `flux_enabled`, `fluxEnabled`, or some other permutation isn't totally clear to me, but it'd be flux-enabled in the config file.  I don't quite understand mapping that to the Helm chart to the Phalanx config yet.


afausti (commented):
Yeah, let's talk about the sasquatch organization in phalanx after the webinar.
You'll see there's a value.yaml in the phalanx service, which is kind of a new thing as proposed in RFC-830.

PR #1243, opened 2022-03-03
tickets/DM-33827: Apparently it really is a dash in the config

athornton:


afausti (approved):
I agree that's confusing

PR #1245, opened 2022-03-04
[DM-33924] Always use dash in the InfluxDB configuration keys

afausti:
The upstream InfluxDB helm chart used in sasquatch does not rewrite the configuration keys replacing "_" with "-" anymore. Glad we found this today, thanks @athornton


athornton (approved):
LGTM

PR #1272, opened 2022-03-15
[DM-33846] Deploy sasquatch on NCSA int

afausti:
This PR enables sasquatch on int and adds specific configuration for this environment.


athornton (approved):
Seems fine.

PR #1275, opened 2022-03-18
tickets/DM-33976: break out telegraf and point at monitoring.lsst.codes

athornton:


science-platform/templates/telegraf-application.yaml
```diff
@@ -0,0 +1,29 @@
...
+  destination:
+    namespace: telegraf
+    server: https://kubernetes.default.svc
+  project: default
+  source:
+    path: services/telegraf
+    repoURL: {{ .Values.repoURL }}
+    targetRevision: {{ .Values.revision }}
+    helm:
+      valueFiles:
```
afausti, line 27: Here I think you also need to list the `values.yam`l file which has the default telegraf configuration. It has to be first in the list followed by the environment-specific configuration so that they take precedence.

afausti (changes requested):
Looks great, but see my comment on getting the default configuration for telegraf.

I agree with keeping telegraf in Sasquatch for now, thanks for putting it back.

afausti (approved):

PR #1284, opened 2022-03-23
DM-33922: exposurelog: add a writable /tmp dir

r-owen:
in order to make matplotlib happy.
Also tweak existing volumeMount entries to put the name first, for consistency with other deployments.
Also add README files for exposurelog and narrativelog.


athornton (approved):
LGTM

PR #1298, opened 2022-03-24
values for idf dev culling

dspeck1:
Changes to test culling in rsp dev


athornton (approved):
LGTM!

athornton (approved):
LGTM, as long as...we're not running these values in dev forever, right?

PR #1318, opened 2022-03-31
Tickets/DM-34197B: enable metrics collection for K8s via telegraf-ds

athornton:


afausti (approved):
That looks great, telegraf-ds as is it provides an opinionated configuration for monitoring Kubernetes. Note you can also use the `override_config` method to customize it if needed. I did that in the EFD to use the `kube_inventory` input plugin because I wanted to monitor pod restarts. But this is looking great - collecting metrics from all those different Prometheus endpoints ohoo!

PR #1319, opened 2022-03-31
deploying new cachemachine gar image to dev

dspeck1:
deploying cachemachine gar to idf dev


athornton (changes requested):
Rather than putting something into the config you know you're gonna take out (that is, the cachemachine image tag)...I'd just edit the CM and the deployment on the fly with `kubectl edit`, and then fix up the image tag tag in the PR to point at the new actual release, which you can cut once you know your ticket build is working fine with the new configmap.

athornton (approved):
I'm not wild about having hardcoded the 'sciplat-lab' image name in cachemachine's code, but this will certainly work for our present use case.

PR #1320, opened 2022-03-31
tickets/DM-34218: expand "clear .local"

athornton:
And while I'm in there, modernize nublado2 helm chart (pull chart into Phalanx).

### DIFF
--- a/services/nublado2/values.yaml
+++ b/services/nublado2/values.yaml
@@ -1,0 +1,480 @@
+# Default values for nublado2.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+jupyterhub:
+  hub:
+    authenticatePrometheus: false
+    image:
+      name: lsstsqre/nublado2
+      tag: "2.1.0"
+    config:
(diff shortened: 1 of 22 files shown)


services/nublado2/values.yaml
```diff
@@ -0,0 +1,483 @@
...
+      kind: VaultSecret
+      metadata:
+        name: pull-secret
+        namespace: "{{ user_namespace }}"
+      spec:
+        path: "{{ pull_secret_path }}"
+        type: kubernetes.io/dockerconfigjson
+
+# Note: See note above about existingSecret.
+vault_secret_name: "nublado2-secret"
```
rra, line 477: May as well just hard-code this in the resource since we don't ever override it.

rra (approved):
This isn't really reviewable since it's mostly whitespace changes that GitHub doesn't show very well. I'm assuming that all the individual values files are correct.

services/nublado2/values.yaml
athornton: OK.

PR #1328, opened 2022-04-06
DM-34343: Migrate squareone chart templates into Phalanx

jonathansick:
This PR migrates the squareone Helm chart into phalanx/services, per RFC-830.

This update also takes advantage of the Argo CD Application to pass environment settings down to the squareone chart as globals. Some things that this PR doesn't fully resolve are:

- It's still using vault secret path baked-in `values-*.yaml` for the pull secret; ultimately I plan to drop the pull secret anyways on the next squareone release, which will be available from `ghcr.io`.
- The ingress `tls[].hosts` section still uses the hostname in `values-*.yaml` files. I haven't totally thought-through how to plumb the environment's host parmeter down to the ingress TLS, so I'm open to ideas.


services/squareone/templates/NOTES.txt
```diff
@@ -0,0 +1,22 @@
+1. Get the application URL by running these commands:
```
rra, line 1: We've been deleting these files (which come from the default Helm starter) since when using Argo CD you'll never see them, and now that the chart is only in Phalanx, it's unlikely anyone would install it manually with Helm and thus see this message.

services/squareone/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
...
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "squareone.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create }}
+{{- default (include "squareone.fullname" .) .Values.serviceAccount.name }}
+{{- else }}
+{{- default "default" .Values.serviceAccount.name }}
+{{- end }}
+{{- end }}
```
rra, line 62: Do you need a service account? You only need this if the application is going to make Kubernetes calls (or in some ways of configuring Kubernetes that we don't use and thus Phalanx doesn't support). I think this might be left over from the Helm starter.

services/squareone/templates/configmap.yaml
```diff
@@ -0,0 +1,16 @@
...
+    {{- include "squareone.labels" . | nindent 4 }}
+data:
+  squareone.config.yaml: |
+    siteName: {{ .Values.config.siteName | quote }}
+    baseUrl: {{ .Values.global.baseUrl | quote }}
+    siteDescription: |
+      {{ .Values.config.siteDescription }}
+    {{- if .Values.config.semaphoreUrl }}
+    semaphoreUrl: |
+      {{ .Values.config.semaphoreUrl }}
```
rra, line 15: If the description contains a newline without the right indentation, I think this may result in invalid YAML. I would tend to just use `| quote` even though this may be a long string and let Helm deal with any special characters, which also has the benefit of making all these settings uniform.

services/squareone/templates/deployment.yaml
```diff
@@ -0,0 +1,82 @@
...
+        {{- with .Values.podAnnotations }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      labels:
+        {{- include "squareone.selectorLabels" . | nindent 8 }}
+    spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
```
rra, line 27: I generally remove the conditional and just hard-code the name `pull-secret` since we never change it.

services/squareone/templates/deployment.yaml
```diff
@@ -0,0 +1,82 @@
...
+      serviceAccountName: {{ include "squareone.serviceAccountName" . }}
```
rra, line 28: If you don't need to make Kubernetes calls, this should instead be `automountServiceAccountToken`.

services/squareone/templates/ingress.yaml
```diff
@@ -0,0 +1,37 @@
+{{- if .Values.ingress.enabled -}}
+{{- $fullName := include "squareone.fullname" . -}}
+{{- $svcPort := .Values.service.port -}}
```
rra, line 3: Consider just hard-coding the service port. I don't think we'd ever change it.

services/squareone/templates/ingress.yaml
```diff
@@ -0,0 +1,37 @@
...
+  {{- if .Values.ingress.tls }}
+  tls:
+    {{- range .Values.ingress.tls }}
+    - hosts:
+        {{- range .hosts }}
+        - {{ . | quote }}
+        {{- end }}
+      secretName: {{ .secretName }}
+    {{- end }}
+  {{- end }}
```
rra, line 25: I think what I'd do here is have a simple boolean in `values.yaml` that's `ingress.tls` that you can set to true or false, and if it's set to true, insert the correct TLS stanza for `.Values.global.host`, hard-coding the secret name that we currently set in the values files. That minimizes the amount of stuff you have to put into the per-environment values files.

If that sounds like a good idea, I'd also add the required annotation if `ingress.tls` is set to true, since we always use the same cert-issuer name, and then you can remove that from values as well.

services/squareone/templates/service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "squareone.fullname" . }}
+  labels:
+    {{- include "squareone.labels" . | nindent 4 }}
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+    - port: {{ .Values.service.port }}
```
rra, line 10: Here too I'd just hard-code the service type and port. We're always going to use `ClusterIP` and won't need to customize the port number.

services/squareone/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.serviceAccount.create -}}
```
rra, line 1: As mentioned, I think you can probably delete this file.

services/squareone/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
rra, line 1: I suspect this won't ever do anything given our current setup, although no objections to leaving it if you think it might be useful some day.

services/squareone/values.yaml
```diff
@@ -0,0 +1,105 @@
+# Default values for squareone.
+#
+# Global parameters will be set by parameters injected by Argo CD and should
+# not be set in the individual environment values files.
+global:
```
rra, line 5: I generally move these to the bottom of values.yaml because people never need to set them or see them.

services/squareone/values.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+  # -- Image pull policy (tip: use Always for development)
+  pullPolicy: IfNotPresent
+
+  # -- Overrides the image tag.
+  # @default -- Chart's appVersion
+  tag: ""
+
+# -- Secret names to use for all Docker pulls
+imagePullSecrets: []
```
rra, line 33: Can be deleted if my suggestions above sound good.

services/squareone/values.yaml
```diff
@@ -0,0 +1,105 @@
...
+fullnameOverride: ""
+
+serviceAccount:
+  # -- Specifies whether a service account should be created
+  create: true
+  # -- Annotations to add to the service account
+  annotations: {}
+  # -- The name of the service account to use.
+  # @default -- Generated using the fullname template
+  name: ""
```
rra, line 48: Can be deleted if my suggestions above sound good.

services/squareone/values.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+# -- Annotations for squareone pods
+podAnnotations: {}
+
+service:
+  # -- Type of service to create
+  type: ClusterIP
+
+  # -- Port of the service to create and map to the ingress
+  port: 80
```
rra, line 58: Can be deleted if my suggestions above sound good.

services/squareone/values.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+ingress:
+  # -- Enable ingress
+  enabled: true
+
+  # -- Additional annotations to add to the ingress
+  annotations: {}
+
+  tls: []
```
rra, line 67: This would become just `tls: false` (or true, whatever makes more sense as a default).

rra (approved):
I think we still need a pull secret for GitHub, don't we? At least, there is an authentication credential for GitHub in our current pull secret.

Lots of other comments but they're all optional; I think this will work as-is. There are just some more things that you could delete to simplify the configuration.

jonathansick:
Thanks for the pointers for further simplifying the chart, that's exactly what I was hoping for.

> I think we still need a pull secret for GitHub, don't we? At least, there is an authentication credential for GitHub in our current pull secret.

@rra Public images on ghcr.io don't need a pull secret. However, are we thinking that eventually GitHub will institute unauthenticated pull rate limits akin to Docker Hub so we might as well get ahead of the curve?

PR #1330, opened 2022-04-07
[DM-34317] Add sherlock secrets, fix vo-cutouts

cbanek:
So the vocutouts thing had a keyerror exception when running it,
because postgres was generating it one way, and then we tried pulling
out the wrong key.  So that's the vo-cutouts fix.

Now that we have that working, add the sherlock generated secret
for publishing.


cbanek:
Just making sure I'm not missing something obvious on the vocutouts secret

rra (approved):
Whoops, sorry, thanks for fixing that.

PR #1334, opened 2022-04-08
tickets/DM-34282: reconfigure telegraf-ds configs

athornton:
Also turn off telegraf/telegraf-ds at several sites


afausti (commented):
This looks really great, I'm happy to see the app metrics going to their own buckets. I think the mapping bucket -> namespace makes a lot of sense... 

Good job getting the Prometheus metrics from multiple sources going to individual measurements with the override plugin. And the script that generates the TOML configuration seems very useful too. 

I would like to propose a simplification though. We are very close to have one bucket per namespace (application). The current organization with the `k8s_` and `prometheus_` buckets reflects the fact that we are running telegraf-ds and telegraf - but that's kind of artificial. 

What about having one bucket per application namespace and send both k8s and prometheus metrics to the same bucket? I think that's a nice simplification, everything we want for app X can be found on bucket X.

athornton:
I like that idea.  I'll retool everything to do that, and then each bucket is just the app name?

afausti:
Exactly! This way we have an app-centric organization in monitoring too, which is how our minds are used to thinking...

afausti (approved):
Looks good! Now every application has its own bucket and we are sending both k8s and Prometheus metrics to it.

PR #1351, opened 2022-04-12
tickets/DM-34422: consolidate values files

athornton:
Rationalize/DRY config and enable metric collection for supported services


athornton:
Too big, too confusing.  Breaking up and doing in stages.

PR #1364, opened 2022-04-14
initial commit of cachemachine gar doc

dspeck1:


athornton (changes requested):
The content is fine--no objections there.

My inclination would be to put this in "ops" rather than "service-guide" because it's really only useful if you're running an instance inside Google.  Maybe relocate to "ops/cachemachine" and then add a reference to the new doc from index.rst inside there (probably under the Guides rubric).

rra (approved):

PR #1374, opened 2022-04-20
Bumping portal container to a newer image

frossie:
This will remain on branch, just testing.


rra:
Old example and now obsolete, closing.

PR #1379, opened 2022-04-21
Added a new suit deployment, fixed the proxy in ingress.yaml

robyww:
- updates to suit version 2022.1
  - _fixes_:  when firefly is proxied, we need to know where it’s proxied from


rra (approved):
Thanks for tracking this down!

PR #1389, opened 2022-04-26
tickets/DM-34540: Send list of enabled services to telegraf/ds

athornton:


athornton:
Oh well.  The generated config looked fine enough to me, but the error is Helm Obtuseness for "hey you generated a resource that isn't a K8s resource".  This is part of the overall trouble I'm having with when YAML is an object and when it's a string.

I'll try replacing the string-that-looks-like-a-YAML-list with a straight-up comma-separated string, and see about unpacking it on the telegraf/ds side.

athornton:
Oh that's interesting.  Something that differs between idfdev and minikube is what's causing it to choke.

athornton:
minikube / roe / [squash-sandbox] / stable / tucson-teststand

That's a very strange list.

afausti (approved):
Looks good to me, a global variable with enabled services is useful to have.

PR #1391, opened 2022-04-26
tickets/DM-34540: Generate telegraf/ds configuration via helm rather than externally

athornton:


athornton:
A little commentary to aid in understanding this change:

I removed all the individual environment config files from each of telegraf and telegraf-ds.  

Then I created a new template, for a configmap, which is then used as the telegraf configuration file.  That template loops over enabled applications at a given instance (and in the telegraf case, over all Prometheus-enabled apps, selecting only the enabled ones) and builds input and output configurations for each app.

The list of enabled applications is provided as a global variable, as is (now) the instance hostname and the vault-secrets path.

Currently Prometheus endpoints are provided in the telegraf values.yaml file and so would need to be manually updated as we added more prometheus-enabled services.  An operator-based approach (perhaps telegraf-operator itself) that relied on pod or deployment annotations to know to harvest metrics would be more elegant.

afausti (approved):
This looks really good! Good job in replacing the python generator code. I like the approach of using a config map and Helm to template it. I have done this a bit in `kafka-connect-manager`. It looks much cleaner now in my opinion.

PR #1392, opened 2022-04-27
[DM-34430] Add test CSC and Kafka producer in sasquatch

afausti:
This PR adds the Test CSC and the Kafka Producer charts from the T&S charts repository as dependencies. The goal is to have data being produced to test new deployments of sasquatch.

Both charts are deployed in the sasquatch namespace.

It assumes the secret to pulling images from `ts-dockerhub.lsst.org` is in Vault under the `pull-secret` key for the deployment environment, and it creates the k8s secret expected by the charts in the sasquatch namespace.

### DIFF
--- a/services/sasquatch/Chart.yaml
+++ b/services/sasquatch/Chart.yaml
@@ -22,3 +22,9 @@ dependencies:
   - name: telegraf
     version: 1.8.18
     repository: https://helm.influxdata.com/
+  - name: csc
+    version: 0.9.2
+    repository: https://lsst-ts.github.io/charts/
+  - name: kafka-producers
+    version: 0.10.1
+    repository: https://lsst-ts.github.io/charts/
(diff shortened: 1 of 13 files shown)


athornton:
The only problem I have with it is that every time I try to update the pull-secret with a new entry, it turns into a horrendous nightmare.  I don't have a problem with putting it everywhere except for that, but I'd ask Michael Reuter before doing it, anyway; he may have some good reason he doesn't want that credential available at non T&S sites.

afausti:
@mareuter would you have a look at this PR? Any objections to deploying the Test CSC and Kafka Producer for example in Google?

mareuter:
You'll have to forgive me but I've always been under the impression that cloud providers don't support multicast communication which our CSCs rely on. I'm OK with you using the configuration as long as you're only planning to use this for non T&S sites since we already have those systems running and available for you to use there.

afausti:
Thanks, @mreuter! my goal is to produce some data to test Sasquatch during development - for example, on our development cluster on Google. I completely forgot about the multicat thing but I'll see what happens in Google. This is still useful for me in the Teststand environments but it is missing the option to enable/disable this since it is not supposed to run in production mode.

PR #1395, opened 2022-04-28
DM-34595: Add pre-commit configuration for helm-docs and black code formatting

jonathansick:
This adds pre-commit to help us maintain code and documentation quality. Like most of our apps and libraries, the pre-commit hooks can be installed for individual developers via a `make init` command. Pre-commit is also run as part of our CI.

Pre-commit hooks include:

- trailing-whitespace (part of the default hooks)
- helm-docs (along with a default template that all apps can use)
- black
- isort
- flake8
- blacken-docs (black but for rst code embeds)

To conform with these hooks, we've updated the Phalanx "codebase", particularly adding and updating READMEs for all Helm charts and reformatting our Python code


rra (approved):
Looks good to me! Thank you for doing this!

PR #1425, opened 2022-05-12
Bump non-prod JH versions

athornton:


athornton:
Closing in favor of bumping value in Chart.yaml

PR #1436, opened 2022-05-14
Hotfix/dry out argocd

athornton:


athornton:
Still not properly doing globalv injection.  Give up on that for now.

PR #1440, opened 2022-05-17
[DM-34802] Add HiPS service and starter

rra:
Add a chart for the HiPS service using crawlspace as the underlying
container implementation.  Enable it only on IDF dev.

Add a starter for web services based on the charts starter, but with
additional simplifications.

### DIFF
--- a/starters/web-service/templates/ingress.yaml
+++ b/starters/web-service/templates/ingress.yaml
@@ -1,0 +1,29 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: "<CHARTNAME>"
+  labels:
+    {{- include "<CHARTNAME>.labels" . | nindent 4 }}
+  annotations:
+    kubernetes.io/ingress.class: "nginx"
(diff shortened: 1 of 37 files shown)


starters/web-service/templates/ingress.yaml
```diff
@@ -0,0 +1,29 @@
...
+    {{- end }}
+    {{- with .Values.ingress.annotations }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  rules:
+    - host: {{ required "global.host must be set" .Values.global.host | quote }}
+      http:
+        paths:
+          - path: "/<CHARTNAME>"
```
jonathansick, line 23: Something I've found very helpful is to do:

```yaml
- path: {{ .Values.ingress.path }}
```

That way I can also include ``Values.ingress.path`` in my app's config map to mount it on a consistent path.

For example, Times Square is at `/times-square/api`, which isn't its chart name, so that I can let `/times-square` be a URL served by Squareone.

PR #1443, opened 2022-05-18
DM-34865: Deploy noteburst

jonathansick:
Deploys noteburst from https://github.com/lsst-sqre/noteburst/pull/32


jonathansick:
Moved to #1458

PR #1444, opened 2022-05-19
Fix syntax errors in imagePullSecrets

rra:
In a couple of charts, this wasn't an array.  Fix it to always be
an array and use consistent quoting.


r-owen (commented):
Thank you for the fix. A small suggestion to lose the double-quotes. They should not be needed (yaml is pretty good about figuring out what a string looks like, at least as long as it doesn't contain a colon).

r-owen (approved):

PR #1445, opened 2022-05-19
tickets/DM-34835

athornton:


athornton:
I'm going to self-merge, since it's a version bump and turning some Helm YAML comments into helm-doc-readable comments.

PR #1449, opened 2022-05-20
Document pull secret updating

athornton:


docs/service-guide/update-pull-secret.rst
```diff
@@ -0,0 +1,43 @@
...
+this::
+
+  { ".dockerconfigjson": "{\"auths\": {\"ghcr.io\": {\"auth\": \"base64string\", \"password\": \"cleartexttoken\",\"username\": \"token\"},\"index.docker.io\": {\"auth\":\"base64string\",\"password\":\"cleartextpassword\",\"username\":\"sqrereadonly\"}}}}
+
+In short: the value is a *string* (not a JSON object) with all keys and
+values quoted with backslash-escaped double quotes.
+
+Once you have this file created, run:
+
+``vault kv put secret/k8s_operator/<environment>/pull-secret @pull-secret.json``
```
rra, line 33: I would not modify secrets this way, since it requires this extra level of encoding and JSON. The way I always update secrets in Vault (not just this secret, every secret), is:

#. Highlight the correct secret value and copy it to my paste buffer.
#. Start with `vault kv patch secret/k8s_operator/<environment>/pull-secret .dockerconfigjson='` (note the single quote)
#. Paste the secret into the command line
#. Type `'` and press Enter

I feel like most of the problems we've had with Vault secret management is using ways of updating or viewing secrets that introduce another level of JSON, either using JSON output mode rather than the default or using this syntax to modify the Vault secret that takes JSON input. I would just avoid all use of JSON with Vault; it seems to cause problems.

athornton: I've updated the doc with your technique, which I agree is better than mine.

rra (approved):

PR #1459, opened 2022-05-24
tickets/DM-34944: Add kube_inventory input

athornton:


afausti (approved):
Perfect I see the pod restart metrics already in data-dev

PR #1463, opened 2022-05-26
Change to recommended_int

athornton:


services/cachemachine/values-idfint.yaml
```diff
@@ -25,6 +25,7 @@ autostart:
           "type": "SimpleRepoMan",
           "images": [
             {
+              "image_url": "us-central1-docker.pkg.dev/rubin-shared-services-71ec/sciplat/sciplat-lab:w_2022_22",
```
SimonKrughoff, line 28: What does adding this line do in this case?  Is it needed in general?

SimonKrughoff (approved):
One question about the SimpleRepoMan config

athornton:
I know that I am keeping w_12 in cache too.  That's intentional: if users don't want the new recommended I don't want them to have to wait for the pull.

PR #1464, opened 2022-05-27
[DM-34998] Inject butlerRepositoryIndex

rra:
Both vo-cutouts and datalinker need DAF_BUTLER_REPOSITORY_INDEX to
resolve Butler locations.  This is a per-site configuration that
should also eventually be injected into nublado2, so add it to the
deployment values file and inject it via Argo CD.  Adjust datalinker
and vo-cutouts charts accordingly, and set the environment variable
on the relevant deployments.

For vo-cutouts, we can now drop CUTOUT_BUTLER_REPOSITORY, since this
replaces that setting.

### DIFF
--- a/science-platform/templates/datalinker-application.yaml
+++ b/science-platform/templates/datalinker-application.yaml
@@ -2,33 +2,35 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: datalinker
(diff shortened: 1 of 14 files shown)


cbanek (approved):
Looks great, thanks!  Love this global thing.

PR #1499, opened 2022-06-10
[DM-34613] Update for latest Gafaelfawr

rra:
- Bump version of Gafaelfawr and run it from a branch on dev and minikube for now
- Update Gafaelfawr chart for the new LDAP configuration structure
- Add `ldap-password` secret for Gafaelfawr
- Add LDAP and Firestore configuration for Gafaelfawr on IDF dev
- Stop specifying UIDs for mobu on data-dev
- Rename all mobu users to start with `bot-` and be more informative about what flock they're running
- Rename all other Gafaelfawr service users to start with `bot-`


installer/generate_secrets.py
```diff
@@ -221,6 +221,11 @@ def _gafaelfawr(self):
                 f"Invalid gafaelfawr cloudsql value {use_cloudsql}"
             )
 
+        self.input_field("gafaelfawr", "ldap", "Use LDAP? (y/n):")
```
athornton, line 224: Should the use-cilogon-or-github go in the else clause, or do we still need it even if we're using LDAP?

athornton (approved):
Other than probably-my-lack-of-understanding with my comment about whether we still need cilogon-or-github, this looks good.

installer/generate_secrets.py
rra: They're independent; you can use any authentication source and then still use LDAP for data (although in practice we will probably only use LDAP for the CILogon and OpenID Connect cases).

rra:
Yes, indeed, thank you!

PR #1565, opened 2022-07-06
Dm-35468 trey update suit

robyww:
- Update SUIT to support firefly 2022.2-PRE-3
- Update deployment.yaml with example support PROPS_FIREFLY_OPTIONS

_note:_ - I tried to make the work with the values architecture but it seems to have trouble with block scalars so I can figure it out, but this would be the next step

### DIFF
--- a/services/portal/README.md
+++ b/services/portal/README.md
@@ -11,6 +11,7 @@ Rubin Science Platform portal aspect
 | affinity | object | `{}` | Affinity rules for the Portal pod |
 | config.cleanupInterval | string | `"36h"` | How long results should be retained before being deleted |
 | config.debug | string | `"FALSE"` | Set to `TRUE` to enable service debugging |
+| config.hipsUrl | string | `/api/hips/images/color_gri` in the local Science Platform | URL for default HiPS service |
 | config.visualizeFitsSearchPath | string | `"/datasets"` | Search path for FITS files |
 | config.volumes.configHostPath | string | Use an `emptyDir` | hostPath to mount as configuration.  Set either this of `configNfs`, not both. |
 | config.volumes.configNfs | object | Use an `emptyDir` | NFS information for a configuration.  If set, must have keys for path and server, Set either this of `configHostPath`, not both. |
--- a/services/portal/templates/deployment.yaml
+++ b/services/portal/templates/deployment.yaml
@@ -46,19 +46,19 @@ spec:
               value: {{ .Values.global.host | quote }}
             - name: "PROPS_FIREFLY_OPTIONS"
               value: >-
-                $'{ 
-                   "coverage":  {"hipsSourceURL" : "ivo://CDS/P/2MASS/color"},  
-                   "tap" : { 
-                      "additional": { 
-                          "services": [ { 
-                               "label": "LSST RSP", 
-                               "value": "https://data-dev.lsst.cloud/api/tap", 
-                               "hipsUrl": "https://irsa.ipac.caltech.edu/data/hips/CDS/DSS2/color/", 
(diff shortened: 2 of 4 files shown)


rra (approved):
Looks good to me with the fixes I pushed.

services/portal/values-idfdev.yaml
```diff
@@ -4,7 +4,11 @@ resources:
   limits:
     memory: "2Gi"
 
+image:
+  tag: "suit-2022.3"
```
gpdf, line 8: I think this tag has already been used.  This should be advanced to 2022.4.

services/portal/values-idfdev.yaml
```diff
@@ -4,7 +4,11 @@ resources:
...
+
 config:
+  hipsUrl: "https://irsa.ipac.caltech.edu/data/hips/CDS/DSS2/color/"
```
gpdf, line 11: Why is DSS2/color used here (in values-idfdev.html)?  Is a local HiPS not available on data-dev?

services/portal/values-idfdev.yaml
gpdf: Also, the internal version reporting in `app.config` needs to be updated accordingly.

services/portal/templates/deployment.yaml
```diff
@@ -44,6 +44,28 @@ spec:
               value: {{ include "portal.fullname" . }}-redis
             - name: "PROPS_sso__req__auth__hosts"
               value: {{ .Values.global.host | quote }}
+            - name: "PROPS_lsst__hips__masterUrl"
+              value: https://irsa.ipac.caltech.edu/data/hips/list
+            - name: "PROPS_FIREFLY_OPTIONS"
+              value: >-
+                $'{
+                   "coverage":  {"hipsSourceURL" : "ivo://CDS/P/2MASS/color"},
```
gpdf, line 52: This needs to be the Rubin `color_gri` now, right?

gpdf (commented):
I haven't fully understood the layering of the HiPS configuration options; perhaps we need to walk through that together?

services/portal/templates/deployment.yaml
robyww: Yes, but I could not test it because I don't think it is available on dev

services/portal/README.md
```diff
@@ -11,6 +11,7 @@ Rubin Science Platform portal aspect
 | affinity | object | `{}` | Affinity rules for the Portal pod |
 | config.cleanupInterval | string | `"36h"` | How long results should be retained before being deleted |
 | config.debug | string | `"FALSE"` | Set to `TRUE` to enable service debugging |
+| config.hipsUrl | string | `/api/hips/images/color_gri` in the local Science Platform | URL for default HiPS service |
```
gpdf, line 14: Has it been possible to test that `/api/hips/images/color_gri` actually works now - i.e., that the auth token is being passed along correctly?

services/portal/values-idfdev.yaml
robyww: yes, it is not available.

services/portal/values-idfdev.yaml
robyww: yes, we can make it 2022.4

services/portal/templates/deployment.yaml
robyww: also we will have add more to the values table.

services/portal/values-idfdev.yaml
gpdf: See https://github.com/lsst/suit/pull/41 - it would also be OK for the new version to be 2022.3.1, but we should

PR #1568, opened 2022-07-06
Add persistence to gafaelfawr redis at T&S sites

athornton:


rra (approved):
Thank you!

afausti:
Also, these PVs are created with the reclaim policy "Delete" (which means the volume is automatically deleted when the PVC is deleted). 

In some cases, it might be appropriate to change the reclaim policy to "Retain". 

`kubectl patch pv <your-pv-name> -p '{"spec":{"persistentVolumeReclaimPolicy":"Retain"}}'`

PR #1571, opened 2022-07-07
[DM-35495] Protect Portal admin URLs with Gafaelfawr

rra:
Disable Portal's internal admin authentication and instead protect
the admin endpoints with Gafaelfawr using a different scope
(exec:admin).  This will work with Portal 2022.4 or later, which
is still being deployed, but should be harmless to apply to earlier
versions (just not effective).


robyww (approved):
From the firefly side (`deployment.yaml`) it looks right. I can't speak to the ingress side.

athornton (approved):
Looks like standard ingress boilerplate and a switch to control deployment.  Seems fine.

PR #1587, opened 2022-07-14
Enable telegraf and telegraf-ds at TTS

afausti:
- Now that Sasquatch is running at TTS we can tear down the old EFD deployment and deploy telegraf and telegraf-ds as part of the monitoring infrastructure.


athornton (approved):
LGTM

PR #1595, opened 2022-07-15
Fix plot-navigator.

ctslater:
Currently all requests return a 503 from nginx, and don't seem to ever reach panel. No messages in the logs.


ctslater:
Switching the service to use `selectorLabels` did the trick; I guess the deployment got DRY-ized without the service's selector.

PR #1640, opened 2022-08-03
Bump memory limits for strimzi-cluster-operator

rra:
It was crashing on the Tucson teststand at 384MiB with out of
memory errors.  Bump it to 512MiB everywhere.


afausti (approved):
LGTM

PR #1641, opened 2022-08-04
[DM-35482] Serve /api/hips/list from datalinker

rra:
datalinker will be the service that dynamically generates the HiPS
list based on the properties files of the HiPS data sets for a given
Science Platform deployment.

### DIFF
--- a/services/datalinker/templates/gafaelfawr-token.yaml
+++ b/services/datalinker/templates/gafaelfawr-token.yaml
@@ -1,0 +1,10 @@
+apiVersion: gafaelfawr.lsst.io/v1alpha1
+kind: GafaelfawrServiceToken
+metadata:
+  name: {{ include "datalinker.fullname" . }}-gafaelfawr-token
+  labels:
+    {{- include "datalinker.labels" . | nindent 4 }}
+spec:
+  service: "bot-datalinker"
+  scopes:
+    - "read:image"
--- a/services/datalinker/templates/ingress-anonymous.yaml
+++ b/services/datalinker/templates/ingress-anonymous.yaml
@@ -1,16 +1,10 @@
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
-  name: {{ include "datalinker.fullname" . }}
+  name: {{ include "datalinker.fullname" . }}-anonymous
   labels:
     {{- include "datalinker.labels" . | nindent 4 }}
   annotations:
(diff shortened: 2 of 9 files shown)


services/datalinker/templates/gafaelfawr-token.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: gafaelfawr.lsst.io/v1alpha1
+kind: GafaelfawrServiceToken
```
cbanek, line 2: Oooo.

services/datalinker/templates/ingress-anonymous.yaml
```diff
@@ -0,0 +1,23 @@
...
+    {{- with .Values.ingress.annotations }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  ingressClassName: "nginx"
+  rules:
+    - host: {{ required "global.host must be set" .Values.global.host | quote }}
+      http:
+        paths:
+          - path: "/api/hips/list"
```
cbanek, line 17: This is a completely random comment that isn't really about this, but is only tangentially related.  If we mess up the ingress and go to say /api/hipz, then we don't get routed through this ingress rule (obviously) but also we end up getting caught by the homepage I think.  This can make things confusing with apis and command line clients when they get a good helping of javascript with a 200.  But I'm not sure how to fix that, and this is all fine.

cbanek: It'd also be good if we could say for all of /api if you aren't authenticated to give back a 403 rather than an auth redirect too - we always forget that one, and that also is another common confuzzlement I have

cbanek (approved):

services/datalinker/templates/ingress-anonymous.yaml
rra: Yeah, agreed that we need somewhat better handling of 404s. I think that's something squareone should handle as part of being the default service. For the record, though, it does return a 404, not a 200, confirmed with `curl` just now (but yes, the page is full of JavaScript).

I had intentionally made `/api/hips` support the login redirect because unlike most API services I thought people were going to look at it in a web browser, but so far we've chosen not to publish an `index.html` and are using the Portal as the HiPS viewer instead of an embedded Aladin Lite instance. So I'll remove the redirect configuration for now.

PR #1642, opened 2022-08-05
[DM-35849] Exclude Kafka PVCs from the app sync status

afausti:
In Strimzi operator 0.30.0 Kafka `PersistenVolumeClaim` resources show up as `OutOfSync (requires prune)` in Argo CD, affecting the overall app sync status. 

Strimzi provides a mechanism for [customizing k8s resources](https://strimzi.io/docs/operators/latest/configuring.html#assembly-customizing-kubernetes-resources-str).  Use that to add annotations to [exclude the PVCs from the app sync status](https://argo-cd.readthedocs.io/en/stable/user-guide/compare-options/#ignoring-resources-that-are-extraneous) and to [prevent them from being pruned](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#no-prune-resources).

### DIFF
--- a/services/sasquatch/charts/strimzi-kafka/templates/kafka.yaml
+++ b/services/sasquatch/charts/strimzi-kafka/templates/kafka.yaml
@@ -5,6 +5,12 @@ metadata:
   name: {{ .Values.cluster.name }}
 spec:
   kafka:
+    template:
+      persistentVolumeClaim:
+        metadata:
+          annotations:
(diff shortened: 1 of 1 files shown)


rra:
Change looks good to me. Note that from experience with the nublado2 configuration, this will keep the out of date state from propagating to the application, but Argo CD will still show the PVCs themselves as out of date and requiring pruning. There doesn't seem to be a way to suppress that entirely, unfortunately.

rra (approved):

PR #1677, opened 2022-08-24
[DM-35977] Use ingressClassName instead of an annotation

rra:
The current Ingress API supports ingressClassName to configure the
name of the ingress controller that should serve this ingress.  Use
this instead of the older annotation to configure all ingresses to
use nginx.

Drop an old tls configuration for cachemachine. We only do tls via
squareone.

Hide another Gafaelfawr route only used for OpenID Connect if OpenID
Connect is not enabled.


afausti (approved):
Reviewed sasquatch values and template files lgtm.

PR #1694, opened 2022-08-31
 [DM-36074] Fix template for S3 

gabrimaine:
The secrets are not set correctly when S3 is used instead of GCS


rra:
Whoops, sorry, got `S3` right in one place but not in another place. Thanks for the fix! I agree that `S3` is better since it may not be actual AWS.

rra (approved):

PR #1701, opened 2022-09-02
sqlproxy butler int

dspeck1:
Add sqlproxy as a standalone service.  Included in this PR is helm chart for the service and the argo CD configuration to deploy as an app.  Tried to make the sqlproxy service as generic so it can be used again.  The argo CD app is specific for the butler-int integration.

### DIFF
--- a/science-platform/templates/sqlproxy-butler-int-application.yaml
+++ b/science-platform/templates/sqlproxy-butler-int-application.yaml
@@ -1,0 +1,33 @@
+{{- if .Values.sqlproxy_butler_int.enabled -}}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: sqlproxy-butler-int
+spec:
+  finalizers:
+    - kubernetes
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: sqlproxy-butler-int
+  namespace: argocd
+  finalizers:
+    - resources-finalizer.argocd.argoproj.io
+spec:
+  destination:
+    namespace: sqlproxy-butler-int
+    server: https://kubernetes.default.svc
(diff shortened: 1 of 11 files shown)


athornton (commented):
Why are we introducing a new application with "-int" in the name, and then in services, it's got "gcp" in the name?  Should it not have either of those and only be enabled in the int instance?

dspeck1:
@athornton - adjusted naming.  Please review again.  Thanks!

science-platform/README.md
```diff
@@ -27,6 +27,7 @@
 | sasquatch.enabled | bool | `false` |  |
 | semaphore.enabled | bool | `false` |  |
 | sherlock.enabled | bool | `false` |  |
+| sqlproxy_butler_int.enabled | bool | `false` |  |
```
athornton, line 30: Should this not be sqlproxy_cross_project.enabled?

science-platform/templates/sqlproxy-cross-project-application.yaml
```diff
@@ -0,0 +1,33 @@
+{{- if .Values.sqlproxy_butler_int.enabled -}}
```
athornton, line 1: Here too?

science-platform/values.yaml
```diff
@@ -38,6 +38,8 @@ semaphore:
   enabled: false
 sherlock:
   enabled: false
+sqlproxy_butler_int:
```
athornton, line 41: Here too?

athornton (changes requested):
Either you missed a couple of nomenclature updates, or the nomenclature is very confusing.

athornton (approved):
Looks good to me.

PR #1707, opened 2022-09-06
[DM-36093] Update nublado2 templates for latest Gafaelfawr

rra:
- Update to latest nublado2 release
- Expose the user's primary GID in `EXTERNAL_GID` parallel to `EXTERNAL_UID`
- Drop the provisionator user and group, which I believe are no longer used
- Don't add the user as an additional member of the group matching their primary GID, following normal UNIX semantics
- Set the user's GID correctly if it differs from their UID


athornton (approved):
LGTM

PR #1708, opened 2022-09-06
Remove  requirement from summit cachemachine

athornton:


afausti (approved):
LGTM if we don't need that anymore.

PR #1715, opened 2022-09-08
Missing slack-webhook value causes generate_secrets script to fail

stvoutsin:
**Description**:
When the `slack-webhook` key is missing from the `secrets[rsp-alerts]` dictionary, the generate_secrets scripts fails at the following line, where it can not find that key in the dict:
https://github.com/lsst-sqre/phalanx/blob/master/installer/generate_secrets.py#L258

**Fix**:
Add a check to see that the key exists, and only then set the `secrets[gafaelfawr][slack_webhook] `value

If there are other better ways to fix this, or if this works as intended please ignore/close

### DIFF
--- a/installer/generate_secrets.py
+++ b/installer/generate_secrets.py
@@ -255,7 +255,10 @@ class SecretGenerator:
         else:
             raise Exception(f"Invalid auth provider {auth_type}")
             
-        if  self.secrets.get("rsp-alerts", {}).get("slack-webhook", None) != None:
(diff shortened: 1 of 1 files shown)


rra:
Thanks for catching this. The data structure didn't work the way that I had thought it did. I fixed this in a slightly different way in https://github.com/lsst-sqre/phalanx/pull/1717 which I think will work, but let me know if that still has problems.

PR #1736, opened 2022-09-20
DM-36302: Delete /etc/shadow and /etc/gshadow for labs

rra:
We no longer support PAM, sudo, or privilege escalation in labs, so nothing should be reading these files.  Simplify the config by removing the generated ConfigMaps and mounts.

Also delete obsolete settings for vault_secrets_path and pull-secret, which were migrated to use the injected global path.

The sciplat-lab build already deletes /etc/shadow and /etc/gshadow, so with this change those files simply won't exist.


athornton (approved):
This seems fine, but I have a vague memory of stubbing my toes on vaultSecretPath or at least something associated with making the globals work with respect to nublado2.  But if this works, great!

rra:
I'll give it a shot and see if something breaks, and if so, I'll iterate a bit. Thanks!

PR #1748, opened 2022-09-30
Deploy latest Firefly to production

gpdf:
This is the update that had been planned for yesterday, 2022-09-29, and discussed in RSP Ops.


athornton (approved):
LGTM

gpdf:
Merged based on @athornton review.

PR #1749, opened 2022-09-30
[DM-36233] Evaluate Telegraf as replacement for the InlfuxDB Sink connector in Sasquatch

afausti:
- Add a subchart to deploy a list of Telegraf Kafka consumer agents to consumer Kafka topics and write them to InfluxDB
- Enable the Telegraf Kafka consumers and deploy on TTS to compare performance with the Lenses InfluxDB Sink Kafka connector.

### DIFF
--- a/services/sasquatch/values-idfdev.yaml
+++ b/services/sasquatch/values-idfdev.yaml
@@ -20,12 +20,19 @@ influxdb:
     enabled: true
     hostname: data-dev.lsst.cloud
 
-kafka-connect-manager:
-  influxdbSink:
-    connectors:
-      test:
-        enabled: true
-        topicsRegex: ".*Test"
+telegraf-kafka-consumer:
+  kafkaConsumers:
+    test:
+      enabled: true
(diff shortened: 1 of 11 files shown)


services/sasquatch/values-idfdev.yaml
```diff
@@ -20,12 +20,19 @@ influxdb:
...
+    test:
+      enabled: true
+      replicaCount: 1
+      topicRegexps: |
+        [ ".*Test" ]
+    atmcs:
+      enabled: true
+      replicaCount: 1
+      topicRegexps: |
+        [ ".ATMCS" ]
```
athornton, line 34: Shouldn't this be ".*ATMCS" ?

athornton (approved):
Looks good, although with one minor question about one of the regexps.

services/sasquatch/values-idfdev.yaml
afausti: Yes, thanks!

PR #1756, opened 2022-10-03
tickets/SITCOM-445

mareuter:
TTS: Add and fix generic camera related sasquatch items.


athornton (commented):
I don't understand why we want GenericCamera handled by both obssys and genericcamera.  Other than that it seems fine to me.

mareuter:
obssys references were removed in this commit.

PR #1758, opened 2022-10-03
Move TTS to external hosted Postgres

athornton:


services/exposurelog/templates/deployment.yaml
```diff
@@ -56,18 +56,18 @@ spec:
...
-            - name: EXPOSURELOG_DB_PORT
-              value: "5432"
-            - name: EXPOSURELOG_DB_DATABSE
-              value: exposurelog
+            - name: NARRATIVELOG_DB_HOST
+              value: {{ .Values.db.host | quote }}
+            - name: NARRATIVELOG_DB_PORT
+              value: {{ .Values.db.port | quote }}
+            - name: NARRATIVELOG_DB_DATABSE
+              value: {{ .Values.db.database | quote }}
```
rra, line 70: The names here are different; was that intentional?

services/exposurelog/values.yaml
```diff
@@ -20,6 +20,17 @@ image:
   # @default -- The appVersion of the chart
   tag: ""
 
+# -- Postgres database connection information
```
rra, line 23: You want to omit the comment here since right now it's documenting both `db` and all its child settings.

services/narrativelog/values.yaml
```diff
@@ -20,6 +20,17 @@ image:
   # @default -- The appVersion of the chart
   tag: ""
 
+# -- Postgres database connection information
```
rra, line 23: Likewise here on the comment.

services/exposurelog/templates/deployment.yaml
athornton: I don't know why Russell left out the last `A` in Database for the env var.  So, yeah, because I couldn't bring myself to put a `databse` configuration setting.

athornton:
(I'm going to wait on disabling internal Postgres until we know this works--that'll be a separate PR)

services/exposurelog/templates/deployment.yaml
athornton: Derp.  That musta been on me.  As was all the confusion between exposure and narrativelog.

rra (approved):

PR #1768, opened 2022-10-05
Upgrade JupyterHub database at summit

rra:
Enable automatic upgrade of the JupyterHub database at the summit for the JupyterHub 3.0 migration.


athornton (approved):
LGTM

PR #1776, opened 2022-10-10
[DM-36429] Add InfluxDB v2 and bucketmapper to Sasquatch

afausti:


athornton (approved):
This seems fine to me.

services/sasquatch/values.yaml
```diff
@@ -70,6 +70,52 @@ influxdb:
...
+    # @default 1Ti for teststand deployments.
+    size: 1Ti
+  ingress:
+    # -- InfluxDB2 ingress configuration
+    enabled: false
+    hostname: ""
+    annotations:
+      nginx.ingress.kubernetes.io/rewrite-target: /$2
+    className: "nginx"
+    path: /influxdb2(/|$)(.*)
```
athornton, line 95: Not sure this is going to work.

afausti: Unfortunately, you can't reach the UI ([see issue #15721](https://github.com/influxdata/influxdb/issues/15721)], but you can reach the API at `/influxdb2/api/v2`. I think this is a good solution since we are using Chronograf with InfluxDB2.

afausti: Alternatively I could use `nginx.ingress.kubernetes.io/rewrite-target: /api/v2/$2` to rewrite `/influxdb2` to the API root, for example:

```
> GET /influxdb2/buckets HTTP/2
> Host: data-dev.lsst.cloud
{
	"links": {
		"self": "/api/v2/buckets?descending=false\u0026limit=20\u0026offset=0"
	},
	"buckets": []
```	

I think that's better.

PR #1778, opened 2022-10-12
DM-36585: Add a Python package to Phalanx

jonathansick:
This PR adds a standard Python package to phalanx for the purpose of hosting Python infrastructure for testing, developing, and generating documentation.

Since there is now Python package infrastructure, the docs are now built with tox. The `expand-services` script is also moved into the Python package and installed as an entrypoint along with its GitPython dependency.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@
 name = "phalanx"
 version = "1.0.0"
 description = "Python support code for the Rubin Phalanx platform."
-# license = {file = "LICENSE"}
+license = {file = "LICENSE"}
 readme= "README.rst"
 keywords = [
     "rubin",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,74 @@
+[project]
+# https://packaging.python.org/en/latest/specifications/declaring-project-metadata/
+name = "phalanx"
+version = "1.0.0"
+description = "Python support code for the Rubin Phalanx platform."
+# license = {file = "LICENSE"}
+readme= "README.rst"
+keywords = [
+    "rubin",
+    "lsst",
+]
+# https://pypi.org/classifiers/
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "License :: OSI Approved :: MIT License",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Intended Audience :: Developers",
+    "Operating System :: POSIX",
+]
+requires-python = ">=3.8"
+dependencies = []
+
+[project.optional-dependencies]
+dev = [
+    # Testing
+    "coverage[toml]",
+    "pytest",
+    "pre-commit",
+    "mypy",
+    # Documentation
+    "documenteer[guide]>=0.7.0b4",
+    "sphinx-diagrams"
+]
+
+[project.urls]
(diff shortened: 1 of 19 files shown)


jonathansick:
I'm not entirely sure why the minikube deployment is failing to sync, and whether that's related to these changes or not. Ideas?

src/phalanx/testing/expandcharts.py
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env python3
```
rra, line 1: I am mildly against putting test suite code into the actual Python package for something, so this part of the change feels weird to me and I'd rather keep things that are only useful for testing in the `tests` directory. I can't imagine anyone wanting to run that script manually; it's just a hack to make `helm lint` sort of work and there's probably some better way to do that. That said, just a mild objection, and more on a matter of aesthetics than any actual issue, so do feel free to decide you don't agree.

pyproject.toml
```diff
@@ -1,3 +1,79 @@
...
+    "License :: OSI Approved :: MIT License",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Intended Audience :: Developers",
+    "Operating System :: POSIX",
+]
+requires-python = ">=3.8"
+dependencies = [
+    "GitPython",
+]
```
rra, line 24: I think this should only be a dev dependency since it's only used by a test script and is not necessary to build the docs or install Phalanx, but this is probably just an aspect of my other comment.

rra (approved):
Thanks for fixing all the Gafaelfawr URLs!

I'm not sure what's up with the test failure. It looks like we successfully install Argo CD and it's working fine, but when we convert it to being self-managed, that breaks. Maybe it's because the conversion to self-managed restarts the Argo CD pod and that looks like a failure, but I'm not sure why it would suddenly start failing now and wasn't before. If it's just that the Argo CD pod was restarted, maybe we need to sync Argo CD once by itself and ignore failures and then sync all the apps and hope it works the second time?

src/phalanx/testing/expandcharts.py
jonathansick: Ah I see. My thinking was that the phalanx's python package job was to provide test, documentation build, and development templating support code for phalanx rather than "be phalanx", hence moving the expandcharts script into the phalanx package so that down the line we can use our standard infrastructure for testing that code and providing dependencies for it.

So this isn't test suite code for the phalanx python package, but functionality of the phalanx Python package.

Does that kind of make sense? If not I'm happy to revert.

rra: Oh! I see. That's consistent with the docs, yes. That makes more sense. Okay, I'm convinced.

I was thinking of the phalanx package as primarily being the installer, so had a more limited view of it.

I think we probably should think about whether we want to keep the GitPython dependency out of the main dependencies for the eventual use of the phalanx package as the installer for people who just want to install Phalanx and not test it, but we can also cross that bridge later, and use some other optional dependency name for it.

jonathansick: I'm glad we're on the same page. Yeah, we can be more specific with the dependencies... we can work that out when get cross that bridge with a more developer/user focus.

PR #1794, opened 2022-10-24
DM-36458: Update Gafaelfawr Helm config for Kopf

rra:
Update for Gafaelfawr 7.0.0, which now uses Kopf to implement the Kubernetes operator.  This changes the invocation of the token management pod and removes the status specification in the CRD. Instead, add the annotation that tells Kubernetes the status section is free-form.


services/gafaelfawr/values-idfdev.yaml
```diff
@@ -1,3 +1,7 @@
+image:
+  pullPolicy: "Always"
+  tag: "tickets-DM-36458"
```
athornton, line 3: Don't you want to change this prior to merge?

services/gafaelfawr/values-minikube.yaml
```diff
@@ -1,3 +1,7 @@
+image:
+  pullPolicy: "Always"
+  tag: "tickets-DM-36458"
```
athornton, line 3: Here too?

services/gafaelfawr/values-idfdev.yaml
athornton: ...oh, or are you going to merge, and then do a followup PR with the actual tagged build so it exists when you're ready to sync?  That's assiduous.

athornton (approved):
LGTM, modulo a little confusion on whether you're planning a followup PR or not.

services/gafaelfawr/values-idfdev.yaml
rra: The pinning is just there so that I can test on data-dev from the branch, and then I'll rebase to drop the pins (which are in a separate commit) before I merge the PR branch.

PR #1796, opened 2022-10-25
DM-36709: exposurelog: set PGUSER and PGPASSWORD env vars from the secrets vault

r-owen:


ktlim (changes requested):
See Jira ticket.

ktlim (approved):
OK, I guess setting both env vars means you don't have to change your code.

PR #1812, opened 2022-11-04
DM-36868: Clean up the sqlproxy service

rra:
Rename this to sqlproxy-cross-project to match the science-platform name and overall Phalanx configuration.  Add documentation.

Clean up the Helm chart configuration.  Remove a bunch of unneeded variables and make the interface higher-level, roughly matching how Gafaelfawr's instance of this has been configured.  Remove global variables that were never used.

Update the minikube and Kubernetes versions used for testing.


rra:
The docs build is failing because of a link to a page that is also created by this PR, so should resolve once it's merged.

science-platform/templates/sqlproxy-cross-project-application.yaml
```diff
@@ -2,31 +2,28 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: sqlproxy-cross-project
+  name: "sqlproxy-cross-project"
```
athornton, line 5: I don't understand why you're quoting all the strings in this file, when we don't elsewhere, but it seems unlikely to hurt anything.

athornton (approved):
Seems reasonable to me.

science-platform/templates/sqlproxy-cross-project-application.yaml
rra: I've been adding quoting to each file in Phalanx when I touch it because it means one doesn't have to remember YAML's quoting rules, but I admit that I'm inconsistent about it. (A bunch of the applications have already been converted to fully quoted, though, so this is making them more consistent.)

PR #1818, opened 2022-11-07
Deploy bucketmapper and telegraf-kafka-consumer only if influxdb2 is enabled

afausti:
These components only make sense in environments where influxdb2 is deployed, so make the environments cleaner while transitioning from influxdb to influxdb2.


athornton (approved):
LGTM

PR #1821, opened 2022-11-08
Update nublado summit deployment to cycle 27.

tribeiro:
FYI, this cannot be pushed to the summit until Thursday after the deployment is completed. 

⚠️  This PR needs to remain unmerged until the appropriate time. I will do the merge when it is time to sync.


athornton (approved):
LGTM -- but as Tiago says, do not merge until we're ready to take it live.

PR #1822, opened 2022-11-10
DM-36868: Create a cloudsql-proxy subchart

rra:
Move the critical part of sqlproxy-cross-project into a subchart so that it can be shared with Gafaelfawr (which will be done in a subsequent commit).

### DIFF
--- a/charts/README.md
+++ b/charts/README.md
@@ -1,0 +1,12 @@
+Shared charts
+=============
+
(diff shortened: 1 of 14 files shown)


rra:
On second thought, after looking at what this does to the values file for Gafaelfawr and how complicated it makes things, as well as not handling the sidecars, I don't think this is worth doing. I'm going to leave well enough alone and leave this as a stand-alone service independent of other uses of the Cloud SQL Proxy.

PR #1823, opened 2022-11-10
usdfprod phalanx deployment

yee379:
the long awaited PR - note there are two 'features' in my branch that should probably be stripped out, or PR'd separately.

1) the installer script uses a vault approle rather than a token for installs
2) added http proxy envs to the tap service helm chart to facilitate outside communication.

i haven't had time to cherry pick those the above out, nor squash the commit history to make it more legible; so please consider this a draft until we can clean up the PR.


pav511:
Updated PR for this draft has been submitted:
https://github.com/lsst-sqre/phalanx/pull/1915

rra:
Should we delete this PR given the new one at #1915?

yee379:
@rra - yes; PR #1915 supersedes this PR and thus this PR #1823 is deemed unnecessary.

PR #1825, opened 2022-11-11
DM-35698: Switch IDF int to CILogon

rra:
Add the COmanage configuration and change the authorization groups.


rra:
Not to be merged until the day that we're ready to do the cutover and make everyone re-enroll.

PR #1826, opened 2022-11-12
DM-29086: Switch Gafaelfawr to the Bitnami Redis chart

rra:
Use the Bitnami Redis chart instead of a hand-rolled Redis stateful set.  In theory this will allow some higher-availability features, although for now we're just using the master and the replica is unused.


rra:
Superseded by #1867.

PR #1838, opened 2022-11-23
Tickets/dm-36041

bsmartradio:
Fix strimzi-registry-operator bugs and outdated code so that the alert-schema-registry can build properly and the ingestor can read from it.

### DIFF
--- a/services/alert-stream-broker/values.yaml
+++ b/services/alert-stream-broker/values.yaml
@@ -1,6 +1,7 @@
 strimzi-registry-operator:
   # Should match the cluster name used by the alert-stream-broker
   clusterName: alert-broker
-
   # Should match the namespace where the alert-broker cluster runs
   watchNamespace: alert-stream-broker
+
+  operatorNamespace: "alert-stream-broker"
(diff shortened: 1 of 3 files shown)


services/alert-stream-broker/values.yaml
```diff
@@ -1 +1,6 @@
-# This file intentionally blank - no customization needed
+strimzi-registry-operator:
+  # Should match the cluster name used by the alert-stream-broker
+  clusterName: alert-broker
+
+  # Should match the namespace where the alert-broker cluster runs
+  watchNamespace: alert-stream-broker
```
jonathansick, line 6: This is missing some necessary configurations. See the [sasquatch](https://github.com/lsst-sqre/phalanx/blob/3d565962e3c300a643dd163c9a72c906d05169d9/services/sasquatch/values.yaml#L7-L10) for guidance.

```suggestion
strimzi-registry-operator:
  # Should match the cluster name used by the alert-stream-broker
  clusterName: alert-broker

  # Should match the namespace where the alert-broker cluster runs
  watchNamespace: alert-stream-broker
  
  operatorNamespace: "alert-stream-broker"
```

PR #1841, opened 2022-11-30
Add admin:jupyterlab mapping to Gafaelfawr

athornton:


rra:
You'll need to add the scope to `config.knownScopes` as well, probably directly in `values.yaml` since there's no reason not to deploy it everywhere.

rra (approved):

PR #1846, opened 2022-12-05
DM-37226: Switch remaining ingresses to GafaelfawrIngress

rra:
Switch all the remaining authenticated ingresses except for the nublado2 configuration, which will continue to have to be hand-configured to work with Zero-to-JupyterHub, over to using GafaelfawrIngress to generate the ingress.

In the process, hard-code more things in the GafaelfawrIngress templates and remove them from values.yaml, since we're standardizing on not customizing paths, scopes, and similar bits of the ingress.


athornton (approved):
LGTM although my eyes did kind of glaze over with the replacement of one set of boilerplate with a slightly smaller set of boilerplate, so it's not impossible I missed something.  But on the whole it brings a cheerful consistency to our ingresses.

PR #1876, opened 2023-01-06
DM-37390: Convert noteburst and times-square to new Redis chart

rra:
Use the new internal Redis chart that only creates a single instance, since in practice we were only using the master created by the Bitnami chart.  Use SSD storage for Redis on IDF dev.

Although authentication was enabled in values.yaml for noteburst, Redis authentication was not actually in use so far as I could determine, so do not configure authentication and rely on the NetworkPolicy to control access.  (Ideally we should add a secret, but we can tackle this later after we simplify secret management.)

### DIFF
--- a/services/noteburst/Chart.yaml
+++ b/services/noteburst/Chart.yaml
@@ -11,11 +11,10 @@ maintainers:
   - name: jonathansick
     url: https://github.com/jonathansick
 
-# Additional charts that this chart uses
 dependencies:
   - name: redis
-    version: 17.4.0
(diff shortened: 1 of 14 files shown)


jonathansick (approved):
This is great, thanks.

I didn't have auth enabled for either Times Square or Noteburst, but of course we can add that if you think it's a good idea.

What are the client labels on the Deployments intended for? to enable NetworkPolicy later? Edit: — yes, just realized that the redis chart is already set up for that. 👍

PR #1905, opened 2023-01-24
DM-37659: Handle apps with names that have _ in values

jonathansick:
It was a practice in the environment values.yaml files to use `_` in place of `-` in app names. e.g.

```yaml
times_square:
  enabled: true
```

Instead of using the actual app name:

```yaml
times-square:
  enabled: true
```

This change handles that situation so that many missing apps with `-` in their names but that use `_` in the environment values.yaml files now appear in documentation homepages for environments..

### DIFF
--- a/src/phalanx/docs/models.py
+++ b/src/phalanx/docs/models.py
@@ -326,11 +326,15 @@ class Environment:
                 apps.append(app)
                 continue
 
-            try:
+            if app.name in values:
                 if values[app.name]["enabled"] is True:
                     apps.append(app)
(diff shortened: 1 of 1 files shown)


rra (approved):
Yup, looks good. Ideally we'd just change all the underscores to dashes but I think it makes the Helm templating a bit weird, so it may not be completely trivial to do.

jonathansick:
I noticed that `telegraf-ds` keeps its `-` as a `-` in the values.yaml so maybe we were concerned about a non-issue after all?

PR #1908, opened 2023-01-24
Add roundtable-prod and roundtable-dev

jonathansick:
- Adds roundtable-prod and roundtable-dev values to science-platform/
- Adds service values files for the initial services:
  - argocd
  - cert-manager
  - gafaelfawr
  - ingress-nginx
  - squareone (to have a landing page, initially)
  - vault-secrets-operator
- Documentation homepages for roundtable-prod and roundtable-dev

The ingress-nginx snippet exclusion was needed to prevent a bootstrap sequence issue between ingress-nginx and gafaelfawr.

### DIFF
--- a/applications/gafaelfawr/values-roundtable-dev.yaml
+++ b/applications/gafaelfawr/values-roundtable-dev.yaml
@@ -16,7 +16,7 @@ config:
 
   # Support OpenID Connect clients like Chronograf.
   oidcServer:
-    enabled: true
+    enabled: false
 
   groupMapping:
     "admin:provision":
--- a/applications/gafaelfawr/values-roundtable-dev.yaml
+++ b/applications/gafaelfawr/values-roundtable-dev.yaml
@@ -19,10 +19,6 @@ config:
     enabled: false
 
   groupMapping:
-    "admin:provision":
-      - github:
-          organization: "lsst-sqre"
-          team: "square"
     "exec:admin":
       - github:
           organization: "lsst-sqre"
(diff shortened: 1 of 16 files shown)


jonathansick:
I was able to deploy both to roundtable-dev.lsst.cloud (done during co-work) and roundtable.lsst.cloud (done Friday, and went without issue using the work-arounds we applied.

applications/gafaelfawr/values-roundtable-dev.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+  groupMapping:
+    "admin:provision":
+      - github:
+          organization: "lsst-sqre"
+          team: "square"
+    "exec:admin":
+      - github:
+          organization: "lsst-sqre"
+          team: "square"
```
rra, line 29: It's an interesting question what group mappings to set up for the Roundtable use case. I think I'd start with just `exec:admin`. I think `admin:provision` is going to go away; it was originally created to talk to moneypenny, which is being removed with the new lab controller work. (I also kind of want to rename `exec:admin` to `admin:<something>` in the future, but we can do that later.)

applications/ingress-nginx/values-roundtable-dev.yaml
```diff
@@ -0,0 +1,7 @@
+ingress-nginx:
+  controller:
+    service:
+      loadBalancerIP: "34.29.88.130" # public-ip reserved by Terraform
+    config:
+      server-snippet: ""
+      http-snippet: ""
```
rra, line 7: I *think* if you rebase against the current main you can now drop this.

rra (approved):
Looks great, just a few minor comments.

PR #1909, opened 2023-01-25
[DM-37685] Create prompt-processing Kafka user in Sasquatch

afausti:
Add `prompt-processing` Kafka user and the `test.next-visit` topic for testing the prompt processing infrastructure.


dspeck1 (approved):
Looks good

PR #1913, opened 2023-01-27
DM-37221: Convert to 1Password Connect API

rra:
Rather than using third-party onepassword Python API, which is not pip-installable and doesn't work on all platforms, use the 1Password Connect API against the new server that we're installing on Roundtable.


rra:
This doesn't yet work completely, but that seems to be because we're missing secrets from 1Password that are required for an environment. The 1Password interactions do appear to work properly now.

jonathansick (approved):
Very nice!

One more change to make: The bootstrapping docs mention the `1password` dependency.

PR #1918, opened 2023-01-30
tickets/DM-37649

mareuter:


afausti (approved):
Looks good, and thanks for updating the connector configuration as well. 

We need to resume the work on the Telegaf Kafka Consumer, so we'll have only one set of configurations for the Kafka -> InlduxDB connector. We did start testing the Telegraf Kafka Consumer at TTS that's why we have things duplicated at the moment. Note that at the Summit there's only one set that corresponds to the InfluxDB Sink connector configuration.

PR #1922, opened 2023-02-01
DM-37810: Switch to dashes in the environments configuration

rra:
Previously, we used an underscore form of the application name when configuring its enabled status in the values files in the environments directory. Switch to using a name matching the application directory and chart name in all cases, and use the mildly more complex Helm and `yq` syntax required to handle names containing dashes. Revert the docs change to handle underscores as now unneeded.

Also change some other settings to use camelCase for consistency with Helm conventions, and set the version of the science-platform chart for Helm v3.


jonathansick (approved):
This looks great!

PR #1967, opened 2023-02-28
Disable unused apps usdf rebase

pav511:
This directs usdf deployments to lsst repo, and disables a few apps not in use.
Also includes config for mirrormaker2 and rest-proxy in Sasquatch for prompt processing.


yee379:
note this PR our first step towards /not/ using a fork of phalanx for usdf.

environments/values-usdfdev.yaml
```diff
@@ -19,7 +19,7 @@ gafaelfawr:
 hips:
   enabled: false
 ingress-nginx:
-  enabled: true
+  enabled: false
```
rra, line 22: This does have to be enabled. We don't support a configuration without ingress-nginx enabled. We can still merge this as a temporary step towards more convergence, but wanted to flag that this will have to change back in the future.

rra (approved):
In general, looks good, but flagging ingress-nginx for future attention.

environments/values-usdfdev.yaml
yee379: we are running [loft.sh's vcluster](https://github.com/loft-sh/vcluster/tree/main) and have an ingress-nginx installed in the 'parent' cluster. we are using vcluster in the mode where we run phalanx inside a virtualised cluster and (local) Ingress requests get translated up to the 'parent' - as such an ingress-nginx inside the phalanx'd vcluster isn't necessary and will use the global one that we have installed. this also allows us to apply kyverno policies to limit hijacking of ingress `host:` etc. we have been running like this since we first installed phalanx.

yee379: yep, understood. i did wonder why you created a GF Ingress object type and would like to understand further the specific requirements and implementation details. we (un?)fortunately have to implement some degree of security controls on our side too, and this was deemed a good compromise.

rra: Rationale for `GafaelfawrIngress` is here: https://dmtn-193.lsst.io/#configuration

rra: That at least should be unrelated to this, though, since `GafaelfawrIngress` is processed by the Gafaelfawr Kubernetes operator into normal `Ingress` objects. The tricky part is that they contain a reference to a handler location that needs to be injected into every server in the generated NGINX configuration, which is necessary to get auth subrequests to work properly.

PR #1975, opened 2023-03-01
Add nublado v3

athornton:


rra (commented):
Oh, yeah, something definitely went wrong with helm-docs. It's acting like it's not picking up our configuration in `helm-docs.md.gotempl` at the top level of the repository.

athornton:
Gonna YOLO merge it so I can get data-dev running nublado3 from master before I leave for Chile.

athornton:
Gonna try again in case that was transient.

PR #1981, opened 2023-03-02
DM-38208: Prepare Phalanx for using merge queues

rra:
GitHub has a new feature called merge queues that allows one to queue up a number of PRs for merging, and each will be properly rebased and merged when their tests pass. In order to use it, it has to be mandatory for merges and GitHub Actions that may block a merge have to trigger on `merge_group`.

Take this opportunity to rework the documentation build a bit:

- Always run the documentation build but skip the work if the documentation hasn't changed. This allows us to block merges on documentation syntax errors or missing metadata for new applications or environments.
- Re-add the documentation link check, which we'd disabled in January, as a separate CI workflow so that it doesn't block merges. Run it on pushes and once a week.
- Update to Python 3.11 for documentation builds.
- Use the new shared actions for documentation builds and uploads.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,7 @@
 name: CI
 
 "on":
+  merge_group: {}
   pull_request:
     branches:
       - "main"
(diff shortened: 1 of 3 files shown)


jonathansick (approved):
Excellent

PR #1996, opened 2023-03-08
tickets/SITCOM-692

mareuter:
Summit: Add gis influxdb sink to sasquatch.


afausti (approved):
Looks good to me!

Now that we have replication to USDF (tomorrow probably) we have to add this configuration to values-usdfprod.yaml as well.

PR #2001, opened 2023-03-14
Fix prompt-processing user permissions

afausti:
- Give read access for the prompt-processing user on the lsst.sal.ScriptQueue.logevent_nextVisit topic


dspeck1 (approved):
looks good

PR #2016, opened 2023-03-21
DM-38320: Update plot-navigator for USDF deployment

ctslater:


athornton (approved):
Presuming you've tested it by deploying from the branch at USDF, it LGTM, but I haven't actually tested that the yaml indentation for volumes is correct and stuff.

If you haven't deployed from the branch, lmk.  I can help with that if you don't have access to do it yourself.

PR #2029, opened 2023-03-27
Testing branch for Nublado improvements

rra:
- Move all controller configuration into a config subkey and move the image specification and other deployment configuration for the controller under the controller key.
- Rename the object templates to more clearly distinguish between configuration for the controller and configuration for JupyterHub.
- Add Helm docs comments for all of the configuration and improve the comments that were there. Set a @default string for long settings to keep from dumping large values into the documentation page.
- Stop templating the names of resources. If we want to test the Helm chart, we'll install it in a different namespace.
- Use controller.config.safir.pathPrefix to determine the path of the ingresses of the controller.
- Separate the ConfigMap for the controller from the ConfigMap for JupyterHub so that we don't mount the ConfigMap of the other service and possibly confuse matters.
- Delete some more Zero to JupyterHub settings that should be irrelevant given that we're using our own spawner.
- Remove a default for lab volumes that only works on one IDF instance.
- Prepull the recommended image on data-dev explicitly by tag so that the menu will be properly populated.


rra:
Everything in this branch has now been merged.

PR #2037, opened 2023-03-27
Add placeholder plot-navigator usdfdev values file.

ctslater:


ctslater:
Sorry, I think I broke the main docs build, @rra is this the right fix? I haven't tried to make the contents of the values file correct, that's next on the to-do list.

ctslater:
Thanks!

PR #2058, opened 2023-04-04
DM-38466: Bind nublado-controller to a Google service account

rra:
This is required to be able to access the Google Artifact Registry.


athornton (approved):
LGTM

PR #2060, opened 2023-04-04
DM-38554: Move auto-repo and Butler out of Nublado values.yaml

rra:
Don't pull the tutorial notebooks on every instance by default since they only work at the IDF (and maybe the USDF). Instead, set that in values-idfdev.yaml.

Pull the Butler configuration environment variables that are specific to the IDF out of values.yaml and into values-idfdev.yaml.


athornton (approved):
LGTM

PR #2063, opened 2023-04-06
Prompt processing user test next visit topic permissions

dspeck1:
Add permissions for test next visit topic for prompt processing user


afausti (approved):
Looks good!

PR #2074, opened 2023-04-11
DM-38654: Add new write:sasquatch scope

rra:
Add a new distinguished scope for write access to the telemetry service, since we will eventually want to create service tokens that have access only to that service. For now, add that scope only in the environments that deploy Sasquatch and only for administrators.


afausti (approved):
Looks great thanks!

PR #2079, opened 2023-04-12
DM-38654: Document creating Sasquatch write tokens

rra:
Add documentation for how to create a service token for writing to Sasquatch.


afausti (approved):
Thank you so much for documenting this in a timely manner. The steps to generate a service token look straightforward.

PR #2086, opened 2023-04-14
DM-38554: Add Slack alerting support to Nublado

rra:
Add the Slack webhook to the secret for the new Nublado and inject it into the lab controller pod if Slack alerting is enabled. Enable Slack alerting on IDF dev.


athornton (approved):
LGTM

PR #2099, opened 2023-04-18
DM-38747: Add Kerberos GSSAPI bind support for Gafaelfawr

rra:
Add a new configuration option to set the Kerberos configuration for GSSAPI bind support. If set, create a separate secret for the keytab and mount it in all the containers, and run Gafaelfawr under k5start. Also mount the krb5.conf file and create a /tmp directory, since k5start needs a writable directory to use for ticket caches.


athornton (approved):
LGTM

PR #2104, opened 2023-04-20
DM-38795: Add Butler secret to data-int

rra:
The new Nublado requires explicitly specifying the secrets to mount in the container and uses a different path by default. Adjust the IDF int configuration accordingly.


athornton (approved):
I think this will work.

PR #2105, opened 2023-04-20
DM-38795: Add separate lab secret

rra:
The secrets that we inject into the lab shouldn't use nublado-secret since it has a complex templating setup. Instead, create a separate lab-secret VaultSecret to collect secrets we're injecting into labs, and reference that in the IDF int configuration.


athornton (approved):
LGTM

athornton (approved):
LGTM

PR #2115, opened 2023-04-26
U/pav/mobu configuration

pav511:
dev only mobu config until dev slack alerts are resolved


rra (commented):
I don't know if this was meant to be ready for review, but the change to Argo CD looks accidental, and there is a full usdfprod configuration that doesn't seem to match the PR description.

PR #2116, opened 2023-04-27
tickets/DM-38729: Add fileserver

athornton:
This adds user fileservers to an RSP, managed by jupyterlab-controller.

### DIFF
--- a/applications/nublado/templates/controller-ingress-files.yaml
+++ b/applications/nublado/templates/controller-ingress-files.yaml
@@ -13,7 +13,7 @@ config:
     notebook: {}
 template:
   metadata:
-    name: "controller-user"
+    name: "controller-files"
     annotations:
       nginx.ingress.kubernetes.io/use-regex: "true"
   spec:
--- a/applications/nublado/templates/controller-ingress-files.yaml
+++ b/applications/nublado/templates/controller-ingress-files.yaml
@@ -1,0 +1,30 @@
+apiVersion: gafaelfawr.lsst.io/v1alpha1
+kind: GafaelfawrIngress
+metadata:
+  name: "nublado-controller-files"
+  labels:
+    {{- include "nublado.labels" . | nindent 4 }}
+config:
+  baseUrl: {{ .Values.global.baseUrl | quote }}
+  scopes:
+    all:
+      - "exec:notebook"
+  delegate:
+    notebook: {}
+template:
+  metadata:
+    name: "controller-user"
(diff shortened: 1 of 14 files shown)


applications/nublado/templates/controller-ingress-files.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: gafaelfawr.lsst.io/v1alpha1
```
rra, line 1: If this doesn't need to be usable as a separate WebDAV route (with basic auth and whatnot), you can merge this with `controller-ingress-user.yaml` as an additional route and save the overhead of having another ingress.

applications/nublado/templates/controller-serviceaccount.yaml
```diff
@@ -61,6 +77,30 @@ rules:
...
+      - "get"
+      - "list"
+      - "watch"
+  - apiGroups: ["apps"]
+    resources:
+      - "deployments/status"
+    verbs:
+      - "get"
+      - "list"
+      - "watch"
```
rra, line 94: We didn't end up using deployments, so these can be dropped.

environments/templates/nublado-application.yaml
```diff
@@ -7,6 +7,14 @@ spec:
...
     - "kubernetes"
 ---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: "fileservers"
+spec:
+  finalizers:
+    - "kubernetes"
+---
```
rra, line 17: We want the created fileservers to be visible in Argo CD, so this should use the same approach as we use for the nublado-users application (but also create a namespace). So there needs to be another Argo CD application that we enable if the file server is enabled.

I've been thinking that it may be better to create those Argo CD applications directly in the nublado Helm chart rather than creating them in environments, since they're really implementation details of nublado and Argo CD supports that app of app construction. So ideally you'd also move the creation of the nublado-users app into nublado along with this.

rra:
Oh, also, create a generic `NetworkPolicy` in the fileservers namespace that will match all running fileserver pods and do the normal work of ensuring that only the ingress can talk to them. You could also block all egress except to the ingress, I think? Although I'm not sure how `NetworkPolicy` affects mounted NFS volumes.

athornton:
Once I've updated work on DM-38972 with this, we should close this one unmerged.

rra:
I think this branch is now obsolete.

PR #2122, opened 2023-05-01
Redirect dev TAP to USDF Qserv

fritzm:


rra:
This PR is quite old and would no longer apply. I added a comment in #2720 to record the IP address, and am closing this since we'd want to create it afresh if we needed it again.

PR #2124, opened 2023-05-02
Add a new connector for lsst.rubintv

afausti:
This PR adds a new connector for the lsst.rubintv namespace in Sasquatch at USDF dev.


hsinfang:
Thank you very much @afausti ! 

We'll also use Sasquatch at the summit, so does that mean that the same thing needs to be added to `values-summit.yaml` too? 

Can the tags be changed later? These look like good tags to me at the moment, but I am just wondering whether we can change our mind later or tweak them.  @mfisherlevine  might want to weigh in.

PR #2133, opened 2023-05-04
DM-39014: Configure Kafka/Sasquatch for Roundtable

jonathansick:
This PR configures a Kafka cluster for Roundtable. We're doing this by configuring Sasquatch so that down the road we can easily add Sasquatch's monitoring capabilities to Roundtable given that Sasquatch itself is also Kafka based.

- [x] Adds toggle configurations to Sasquatch to turn off individual components. This lets us initially deploy only Kafka infrastructure
- [x] Adds toggles to enable specific Kafka listeners. Initially Roundtable will only use the internal TLS-based listeners.
- [x] Adds toggles to enable specific default users.
- [x] Adds configurations to enable Kafka brokers and Zookeeper servers to tolerate taints on node pools.
- ~~[ ] Configures Kafdrop to use TLS authentication~~ May need to be done later.

The Sasquatch deployment on roundtable-dev is here: https://roundtable-dev.lsst.cloud/argo-cd/applications/argocd/sasquatch

### DIFF
--- a/applications/sasquatch/Chart.yaml
+++ b/applications/sasquatch/Chart.yaml
@@ -26,6 +26,7 @@ dependencies:
     version: 1.2.5
     repository: https://helm.influxdata.com/
   - name: kapacitor
+    condition: kapacitor.enabled
     version: 1.4.6
     repository: https://helm.influxdata.com/
   - name: kafdrop
--- a/applications/sasquatch/Chart.yaml
+++ b/applications/sasquatch/Chart.yaml
@@ -22,6 +22,7 @@ dependencies:
     condition: kafka-connect-manager.enabled
     version: 1.0.0
   - name: chronograf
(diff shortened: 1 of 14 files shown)


jonathansick:
@afausti Let me know what you think of the extra configurations I'm suggesting here. Overall the goal here is to be able to deploy a fully-pared down Kafka cluster, and then but re-enabling different toggles an environment can start adding full Sasquatch capabilities.

afausti (approved):
This looks great, thanks for making Sasquatch more generic.
I'm happy you managed to reuse the app in Roundtable. 
I can move some of the users to their respective subcharts as I did for the replicator user, that would make the configuration a bit cleaner.
I didn't try TLS authentication with Kafdrop, but it seems supported. 
I used SASL/SCRAM because it was easier, but it would be nice to have that.

PR #2143, opened 2023-05-08
Enable Alert Stream Broker in USDF Dev

dspeck1:
Enable Alert Stream Broker in USDF Dev.   Some settings commented out for now until values determined.


bsmartradio (approved):
Everything looks good. Nothing jumps out to me that might break things, though we will see when we try it out.

PR #2164, opened 2023-05-15
Use JSON logging format for cert-manager

rra:
cert-manager now supports configuring structured logging. Do that, and improve the chart values.yaml documentation. Remove the name overrides, since we don't use them, and simplify the resources we create directly.


rra:
This doesn't work yet; we're waiting for cert-manager 1.12.0

PR #2169, opened 2023-05-17
tickets/DM-38972: add user fileserver support to nublado

athornton:


applications/nublado/templates/controller-ingress-files.yaml
```diff
@@ -0,0 +1,28 @@
...
+config:
+  baseUrl: {{ .Values.global.baseUrl | quote }}
+  scopes:
+    all:
+      - "exec:notebook"
+  delegate:
+    notebook: {}
+template:
+  metadata:
+    name: "controller-files"
```
rra, line 16: Name should match the name of the `GafaelfawrIngress`.

applications/nublado/templates/controller-serviceaccount.yaml
```diff
@@ -53,6 +53,22 @@ rules:
...
+  - apiGroups: ["batch"]
+    resources: ["jobs"]
+    verbs:
+      - "create"
+      - "delete"
+      - "get"
+      - "list"
+      - "watch"
+  - apiGroups: ["networking.k8s.io"]
+    resources: ["networkpolicies", "ingresses"]
```
rra, line 65: `networkpolicies` is already handled below and shouldn't be listed here.

applications/nublado/templates/controller-serviceaccount.yaml
```diff
@@ -61,6 +77,30 @@ rules:
...
+      - "get"
+      - "list"
+      - "watch"
+  - apiGroups: ["apps"]
+    resources:
+      - "deployments/status"
+    verbs:
+      - "get"
+      - "list"
+      - "watch"
```
rra, line 94: Delete; we're not using deployments.

applications/nublado/values-idfdev.yaml
```diff
@@ -2,9 +2,13 @@ controller:
   googleServiceAccount: "nublado-controller@science-platform-dev-7696.iam.gserviceaccount.com"
   slackAlerts: true
   image:
-    tag: "tickets-DM-38795-queue"
-    pullPolicy: "Always"
+    tag: "0.4.0"
```
rra, line 5: Don't hard-code the tag; it should come from `Chart.yaml` (which needs to be updated with the new version; I don't think you've done that).

applications/nublado/values.yaml
```diff
@@ -41,8 +41,34 @@ controller:
   # set in the corresponding Nublado Vault secret.
   slackAlerts: false
 
+
```
rra, line 44: Stray blank line.

applications/nublado/values.yaml
```diff
@@ -41,8 +41,34 @@ controller:
...
   # Passed as YAML to the lab controller.
   config:
+  # The controller also manages file serving environments
```
rra, line 47: Comment not correctly indented (and IMO I don't think this comment adds anything, so I'd just drop it).

applications/nublado/values.yaml
```diff
@@ -41,8 +41,34 @@ controller:
...
+      # -- Pull policy for fileserver container
+      # @default -- `IfNotPresent`
+      pullPolicy: IfNotPresent
+
+      # -- Timeout for user fileservers, in seconds
+      # @default -- 3600
+      timeout: 64800
+
+      # -- Namespace for user fileservers
+      # @default -- `fileservers`
```
rra, line 70: `@default` is not something you should routinely set. You should only set it when the default isn't what's present in this file. It's the default from the perspective of someone writing a `values-<environment>.yaml` file. All of these `@default` lines should therefore be deleted (in some cases they're not correct since they don't match the actual default in this file).

applications/nublado/values.yaml
```diff
@@ -128,10 +154,16 @@ controller:
...
-      volumes: {}
+      # -- Volumes that should be mounted in lab pods. This supports NFS,
+      # HostPath, and PVC volume types (differentiated in source.type)
+      # volumes:
+      # - containerPath: "/project"
+      #   mode: "rw"
+      #   source:
+      #     type: nfs
+      #     serverPath: "/share1/project"
+      #     server: "10.87.86.26"
```
rra, line 165: If you really want to include an example, it has to go after the setting, not before, or it gets sucked into the Markdown documentation and then badly misformated and makes it hard to read.

applications/nublado/templates/controller-ingress-files.yaml
```diff
@@ -0,0 +1,28 @@
...
+  name: "nublado-controller-files"
+  labels:
+    {{- include "nublado.labels" . | nindent 4 }}
+config:
+  baseUrl: {{ .Values.global.baseUrl | quote }}
+  scopes:
+    all:
+      - "exec:notebook"
+  delegate:
+    notebook: {}
```
rra, line 13: This is much too much power for the fileserver. You only need an internal token with no scopes.

rra (commented):
I think this is still missing the creation of the Argo CD application that everything will be owned by.

applications/nublado/templates/controller-serviceaccount.yaml
athornton: Rather, I should delete the below stanza, since networkpolicies and ingresses want the same verbs.

applications/nublado/templates/controller-ingress-files.yaml
athornton: ...don't I need a regular user token with scope "exec:notebook" ?  Just leave off the delegate: stanza, right?

athornton:
Is there a separate argocd application?  It's all rolled up under nublado, isn't it?

athornton:
ah.  cf https://github.com/lsst-sqre/phalanx/pull/2116 and address those too.

applications/nublado/templates/controller-ingress-files.yaml
rra: No, you don't need any scopes at all; the only thing you're doing with the token is asking for the user's UID, GID, and group membership, which doesn't require any scopes. You therefore want:
```
delegate:
  internal:
    service: "nublado"
    scopes: []
```
See https://gafaelfawr.lsst.io/user-guide/gafaelfawringress.html#internal-tokens

applications/nublado/templates/controller-ingress-files.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: GafaelfawrIngress
+metadata:
+  name: "nublado-controller-files"
+  labels:
+    {{- include "nublado.labels" . | nindent 4 }}
+config:
+  baseUrl: {{ .Values.global.baseUrl | quote }}
+  scopes:
+    all:
+      - "exec:notebook"
```
rra, line 11: Now you don't have a delegated token at all. Don't you need the user's UID, GID, and group information to spawn their fileserver?

rra (commented):
`values-idfint.yaml` also needs an update for the new volume syntax.

applications/nublado/templates/controller-ingress-files.yaml
```diff
@@ -0,0 +1,27 @@
...
+metadata:
+  name: "nublado-controller-files"
+  labels:
+    {{- include "nublado.labels" . | nindent 4 }}
+config:
+  baseUrl: {{ .Values.global.baseUrl | quote }}
+  delegate:
+    internal:
+      service: "nublado"
+      scopes: []
```
rra, line 12: Sorry, apparently that wasn't clear. I didn't mean to delete the scopes restriction. This would be syntactically invalid and will be rejected by the CRD. You have to specify both (a) the scope required to access this service and (b) the type of token to delegate to it.

In other words:
```suggestion
  scopes:
    all:
      - "exec:notebook"
  delegate:
    internal:
      service: "nublado"
      scopes: []
```

rra (approved):
Ingress still isn't quite right, but everything else looks good.

rra (approved):
Ingress still isn't quite right, but everything else looks good.

PR #2172, opened 2023-05-18
DM-39232: Improve nublado documentation and roles

rra:
Add helm-docs comments to all of the overrides for Zero to JupyterHub configuration so that we get more useful chart documentation.

Delete the creation of a new self role. So far as I can tell, this was never used or assigned to anything, and therefore wasn't doing anything. I think this was an earlier attempt to solve a problem that was solved by setting authenticatePrometheus to false.

Change the default permissions of the server role (for spawned labs) to self rather than inherit. This only affects labs spawned by JupyterHub administrators and ensures that the lab token only inherits their regular user permissions and not their admin permissions.

### DIFF
--- a/applications/nublado/values.yaml
+++ b/applications/nublado/values.yaml
@@ -269,32 +269,57 @@ hub:
 # Configuration for the Zero to JupyterHub subchart.
 jupyterhub:
   hub:
+    # -- Whether to require metrics requests to be authenticated
     authenticatePrometheus: false
+
     image:
+      # -- Image to use for JupyterHub
       name: ghcr.io/lsst-sqre/rsp-restspawner
+
+      # -- Tag of image to use for JupyterHub
       tag: 0.2.0
+
+    # -- Resource limits and requests
     resources:
       limits:
         cpu: 900m
         memory: 1Gi  # Should support about 200 users
+
     db:
-      # Password comes from the nublado-secret.
+      # -- Type of database to use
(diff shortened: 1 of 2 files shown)


applications/nublado/values.yaml
```diff
@@ -244,100 +244,165 @@ hub:
...
       enabled: false
+
     loadRoles:
-      self:
-        scopes: ['admin:servers!user', 'read:metrics']
       server:
-        scopes: ['inherit']  # Let server use API like user
+        # -- Default scopes for the user's lab, overridden to allow the lab to
+        # delete itself (which we use for our added menu items)
+        scopes: ["self"]
```
athornton, line 323: I think if you do this it is going to break the items we added for save?-and-exit-and-logout?

athornton (commented):
All the doc stuff is great.  Definitely do that.

But I don't think that the server role is going to do what you think it is.  I'm pretty sure that "self" will break the menu items we add that require talking back to the hub.  If you've tested this and, say, "Save, Exit, and Logout" still works, then you're right and you should merge this, but I'm pretty sure I added inherit so that the lab user can use the Hub API, which was a change, IIRC, between JH 2 and JH 3.

It should be possible to preserve API access with `access:services` and some horizontal filter, but I am suspicious that this change, as is, will break our menu items, leaving us with no way for users to kill their own labs except by going to the hub menu.

PR #2177, opened 2023-05-19
DM-39232: Add Nublado v3 configuration for USDF dev

rra:
Add a new configuration for USDF dev modelled after its nublado2 configuration. USDF overrode the ingress annotations, so add support for setting ingress annotations for the proxy so that it can continue to do this. That uncovered missed annotations that we added to nublado2 and not nublado, which have now been copied over.

Rename the ingress to reflect that it's an ingress for the proxy, not the hub (the proxy routes to either the hub or the lab).

### DIFF
--- a/applications/nublado/Chart.yaml
+++ b/applications/nublado/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nublado
-version: 3.0.0
+version: 1.0.0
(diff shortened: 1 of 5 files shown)


rra:
This won't work yet because currently the lab controller doesn't support `subPath`. I'm working on that next.

athornton (approved):
LGTM

PR #2181, opened 2023-05-22
Updates to roe environment (S3 buckets, default hips url, fixed Qserv)

stvoutsin:
Changes include:

- Modify portal/values-roe.yaml to scale up replicas & set a default hipsUrl
- Modify tap/values-roe.yaml to change QServ IP & setup S3 buckets for async

### DIFF
--- a/applications/portal/values-roe.yaml
+++ b/applications/portal/values-roe.yaml
@@ -1,3 +1,8 @@
+replicaCount: 2
+
 resources:
   limits:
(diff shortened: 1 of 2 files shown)


rra (approved):
Looks good, go ahead and click "Merge when ready" if you don't have any more changes you need to make.

rra:
(If you don't have permissions, let me know.)

PR #2182, opened 2023-05-22
Run nubladov2 and v3 in parallel: IDF dev/int, USDF dev

athornton:


rra (approved):
noteburst needs to be pointed to `/n3` instead of `/nb` in dev as well.

athornton:
I'm going to merge this, as noteburst is not yet configurable.  When that change hits noteburst, then we will need another PR to Phalanx to set the noteburst-hub-url.

PR #2183, opened 2023-05-22
Move nublado2 to /n2 at usdf-dev

athornton:


rra (approved):

athornton:
Gonna close without merging, since Jonathan added noteburst support to point at whatever, and keeping n2 on /nb and n3 on /n3 is less confusing.

PR #2187, opened 2023-05-22
DM-39325: Run system-test notebooks on data-dev

rra:
Upgrade the mobu testing on data-dev from a simple Python loop to the system-test notebooks, now that Butler access has been configured. Eventually we'll want to upgrade all the way to running the tutorial notebooks, but currently they still have an assertion preventing this.


rra:
This unfortunately doesn't currently work because the test Butler repository in the data-dev environment is out of date and needs to be refreshed from int or prod.

PR #2198, opened 2023-05-25
U/pav/broadcast config

pav511:
Configuration for Semaphore message banners at USDF

### DIFF
--- a/applications/semaphore/values-usdfdev.yaml
+++ b/applications/semaphore/values-usdfdev.yaml
@@ -2,10 +2,7 @@ semaphore:
   ingress:
     enabled: true
     annotations:
-      kubernetes.io/ingress.class: nginx  
-pull-secret:
-  enabled: true
-  path: secret/rubin/usdf-rsp-dev/semaphore
+      kubernetes.io/ingress.class: nginx
 config:
   phalanx_env: "usdfdev"
   github_app_id: "337324"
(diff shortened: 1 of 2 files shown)


applications/semaphore/values-usdfdev.yaml
```diff
@@ -1,9 +1,12 @@
...
-    github_app_id: "131502"
-    enable_github_app: "True"
   ingress:
     enabled: true
     annotations:
-      kubernetes.io/ingress.class: nginx
+      kubernetes.io/ingress.class: nginx  
+pull-secret:
+  enabled: true
+  path: secret/rubin/usdf-rsp-dev/semaphore
```
rra, line 8: This is for a Kubernetes pull-secret. You don't want to set this (we should probably delete it from the configuration entirely). I think Semaphore always reads its normal secret from Vault.

applications/semaphore/values-usdfprod.yaml
```diff
@@ -1,2 +1,12 @@
+semaphore:
+  ingress:
+    enabled: true
+    annotations:
+      kubernetes.io/ingress.class: nginx  
+pull-secret:
+  enabled: true
+  path: secret/rubin/usdf-rsp/semaphore
```
rra, line 8: Likewise here, you don't want to set this.

rra (commented):
Looks like there's trailing whitespace somewhere that's causing lints to fail.

rra (approved):

PR #2199, opened 2023-05-26
DM-39435: Add values-level support for adding SSO TAP to Portal menu

gpdf:
Allows turning on the SSO TAP service in the Portal on a per-instance basis, initially for data-int.


rra (approved):

gpdf:
Merging based on @rra review and subsequent discussion on dm-rsp-team.

PR #2231, opened 2023-06-02
Remove /fs/nfs from USDF environments

athornton:


applications/nublado2/values-usdfprod.yaml
```diff
@@ -94,8 +94,6 @@ config:
 #   - name: fs-ddn-sdf-group-rubin
 #     mountPath: /instrument
 #     subPath: lsstdata/offline/instrument
-    - name: fs-nfs
```
ktlim, line 97: Need to remove lines 64 and 65 as well?

athornton: Thanks!

athornton: Done, and removed some commented-out mountpoints

ktlim (approved):

PR #2238, opened 2023-06-07
DM-39435: Update to match the new `suit` default TAP service label

gpdf:
Required for the final public-facing configuration we want:
* LSST DP0.2 DC2
* LSST DP0.3 SSO


rra (approved):

rra:
It would also be helpful to change the `image.tag` setting in `values-idfint.yaml` to the new image being tested (`suit-2023.1.4` I believe) so that everything is consistent when synced from Git.

PR #2240, opened 2023-06-08
tickets/DM-39454: rejuvenate Kafka/telegraf refresh scans

athornton:


applications/sasquatch/charts/telegraf-kafka-consumer/templates/configmap.yaml
```diff
@@ -44,7 +44,7 @@ data:
       sasl_mechanism = "SCRAM-SHA-512"
       sasl_password = "$TELEGRAF_PASSWORD"
       sasl_username = "telegraf"
-      topic_refresh_interval = "60s"
+      topic_refresh_interval = {{ default "60s" $value.topicRefreshInterval |  quote }}
```
afausti, line 47: Nice, having that parametrized is good.

applications/sasquatch/values-idfdev.yaml
```diff
@@ -60,6 +64,8 @@ chronograf:
   ingress:
     enabled: true
     hostname: data-dev.lsst.cloud
+  image:
+    tag: "1.10.1"
```
afausti, line 68: We should be using 1.9.4 everywhere, not critical for dev, but we don't need this change.

afausti (approved):
There's a small unrelated change otherwise looks good.

applications/sasquatch/values-idfdev.yaml
athornton: OK.  We had had 1.10.1 running in dev, and I didn't want to change it.

PR #2256, opened 2023-06-15
Narrow IDF dev metrics and decrease refresh interval for testing

athornton:


athornton:
Don't need to merge it.  Tested fine on branch.

PR #2283, opened 2023-07-06
DM-39839: Phalanx subchart helm docs

Fireye04:


afausti (approved):
Look great! I see now that all the subchart configuration parameters are also documented at the top level. It also updates the alert-stream-broker README which is fine.

PR #2297, opened 2023-07-10
Point nublado2 mobu in int at /n2

athornton:


rra:
Ideally only settings that are not the defaults should be in the configuration file, so all the `url_prefix: /nb` stuff should also be removed.

You also need to fix IDF dev, which has the same problem.

rra (approved):
Thanks, looks good.

PR #2308, opened 2023-07-13
Enable nublado3 at /n3 in IDF prod

athornton:


athornton:
Link check failing because the nublado/values-idfprod file is new and does not exist on main.

rra (approved):
Looks good except for turning off the file server as discussed.

PR #2320, opened 2023-07-17
Increase idf-int quota

athornton:


applications/gafaelfawr/values-idfint.yaml
```diff
@@ -36,7 +36,7 @@ config:
     default:
       notebook:
         cpu: 9.0
-        memory: 27
+        memory: 36
```
rra, line 39: Shouldn't these match the size of a huge? I'm not sure why they're slightly larger.

athornton: Or maybe since we aren't currently doing dask anyway, just drop the quotas?

athornton: I went with "drop".  Until there's some mechanism for users to spawn multiple pods in a namespace, quotas don't _do_ anything, right?

rra: I'd like to continue to test Gafaelfawr quota support. In the long run, this will be the configuration that we're going to use, so we should know if it breaks anything, and I don't want to regress on our progress towards our production configuration.

rra (approved):

PR #2323, opened 2023-07-18
[DM-40075] Enable backup of InfluxDB to another persistent volume

afausti:
An initial attempt to dump an InfluxDB backup to another PV so that Velero could copy the data to S3.


athornton (approved):
LGTM

rra:
This will be different with InfluxDB Enterprise, so this PR is now out of date.

PR #2324, opened 2023-07-18
tickets/DM-39730: Add giftless git-lfs server to roundtable-dev

athornton:
Create an app for giftless and enable it on roundtable-dev.

Also implies adding (incomplete, but good enough for giftless) multi-FQDN support to Gafaelfawr.


applications/giftless/templates/cert.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: cert-manager.io/v1
```
rra, line 1: You shouldn't need to create a separate cert resource. Just pick one of the two ingresses to own the public certificate and add the `cert-manager.io/cluster-issuer: "letsencrypt-dns"` annotation to it. See the squareone ingress.

applications/gafaelfawr/templates/ingress.yaml
```diff
@@ -47,3 +47,48 @@ spec:
...
+                  number: 8080
+          - path: "/.well-known/openid-configuration"
+            pathType: Exact
+            backend:
+              service:
+                name: {{ template "gafaelfawr.fullname" $context }}
+                port:
+                  number: 8080
+          {{- end }}
+    {{- end }}
```
rra, line 93: None of this stuff is going to work for right now, so let's only add `/auth` for additional hosts. (Once we have proper multiple domain support, we won't be exposing those URLs in every hostname anyway.)

applications/gafaelfawr/values.yaml
```diff
@@ -275,6 +275,12 @@ config:
   # [DMTN-235](https://dmtn-235.lsst.io/) for more details on scopes.
   groupMapping: {}
 
+  # -- Defines additional FQDNs for Gafaelfawr.  This doesn't work for
+  # cookie or browser authentication, but for token-based services like
+  # git-lfs or the webdav server it does.
+
+  additionalFQDNs: []
```
rra, line 282: The Kubernetes `Ingress` schema calls this parameter `host` for better or for worse, so let's stay consistent with Kubernetes terminology and call this `additionalHosts`. It should go under a top-level `ingress` key, not under `config`, because it customizes the ingress, not the Gafaelfawr configuration.

applications/giftless/templates/deployment.yaml
```diff
@@ -0,0 +1,191 @@
+apiVersion: apps/v1
```
rra, line 1: Two almost-identical copies of a long resource like this make it hard to understand what the differences are between them. I think it's just which config map is mounted? It looks like fixing this in Helm is tricky because you have to understand variable scopes extremely well, so I'm good with leaving that for a later day, but could you at least add a Helm comment saying what's different between the two copies?

applications/giftless/values.yaml
```diff
@@ -0,0 +1,75 @@
...
+  # -- Tag of giftless image to use
+  # @default -- The appVersion of the chart
+  tag: ""
+
+ingress:
+  # -- FQDNs of giftless ingresses
+  # @default -- Must be overridden in environment-specific values file
+  hostname:
+    readonly: ""
+    readwrite: ""
```
rra, line 40: Every setting should be documented individually.

applications/giftless/values.yaml
```diff
@@ -0,0 +1,75 @@
...
+  # -- Additional annotations to add to the ingress
+  annotations: {}
+
+server:
+  # -- Turn on debugging mode
+  debug: false
+  # -- Port for giftless server to listen on
+  port: 5000
```
rra, line 48: Not sure why this is configurable.

applications/giftless/values.yaml
```diff
@@ -0,0 +1,75 @@
...
+  # -- Number of processes for server
+  processes: 2
+  # -- Number of threads per process
+  threads: 2
+
+# -- Configuration for giftless server
+# @default -- YAML will be used as-is.  cf https://giftless.datopian.com/en/latest/configuration.html
```
rra, line 55: This is a philosophical thing and I understand the arguments for doing this each way, but I've never been a fan of this way of writing Helm charts for things that we use internally and thus can modify whenever we want. Since this Helm chart is just for us and doesn't need to be general, I would rather make the Helm chart highly opinionated by putting the configuration directly into the config map template and only exposing in `values.yaml` the things that we want to change per environment.

applications/giftless/values-roundtable-dev.yaml
```diff
@@ -0,0 +1,25 @@
...
+config:
+  readonly:
+    TRANSFER_ADAPTERS:
+      basic:
+        options:
+          storage_class: "giftless.storage.google_cloud:GoogleCloudStorage"
+          storage_options:
+            project_name: "plasma-geode-127520"
+            bucket_name: "rubin-gitlfs-experimental"
+            account_key_file: "/etc/secret/giftless-gcp-key.json"
```
rra, line 16: It's really annoying that this package is using the Google cloud library but doesn't support workload identity because upstream forces you to provide a key. If they would just let you not provide any credentials, workload identity would just work out of the box. It would probably be worth a PR to them at some point to fix that because we really do not want to ever use externalized credentials like this.

rra:
You'll also need to add the stub files for the generated per-application documentation (I think this is why the docs build is failing).

applications/gafaelfawr/values.yaml
athornton: Done.

applications/giftless/values.yaml
athornton: Yeah, me neither.  Made constant.

applications/giftless/values-roundtable-dev.yaml
athornton: When we go from roundtable-dev to a production environment, I want to have different read and write key files as well.  I agree that a PR to fix this to use workload identity would be a good idea.

applications/giftless/templates/deployment.yaml
athornton: Done.

applications/giftless/values.yaml
athornton: I'm not sure that works.  Giftless expects to literally read a YAML file on startup, but we DO need to interpolate environment-specific variables, so it feels like interpolating via values-*.yaml and then running it through toYaml is going to be much easier.

Am I missing something conceptually here?

rra: See what Gafaelfawr does:

https://github.com/lsst-sqre/phalanx/blob/main/applications/gafaelfawr/templates/configmap.yaml

athornton: Per standup, doing it your way.

rra (approved):
Looks good, thanks!

PR #2326, opened 2023-07-19
DM-40095: Prepare for adding a phalanx Python CLI

rra:
The new Phalanx CLI will be written in Python, so first set up Phalanx's Python environment to be closer to our most recent projects. Changes include:

- Convert README to Markdown and remove some now-extraneous details
- Add our current mypy configuration and fix problems it uncovered
- Add our experimental Ruff configuration and fix problems it uncovered, except in the installer directory since that code will be replaced (but let it make the obvious fixes there)
- Drop isort and flake8 in favor of Ruff
- Pin dependencies and add Makefile targets to update them
- Move the remaining clean target from `docs/Makefile` to the top-level clean target

Dependency pinning requires some further justification, since normally a command-line tool would declare dependencies as normal in `pyproject.toml`. However, since the `phalanx` command-line tool will be the supported way to perform any Phalanx operations, we want to be sure it works with a known set of dependencies so that we're not chasing dependency differences on local systems, and only update those dependencies when we are ready, similar to Python services.

There is no documentation yet because there is no tool yet. That will be added once the tool is ready and capable of doing something.


jonathansick (approved):
I get the desire to pin the dependencies here to make the app reliable to run. This heavily relies on the user setting up a virtual environment to run the phalanx CLI in. I wonder if we can use [pipx](https://pypa.github.io/pipx/) here to install the phalanx CLI for users? pipx takes care of installing the CLI in a dedicated virtual environment. `make setup` could handle the installation for the user.

The one piece of the puzzle is making the `pipx install` use these `requirements.txt` files. I wonder if this would work:

```
pipx install -r requirements/main.txt -e .
```

I don't think this works (there's some discussion at https://github.com/pypa/pipx/issues/359)

Another possibility is using poetry because I think it has built-in lock files, but that'd be a disruptive addition to our tooling.

jonathansick:
This is interesting: you can reference a requirements.txt-format file from `pyproject.toml`. E.g.

```
[build-system]
requires = ["setuptools", "setuptools-scm"]
build-backend = "setuptools.build_meta"

[project]
requires-python = ">=3.11"
name = "phalanx"
dynamic = ["dependencies", "optional-dependencies"]

[tool.setuptools.dynamic]
dependencies = { file = ["requirements/main.txt"] }
optional-dependencies.dev = { file = ["requirements/dev.txt"] }
```

I saw this on [the pip-tools docs](https://pip-tools.readthedocs.io/en/latest/#requirements-from-pyproject-toml)

rra:
Alas, not supported by pip with pinned dependencies.
```
packaging backend failed (code=1), with InvalidRequirement: Expected end or semicolon (after version specifier)
    alabaster==0.7.13--hash=sha256:1ee19aca801bbabb5ba3f5f258e4422dfa86f82f3e9cefb0859b283cdd7f62a3--hash=sha256:a27a4a084d5e690e16e01e03ad2b2e552c61a65469419b907243193de1a84ae2
             ~~~~~~~~^
```

PR #2339, opened 2023-07-24
[DM-40165] Increase Kafka data partition size

afausti:
- Kafka at base has two functions now, it stores topics from the teststand and topics replicated from the summit efd.
- Expand Kafka data partition from 500G to 1T


mareuter (approved):
This is fine, but we should make sure that IT knows we need the extra space.

PR #2351, opened 2023-07-25
DM-40060: Define application secrets in YAML, add basic CLI

rra:
Following the specification in SQR-079, add specifications for the Phalanx application secrets. Also add a JSON schema for those secrets files and enable schema checking in the pre-commit hook. Publish that JSON schema as part of the documentation using an extras directory and the `html_extra_path` directive.

Add a basic Phalanx command-line tool. The only action currently supported is to list all of the required secrets for a given environment, which is only marginally useful but serves as a test for all of the conditional secret logic.

This command-line tool uses all new models for the time being, despite substantial overlap with the models needed for documentation generation. These two models will be merged in the future, but the models required for secrets will be in flux as the secret management tool is developed, so the models are kept separate for now to avoid also churning the documentation code. Once the requirements for the models for secrets (and possibly the installer) are nailed down, merging the models with documentation will be easier.


rra:
For review purposes, a lot of this is just secret data or test data. Don't feel the need to look over that in detail; I'll be checking that after fleshing out the secret tool some more, and will catch any bugs at that point. This is just a starting point. The more interesting thing to review is the new code for importing the Phalanx configuration and analyzing the secret configuration.

src/phalanx/storage/config.py
```diff
@@ -0,0 +1,240 @@
...
+
+        Raises
+        ------
+        InvalidEnvironmentConfigError
+            Raised if the configuration for an environment is invalid.
+        UnknownEnvironmentError
+            Raised if the named environment has no configuration.
+        """
+        values_name = f"values-{environment_name}.yaml"
+        values_path = Path.cwd() / "environments" / values_name
```
jonathansick, line 164: This is reasonable for now, but I think for a next iteration we should "discover" the root directory of the phalanx repository and then compute paths within phalanx based on that. This would let us use the phalanx CLI within being at its root.

src/phalanx/models/secrets.py
```diff
@@ -0,0 +1,237 @@
...
+        if v is None and want_value:
+            msg = f"source not set for secret of type {secret_type}"
+            raise ValueError(msg)
+        if v is not None and not want_value:
+            msg = f"source not allowed for secret of type {secret_type}"
+            raise ValueError(msg)
+        return v
+
+
+class RequiredSecretConfig(BaseModel):
```
jonathansick, line 100: I'll say it took my brain a minute to grok the hierarchy of models, but I don't think  have any recommendations. Somehow making `SecretConfig` a subclass of `RequiredSecretConfig` and then making `RequiredSecret` and subclass of `SecretConfig` threw me 😂

src/phalanx/models/applications.py
```diff
@@ -0,0 +1,78 @@
...
+        ----------
+        condition
+            Condition to check.
+
+        Returns
+        -------
+        bool
+            `True` if the condition was met or does not exist, `False`
+            otherwise.
+        """
```
jonathansick, line 69: I think this method is totally fine, but I wonder if its naming/documentation is almost putting a secrets concern on the application? Like, what if it was named `has_truthy_value(self, value_path: str | None)` and then all the secrets code knows that conditions are based on finding truthy values at a value path?

I'm probably making a mountain out of a molehill, so feel free to ignore this.

jonathansick (approved):
This is really great 👍 I just have a few ideas:

***

Is the `secrets.json` JSON schema effectively a derivative of the secrets.py models? I wonder if we can use Pydantic to generate the JSON schema for us?

We could also code our own Pre-commit hook that uses the `phalanx` CLI for linting. I suppose that requires the whole `phalanx` being installed by pre-commit, so its a bit more complicated than just running the pre-commit hook.

***

For including the schemas in the docs, I have a preference to put the `extras/schemas` directory inside a new `_static` directory (i.e. `docs/_static/extra/schemas/secrets.json`. Then you don't have to manually configure an additional path for the Sphinx build in `conf.py`.

To reference these in the docs you'd use the `_static` source path prefix, even though Sphinx will move the extras/ directory to the root of the HTML output (I tested this now). E.g.

```rst
:download:`secrets.json <_static/extras/schemas/secrets.json>`

.. literalinclude:: _static/extras/schemas/secrets.json
   :language: json
```

I also wonder if the copy of the schemas in the docs should be a symlink to schemas in a source `/schemas` directory at the root of the repo so they're a little more discoverable?

***

Just a note, if I run `phalanx secrets list idfprod` now, the fact that any environment is missing secrets throws the command:

```
pydantic.error_wrappers.ValidationError: 7 validation errors for Application
environment_values -> ccin2p3
  none is not an allowed value (type=type_error.none.not_allowed)
environment_values -> idfint
  none is not an allowed value (type=type_error.none.not_allowed)
environment_values -> usdfdev
  none is not an allowed value (type=type_error.none.not_allowed)
environment_values -> usdfprod
  none is not an allowed value (type=type_error.none.not_allowed)
environment_values -> minikube
  none is not an allowed value (type=type_error.none.not_allowed)
environment_values -> idfdev
  none is not an allowed value (type=type_error.none.not_allowed)
environment_values -> idfprod
  none is not an allowed value (type=type_error.none.not_allowed)
```

This probably won't be an issue, but I wonder if we want to make the command resilient to environments that don't have working models yet. I also wonder why its only erroring on these environments?

src/phalanx/models/secrets.py
rra: Yeah, this was bothering me too. I just reworked the models to hopefully clear this up and to not resolve conditions in more than one place, and I think the result is a bit cleaner.

src/phalanx/models/applications.py
rra: This is a good idea. I turned this into `is_values_setting_true` and added a wrapper to deal with the null condition case. Thanks!

rra:
The reason why I didn't use a `_static` directory for the schema was because I wanted to give it a nice URL that can be referenced if needed. The current approach produces https://phalanx.lsst.io/schemas/secrets.json. I had thought that using `_static` would produce a URL of https://phalanx.lsst.io/_static/schemas/secrets.json. Was I wrong about that?

I can't imagine ever wanting to include the schema JSON directly in an rST file, since it's just a giant unreadable blob of JSON to humans, but maybe I'm missing something?

Generating the schema from the Pydantic models is a great idea and I'll look at finding a way to do that before merging this. You hit on exactly why I didn't want to try to run our code from pre-commit: it felt a bit too awkward, particularly when ensuring that pre-commit wouldn't fail if one ran pre-commit while currently in the middle of code changes that cause the code to temporarily not work (I do that all the time to commit WIP checkpoints).

rra:
Oh, wait, maybe what I'm missing is that the `_static` directory is *already* configured as an extras directory? Yeah, if that's the case, I definitely don't need to introduce a second one; I just didn't realize that was the case.

rra:
I don't think I understand how `_static` works. I moved the schema to `docs/_static/schemas/secrets.json`, but that doesn't appear to be served anywhere in the generated documentation. Neither https://phalanx.lsst.io/v/DM-40060/schemas/secrets.json nor https://phalanx.lsst.io/v/DM-40060/_static/schemas/secrets.json exist. If I go back to using `html_extras_dir`, then the `/schemas/secrets.json` route works as expected. I've therefore stuck with that for right now, but please let me know if I'm missing something.

src/phalanx/storage/config.py
rra: Yup, definitely a good idea and on my to-do list.

PR #2362, opened 2023-07-27
disable dev nublado2

athornton:


environments/values-idfint.yaml
```diff
@@ -28,7 +28,7 @@ linters:
 mobu:
   enabled: true
 moneypenny:
-  enabled: true
+  enabled: false
```
rra, line 31: I think we still need this, though, right?

athornton: Whoops.  Yes.

rra (approved):

PR #2366, opened 2023-07-28
DM-40125: Add support for static secret template

rra:
Add a phalanx secrets static-template command that generates a YAML template for the static secrets for an environment.


rra:
The weird change to the documentation heading is because otherwise the pre-commit hook looking for merge conflicts thinks it's a conflict marker. Making it one `=` longer works around the misidentification.

jonathansick (approved):
Looks great. Yeah, it's too bad about the conflict between rST headings and the merge mark linter.

PR #2373, opened 2023-07-31
DM-40125: Add basic Vault support to the Phalanx CLI

rra:
Add secrets vault-secrets and secrets audit commands. The former generates JSON versions of the Vault contents of a given environment and saves them to disk, used primarily for testing and debugging. The latter checks the list of secrets required by a given environment against the list of secrets in Vault, and reports any cases where secrets are missing, incorrect, or unexpectedly present.


jonathansick (approved):
Looks awesome 👍

PR #2376, opened 2023-08-02
DM-40256: Add Vault secret sync support

rra:
Add support for a YAML file containing static application secrets to the Phalanx command line, and use that support for auditing. Add a new sync command that will update Vault with correct static, generated, and copied secrets, and add a `--regenerate` flag to that command that forces regeneration of all generated secrets. This is a slight departure from SQR-079 (which will be updated) by adding a flag rather than a separate command; the flag looked better based on the code structure.

Fix handling of applications for which their Vault secret is completely missing. Previously that threw uncaught exceptions from the hvac layer.

No longer allow unknown static secrets during secret resolution now that we have a way to provide static secrets. Report those as unresolved secrets instead.

Move the configuration of the Vault URL into the environment configuration instead of only the vault-secrets-operator application configuration so that it's accessible to the Phalanx command-line tool where it expects to find it. Inject that value into the vault-secrets-operator chart using Argo CD.

Regenerate Python dependencies to fix a security hole in the cryptography library so that dependabot will stop complaining.


jonathansick (approved):
Looks great, no notes. :+1:

PR #2377, opened 2023-08-03
[DM-40192] Changes to chart for on prem S3 @ USDF

cbanek:
Be explicit about the S3_ENDPOINT_URL which we need to be because we're not using AWS.  Be explicit about the PGUSER so butler can figure out which address
to contact in the PGPASSFILE.

### DIFF
--- a/applications/datalinker/values.yaml
+++ b/applications/datalinker/values.yaml
@@ -43,6 +43,17 @@ config:
   # -- URL containing TAP schema metadata used to construct queries
   tapMetadataUrl: "https://github.com/lsst/sdm_schemas/releases/download/1.2.0/datalink-columns.zip"
 
+  # -- Storage backend to use: either GCS or S3
+  # GCS is the default
+  storageBackend: "GCS"
+
+  # -- User to use from the PGPASSFILE
+  # rubin is the default
+  pgUser: "rubin"
+
+  # -- S3 Endpoint URL
+  s3EndpointUrl: "https://storage.googleapis.com"
+
 # -- Annotations for the datalinker deployment pod
 podAnnotations: {}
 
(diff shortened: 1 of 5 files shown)


applications/datalinker/values.yaml
```diff
@@ -43,6 +43,17 @@ config:
   # -- URL containing TAP schema metadata used to construct queries
   tapMetadataUrl: "https://github.com/lsst/sdm_schemas/releases/download/1.2.0/datalink-columns.zip"
 
+  # -- Storage backend to use: either GCS or S3
+  # GCS is the default
+  storageBackend: "GCS"
+
+  # -- User to use from the PGPASSFILE
+  # rubin is the default
+  PGUSER: "rubin"
```
rra, line 52: I have a minor preference for naming this more like a typical Helm values parameter, like `postgresUser` or something, although I don't feel strongly about it.

rra (approved):

applications/datalinker/values.yaml
cbanek: Yeah I get where you're coming from and I'll do it that way, but it seems like if all you're doing is passing down environment variables it's really clear and obvious that you're passing down an environment variable and what it is.  Coming up with another name feels like we're taking one thing and giving it multiple names just for the reason of matching a case style.

rra: I think both of those ways of thinking about it are reasonable; I would just pick one or the other. In other words, if the goal is to provide a Helm chart mechanism to pass down arbitrary environment variables, I'd have something like `config.environment` that's arbitrary key-value pairs, all of which get stuffed into the pod environment. But if the goal is to tell datalinker what PostgreSQL user to authenticate as, that feels like an API boundary, and the details of that would be hidden (maybe some day we'll want to pass that in explicitly to the PostgreSQL client library instead of using an environment variable, who knows). This just felt a bit between two stools: it's a single setting, but it exposes that it's actually an environment variable.

This is all a *very* minor nit, though, and I'm totally fine with however you want to do it.

PR #2380, opened 2023-08-03
DM-40256: Add support for cleaning up Vault secrets

rra:
Add support for a `--delete` flag to `phalanx secrets sync` that removes stray secrets in Vault, either ones for unknown or no-longer-enabled applications or individual keys for applications no longer configured to have a key by that name.

Also fix `phalanx secrets vault-secrets` to read all secrets from Vault, not just the ones that correspond to enabled applications.


jonathansick (approved):
Looks great!

PR #2384, opened 2023-08-04
[DM-40192] Expose livetap tap service in portal

cbanek:
This is because the livetap service has all the datalinks


rra (approved):
Looks good to me!

PR #2385, opened 2023-08-04
tickets/DM-40289: enable nublado3 at T&S sites

athornton:


applications/nublado/values-base.yaml
```diff
@@ -0,0 +1,78 @@
...
+    every: 300
+    maxAge: 2160000
+  singleuser:
+    extraAnnotations:
+      k8s.v1.cni.cncf.io/networks: "kube-system/dds"
+  hub:
+    baseUrl: "/n3"
+    db:
+      upgrade: true
+      url: "postgresql://jovyan@postgresdb01.ls.lsst.org/jupyterhub"
```
rra, line 78: This, and likewise elsewhere, is the same URL as is used by nublado2, but you're installing this in parallel at `/n3`. I don't think this can possibly work; if both hubs are running at the same time, they need to have separate databases.

athornton: RIght you are.

applications/nublado/values-base.yaml
```diff
@@ -0,0 +1,78 @@
...
+jupyterhub:
+  cull:
+    users: false
+    removeNamedServers: false
+    timeout: 432000
+    every: 300
+    maxAge: 2160000
+  singleuser:
+    extraAnnotations:
+      k8s.v1.cni.cncf.io/networks: "kube-system/dds"
```
rra, line 73: Is this supposed to be an additional annotation to add to the created pod? If so, does this work? I feel like it couldn't work, since the Nublado controller is creating the pod and it doesn't know anything about this setting.

athornton: Uh oh.  Something like this is going to need to work, which may require adding a new config setting to the controller.

Yes, this is the magic annotation on the pod that causes it to couple to the correct multus-provided network.  It is untested, because only T&S sites use this feature.

athornton: OK, it we pass "annotations" into create_pod() it will do the right thing.  Not sure if we've already got annotations in the pod config, but if not it's easy to add.

athornton: Yeah, we'll have to add it as a config setting and then pipe it down through create_user_pod

athornton: cf https://github.com/lsst-sqre/jupyterlab-controller/pull/148

athornton: I've done that now.

rra (approved):

PR #2393, opened 2023-08-10
DM-40343: Clean up environment configuration

rra:
- Rename `repoURL` to `repoUrl` to match our conventions and the expectations of `CamelCaseModel` for Pydantic
- Delete the strimzi-registry-operator application, since this is now only used as a subchart and was disabled in every environment
- Remove the explicit kubernetes finalizer for namespaces, since this is added implicitly and isn't necessary to specify
- Move the fileserver namespace setup to its application configuration
- Add explicit namespace management for vault-secrets-operator
- Add quoting for all of the application configurations for consistency
- Add `values.yaml` to the list of values files for alert-stream-broker, since it seems to be in use.


rra:
Most of this is mechanical and doesn't need review, but I wanted to double-check the removal of strimzi-registry-operator with someone to be sure that is correct and we switched to only using this as a subchart.

jonathansick (approved):
I do think it's correct that the strimzi-registry-operator app can be dropped. Sasquatch never used this top-level app, and alert-stream-broker now seems to be installing the schema registry operator as a subchart at https://github.com/lsst-sqre/phalanx/blob/1a4e22b12bd35ce5bf283a38c3dab1db14fed468/applications/alert-stream-broker/Chart.yaml#L30-L32

PR #2399, opened 2023-08-14
DM-23487: Add support for creating Vault credentials

rra:
Add phalanx vault create-read-approle and phalanx vault create-write-token commands to manage Vault credentials. The current plan is to use AppRoles for vault-secrets-operator and tokens for the phalanx secrets commands.

Switch the test data to using the phalanx Vault path prefix instead of k8s_operator, since we'll be setting up the new Vault keys in a new path when doing the cutover.


jonathansick (approved):
Looks great 👍

PR #2405, opened 2023-08-16
DM-25701: Add phalanx vault audit command

rra:
Add a command to check the authentication credentials for a given environment (both the read AppRole and the write token) and report any problems found. Add a display name to write tokens so that they can be more easily found among all of the other accessors.


jonathansick (approved):
Looks great 👍

PR #2414, opened 2023-08-17
DM-40343: Detect location of the Phalanx configuration

rra:
Add a flag to every phalanx CLI command that uses the Phalanx configuration specifying its location, and default to searching upwards from the current directory to find the top of the configuration tree (by looking for the applications and environments directories).

Add a wrapper function around Click command-line testing for the test suite that adds the path to the test configuration directory and flips the default for catching exceptions to False.

### DIFF
--- a/src/phalanx/cli.py
+++ b/src/phalanx/cli.py
@@ -27,6 +27,37 @@ __all__ = [
 ]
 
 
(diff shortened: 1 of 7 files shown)


jonathansick (approved):
Awesome, thanks 👍

PR #2416, opened 2023-08-17
DM-40343: Merge documentation models with CLI models

rra:
Enhance the models used for the Phalanx CLI so that they, or related additional models, can provide all the information needed by documentation generation. Add a new helper function for building the Jinja templates that do most of the heavy lifting. Remove the old code and switch the documentation build over to these models.

This involved some minor renaming, reshuffling, and moving a few things into methods instead of properties. Update the Jinja templates accordingly. The models no longer load README.md for each chart, since this was handled a different way using other Sphinx features.


jonathansick (approved):
Awesome, thanks for cleaning up and improving the documentation generation.

PR #2419, opened 2023-08-17
tickets/DM-40149: Add RubinTV

athornton:
Create rubintv app and get it running at usdf-dev

### DIFF
--- a/applications/rubintv/Chart.yaml
+++ b/applications/rubintv/Chart.yaml
@@ -5,3 +5,7 @@ description: Continuous integration testing
 sources:
   - https://github.com/lsst-sqre/rubin-tv
 appVersion: 0.1.0
+dependencies:
+  - name: redis
+    version: 1.0.6
+    repository: https://lsst-sqre.github.io/charts/
--- a/applications/rubintv/Chart.yaml
+++ b/applications/rubintv/Chart.yaml
@@ -1,0 +1,7 @@
+apiVersion: v2
+name: rubintv
+version: 1.0.0
+description: Continuous integration testing
+sources:
+  - https://github.com/lsst-sqre/rubin-tv
+appVersion: 0.1.0
--- a/applications/rubintv/Chart.yaml
+++ b/applications/rubintv/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: rubintv
 version: 1.0.0
-description: Continuous integration testing
+description: real-time display front end
 sources:
   - https://github.com/lsst-sqre/rubin-tv
 appVersion: 0.1.0
--- a/applications/rubintv/templates/deployment-frontend.yaml
+++ b/applications/rubintv/templates/deployment-frontend.yaml
@@ -27,12 +27,12 @@ spec:
             - name: "ENVIRONMENT_URL"
               value: {{ .Values.global.baseUrl }}
             - name: "SAFIR_PATH_PREFIX"
-              value: {{ .Values.frontend.config.pathPrefix | quote }}
-            {{- if (not .Values.frontend.config.debug) }}
+              value: {{ .Values.frontend.pathPrefix | quote }}
+            {{- if (not .Values.frontend.debug) }}
             - name: "SAFIR_PROFILE"
               value: "production"
             {{- end }}
-            {{- with .Values.frontend.config.env }}
+            {{- with .Values.frontend.env }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
           image: "{{ .Values.frontend.image.repository }}:{{ .Values.frontend.image.tag | default .Chart.AppVersion }}"
--- a/applications/rubintv/templates/deployment-frontend.yaml
+++ b/applications/rubintv/templates/deployment-frontend.yaml
@@ -61,12 +61,33 @@ spec:
           volumeMounts:
             - name: "tmp"
               mountPath: "/tmp"
+      initContainers:
+      - name: "secret-perm-fixer"
+        image: "busybox"
+        command:
+        - "/bin/sh"
+        - "-c"
+        - |
+          cp /secrets/* /etc/secrets && \
+          chown {{ $.Values.workers.uid }}:{{ $.Values.workers.gid }} \
+           /etc/secrets/* && \
+          chmod 0600 /etc/secrets/*
+        volumeMounts:
+        - name: "secrets"
+          mountPath: "/secrets"
+        - name: "user-secrets"
+          mountPath: "/etc/secrets"
       securityContext:
         runAsNonRoot: true
         runAsUser: 1000
         runAsGroup: 1000
       volumes:
         - name: "tmp"
+          emptyDir: {}
+        - name: "secrets"
+          secret:
+            secretName: {{ include "rubintv.fullname" . }}-secret
+        - name: "user-secrets"
           emptyDir: {}
       {{- with .Values.frontend.nodeSelector }}
       nodeSelector:
--- a/applications/rubintv/templates/deployment-frontend.yaml
+++ b/applications/rubintv/templates/deployment-frontend.yaml
@@ -28,6 +28,8 @@ spec:
               value: {{ .Values.global.baseUrl }}
             - name: "SAFIR_PATH_PREFIX"
               value: {{ .Values.frontend.pathPrefix | quote }}
+            - name: "AWS_SHARED_CREDENTIALS_FILE"
+              value: /etc/secrets/aws-credentials.ini
             {{- if (not .Values.frontend.debug) }}
             - name: "SAFIR_PROFILE"
               value: "production"
--- a/applications/rubintv/templates/deployment-frontend.yaml
+++ b/applications/rubintv/templates/deployment-frontend.yaml
@@ -27,10 +27,13 @@ spec:
             - name: "ENVIRONMENT_URL"
               value: {{ .Values.global.baseUrl }}
             - name: "SAFIR_PATH_PREFIX"
-              value: {{ .Values.ingress.pathPrefix | quote }}
-            {{- if (not .Values.workers.config.debug) }}
+              value: {{ .Values.frontend.config.pathPrefix | quote }}
+            {{- if (not .Values.frontend.config.debug) }}
             - name: "SAFIR_PROFILE"
               value: "production"
+            {{- end }}
+            {{- with .Values.frontend.config.env }}
+            {{- toYaml . | nindent 12 }}
             {{- end }}
           image: "{{ .Values.frontend.image.repository }}:{{ .Values.frontend.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.frontend.image.pullPolicy | quote }}
--- a/applications/rubintv/templates/deployment-frontend.yaml
+++ b/applications/rubintv/templates/deployment-frontend.yaml
@@ -1,0 +1,77 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "rubintv.fullname" . }}
+  labels:
+    {{- include "rubintv.labels" . | nindent 4 }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      {{- include "rubintv.selectorLabels" . | nindent 6 }}
+  strategy:
+    type: "Recreate"
+  template:
+    metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "rubintv.selectorLabels" . | nindent 8 }}
+    spec:
+      automountServiceAccountToken: false
+      containers:
+        - name: {{ .Chart.Name }}
+          env:
+            - name: "ENVIRONMENT_URL"
+              value: {{ .Values.global.baseUrl }}
+            - name: "SAFIR_PATH_PREFIX"
+              value: {{ .Values.config.pathPrefix | quote }}
+            {{- if (not .Values.config.debug) }}
+            - name: "SAFIR_PROFILE"
+              value: "production"
+            {{- end }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
+          ports:
+            - name: "http"
+              containerPort: 8080
+              protocol: "TCP"
+          readinessProbe:
+            httpGet:
+              path: "/"
+              port: "http"
+            timeoutSeconds: 10
+          {{- with .Values.resources }}
+          resources:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
+            readOnlyRootFilesystem: true
+          volumeMounts:
+            - name: "tmp"
+              mountPath: "/tmp"
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 1000
+        runAsGroup: 1000
+      volumes:
+        - name: "tmp"
+          emptyDir: {}
+      {{- with .Values.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
--- a/applications/rubintv/templates/deployment-frontend.yaml
+++ b/applications/rubintv/templates/deployment-frontend.yaml
@@ -61,26 +61,27 @@ spec:
           volumeMounts:
             - name: "tmp"
               mountPath: "/tmp"
-      initContainers:
-      - name: "secret-perm-fixer"
-        image: "busybox"
-        command:
-        - "/bin/sh"
-        - "-c"
-        - |
-          cp /secrets/* /etc/secrets && \
-          chown {{ $.Values.workers.uid }}:{{ $.Values.workers.gid }} \
-           /etc/secrets/* && \
-          chmod 0600 /etc/secrets/*
-        volumeMounts:
-        - name: "secrets"
-          mountPath: "/secrets"
-        - name: "user-secrets"
-          mountPath: "/etc/secrets"
+            - name: "user-secrets"
+              mountPath: "/etc/secrets"
       securityContext:
         runAsNonRoot: true
         runAsUser: 1000
         runAsGroup: 1000
+      initContainers:
+        - name: "secret-perm-fixer-frontend"
+          image: "busybox"
+          command:
+            - "/bin/sh"
+            - "-c"
+            - |
+              cp /secrets/* /etc/secrets && \
+              chown 1000:1000 /etc/secrets/* && \
+              chmod 0600 /etc/secrets/*
+          volumeMounts:
+            - name: "secrets"
+              mountPath: "/secrets"
+            - name: "user-secrets"
+              mountPath: "/etc/secrets"
       volumes:
         - name: "tmp"
           emptyDir: {}
--- a/applications/rubintv/templates/deployment-frontend.yaml
+++ b/applications/rubintv/templates/deployment-frontend.yaml
@@ -19,7 +19,7 @@ spec:
         {{- include "rubintv.selectorLabels" . | nindent 8 }}
     spec:
       automountServiceAccountToken: false
-      {{- with .Values.workers.imagePullSecrets }}
+      {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
       {{ toYaml . | nindent 8 }}
       {{- end }}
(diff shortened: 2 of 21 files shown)


applications/rubintv/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+name: rubintv
+version: 1.0.0
+description: Continuous integration testing
```
rra, line 4: Looks like a copy-paste description here

applications/rubintv/templates/deployment-frontend.yaml
```diff
@@ -0,0 +1,80 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "rubintv.fullname" . }}-frontend
+  labels:
+    {{- include "rubintv.labels" . | nindent 4 }}
+spec:
+  replicas: 1
```
rra, line 8: The current version on Roundtable is running with 5 replicas. Should this be configurable?

applications/rubintv/templates/deployment-frontend.yaml
```diff
@@ -0,0 +1,80 @@
...
+  selector:
+    matchLabels:
+      {{- include "rubintv.selectorLabels" . | nindent 6 }}
+  strategy:
+    type: "Recreate"
```
rra, line 13: Why this? Normally you don't want this, since it forces an outage whenever the application is upgraded.

applications/rubintv/templates/deployment-frontend.yaml
```diff
@@ -0,0 +1,80 @@
...
+      {{- with .Values.frontend.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "rubintv.selectorLabels" . | nindent 8 }}
+    spec:
+      automountServiceAccountToken: false
+      containers:
+        - name: {{ .Chart.Name }}-frontend
```
rra, line 25: These don't need to be globally unique, so rather than make a complicated name that would be annoying if we ever had to specify it explicitly with the `-c` option, you can just use `frontend` here.

applications/rubintv/templates/deployment-frontend.yaml
```diff
@@ -0,0 +1,80 @@
...
+          env:
+            - name: "ENVIRONMENT_URL"
+              value: {{ .Values.global.baseUrl }}
```
rra, line 28: Add `| quote` here as well.

applications/rubintv/templates/deployment-frontend.yaml
```diff
@@ -0,0 +1,80 @@
...
+          image: "{{ .Values.frontend.image.repository }}:{{ .Values.frontend.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.frontend.image.pullPolicy | quote }}
+          ports:
+            - name: "http"
+              containerPort: 8080
+              protocol: "TCP"
+          readinessProbe:
+            httpGet:
+              path: "/"
+              port: "http"
```
rra, line 47: We didn't do this previously because getting `/` is expensive. Is that still the case?

applications/rubintv/templates/deployment-frontend.yaml
```diff
@@ -0,0 +1,80 @@
...
+          {{- end }}
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
+            readOnlyRootFilesystem: true
+          volumeMounts:
+            - name: "tmp"
+              mountPath: "/tmp"
```
rra, line 61: Do you know if this is required? The Roundtable configuration does not mount `/tmp`.

applications/rubintv/templates/deployment-workers.yaml
```diff
@@ -0,0 +1,134 @@
...
+  name: {{ template "rubintv.fullname" . }}-workers
+  labels:
+    {{- include "rubintv.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.workers.replicas }}
+  selector:
+    matchLabels:
+      {{- include "rubintv.selectorLabels" . | nindent 6 }}
+  strategy:
+    type: "Recreate"
```
rra, line 13: Here too I don't think recreate makes sense.

applications/rubintv/templates/deployment-workers.yaml
```diff
@@ -0,0 +1,134 @@
...
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "rubintv.selectorLabels" . | nindent 8 }}
+    spec:
+      automountServiceAccountToken: false
+      containers:
+      - name: {{ .Chart.Name }}-worker
```
rra, line 25: Can just be `worker` as mentioned above

applications/rubintv/templates/deployment-workers.yaml
```diff
@@ -0,0 +1,134 @@
...
+        - |
+          cp /secrets/* /etc/secrets && \
+          chown {{ $.Values.workers.uid }}:{{ $.Values.workers.gid }} \
+           /etc/secrets/* && \
+          chmod 0600 /etc/secrets/*
+        volumeMounts:
+        - name: "secrets"
+          mountPath: "/secrets"
+        - name: "user-secrets"
+          mountPath: "/etc/secrets"
```
rra, line 103: The better way to do this is to have the image startup script copy and chmod the secrets (see datalinker), but that requires getting that change into rubintv itself, so this is probably fine.

applications/rubintv/templates/deployment-workers.yaml
```diff
@@ -0,0 +1,134 @@
...
+        {{- end }}
+      initContainers:
+      - name: "secret-perm-fixer"
+        image: "busybox"
+        command:
+        - "/bin/sh"
+        - "-c"
+        - |
+          cp /secrets/* /etc/secrets && \
+          chown {{ $.Values.workers.uid }}:{{ $.Values.workers.gid }} \
```
rra, line 96: Shouldn't need the `$` here, since this isn't inside some other context that lost `.Values`, and having it is confusing (it makes me wonder what's going on).

applications/rubintv/templates/deployment-workers.yaml
```diff
@@ -0,0 +1,134 @@
...
+      - name: {{ $vol.name | quote }}
+        {{ with $vol.persistentVolumeClaim }}
+        persistentVolumeClaim:
+          claimName: {{ .name | quote }}
+        {{- end }}
+      {{- end }}
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: {{ $.Values.workers.uid }}
+        runAsGroup: {{ $.Values.workers.gid }}
```
rra, line 122: Likewise here, shouldn't need the `$`.

applications/rubintv/templates/ingress.yaml
```diff
@@ -0,0 +1,31 @@
...
+kind: GafaelfawrIngress
+metadata:
+  name: {{ template "rubintv.fullname" . }}
+  labels:
+    {{- include "rubintv.labels" . | nindent 4 }}
+config:
+  baseUrl: {{ .Values.global.baseUrl | quote }}
+  scopes:
+    all:
+      - "exec:notebook"
```
rra, line 11: This definitely shouldn't be `exec:notebook`. I'm not sure off-hand what it should be to read embargo data. We may need a separate scope for that. If everyone at the USDF can read embargo data, then maybe `read:image` is fine, but I'm not sure that's the case.

applications/rubintv/values-usdfdev.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+frontend:
+  debug: true
+  env:
+  - name: http_proxy
+    value: "http://sdfproxy.sdf.slac.stanford.edu:3128"
+  - name: https_proxy
+    value: "http://sdfproxy.sdf.slac.stanford.edu:3128"
+  - name: no_proxy
+    value: ".sdf.slac.stanford.edu,.slac.stanford.edu,localhost,127.0.0.1"
```
rra, line 11: I feel like SLAC found some way to not have to embed these settings in every application, but I don't remember what it was.

applications/rubintv/values-usdfdev.yaml
```diff
@@ -0,0 +1,89 @@
...
+  - name: DAF_BUTLER_REPOSITORY_INDEX
+    value: "s3://rubin-summit-users/data-repos.yaml"
+  - name: DAF_BUTLER_REPOSITORY
+    value: "/sdf/group/rubin/repo/ir2/butler.yaml"
+  envSecrets:
+  - name: ASTROMETRY_NET_API_KEY
+    secretName: rubintv-secret
+    secretKey: astrometry-net-api-key
+  imagePullSecrets:
+  - name: pull-secret
```
rra, line 54: Maybe I'm blind, but I don't see where this is used in the template.

applications/rubintv/values-usdfdev.yaml
```diff
@@ -0,0 +1,89 @@
...
+      memory: "8Gi"
+  # credentialFile: google_write_creds
+  # vaultPrefixPath: secret/rubin/usdf-rubintv-broadcaster-cleanroom
+  # pullSecretsPath: ts/software/nexus3-docker
+  # rubinTvSecretsPath: ts/software/rubintv
+  # credentialSecretsPath: rubintv-broadcaster
+  # butlerSecret:
+  #   key: butler-secret
+  #   containerPath: /secrets
+  #   dbUser: rubin
```
rra, line 89: Shouldn't commit commented-out lines to Git.

applications/rubintv/values-usdfprod.yaml
rra, line 1: This doesn't seem useful. I'd delete this file until there's a real configuration to make it clear this hasn't been configured for that environment yet.

applications/rubintv/values.yaml
```diff
@@ -0,0 +1,182 @@
...
+fullnameOverride: ""
+
+# -- A special tag for letting the scripts know where they are
+# running.  Must be overridden at each site
+siteTag: ""
+
+frontend:
+  # -- If set to true, enable more verbose logging.
+  debug: false
+  # -- Prefix for rubintv's frontend API routes.
```
rra, line 16: Here and elsewhere, blank lines between settings before the `# --` line unless it's the first setting of a nested block, otherwise before the key of the nested block, is our formatting standard.

applications/rubintv/values-usdfdev.yaml
```diff
@@ -0,0 +1,89 @@
...
+  image:
+    tag: tickets-DM-40149
+    pullPolicy: Always
+
+workers:
+  replicas: 1
+  image:
+    repository: lsstts/rubintv-broadcaster
+    tag: c0030_usdf
+    pullPolicy: Always
```
rra, line 21: I'm confused about why on tucson-teststand we get this package from ts-dockerhub but here we get it from Docker Hub. Where is this package actually maintained?

applications/rubintv/values.yaml
```diff
@@ -0,0 +1,182 @@
...
+  # -- how many replicas to use
+  replicas: 5
+  # -- If set to true, enable more verbose logging.
+  debug: false
+  # -- Prefix for the (internal) worker API routes
+  pathPrefix: "/"
+  # -- UID to run as (site-dependent, because of filesystem perms)
+  uid: 1000
+  # -- GID to run as
+  gid: 1000
```
rra, line 48: I would set these to null and check in the template whether they're set and throw an error if they're not, similar to how global.host is handled in the ingress. (Also these values are being used for tucson-teststand; surely that's not correct?)

applications/rubintv/values-tucson-teststand.yaml
```diff
@@ -0,0 +1,3 @@
+fontend:
+  config:
+    debug: true
```
rra, line 3: Is this in use? Some things below make me think it's not, so it should probably be deleted.

applications/rubintv/values-summit.yaml
rra, line 1: I don't think we're running this at the summit, so delete this file.

applications/rubintv/values.yaml
```diff
@@ -0,0 +1,182 @@
...
+  env: []
+  # -- This section holds specifications for secret injection.
+  # If this section is used, each object listed must have the following attributes defined:
+  # _name_ (The label for the secret),
+  # _secretName_ (The name of the vault store reference. Uses the _namespace_ attribute to construct the full name),
+  # _secretKey_ (The key in the vault store containing the necessary secret)
+  envSecrets: []
+  # -- The Vault prefix path
+  vaultPrefixPath: ""
+  # The key for the pull secrets including any sub-paths.
```
rra, line 73: Comment has to start with `# --` for it to be picked up by documentation.

applications/rubintv/values.yaml
```diff
@@ -0,0 +1,182 @@
...
+  pullSecretsPath: ""
+  # Path for the rubin tv specific secrets vault.  ** Let's just stuff
```
rra, line 75: Comment has to start with `# --`.

applications/rubintv/values.yaml
```diff
@@ -0,0 +1,182 @@
...
+  script: slac/rubintv/workerPod1.py
+  # -- This section holds a list of key, value pairs for environmental
+  # variables (name: key, value: value).
+  # NOTE: RUN_ARG is taken care of by the chart using _script_.
+  env: []
+  # -- This section holds specifications for secret injection.
+  # If this section is used, each object listed must have the following attributes defined:
+  # _name_ (The label for the secret),
+  # _secretName_ (The name of the vault store reference. Uses the _namespace_ attribute to construct the full name),
+  # _secretKey_ (The key in the vault store containing the necessary secret)
```
rra, line 69: Here and below, these complex comments are not going to render in any useful way. I think it would be better to put an example into a comment, and then have the documentation comment say to look at `values.yaml` for an example.

docs/applications/rubintv/index.rst
```diff
@@ -0,0 +1,23 @@
+.. px-app:: rubintv
+
+##########################
+rubintv
+##########################
+
+rubintv is the display front end for ... well, stuff.  There are also
+some worker pods supporting that front-end pod.  Merlin Fisher-Levine
+understands it, but the rest of us really don't.
```
rra, line 9: I think merlin provided an updated introduction. Also, the title of every application must be `name — short description`

docs/applications/rubintv/index.rst
```diff
@@ -0,0 +1,23 @@
...
+
+.. jinja:: rubintv
+   :file: applications/_summary.rst.jinja
+
+Guides
+======
+
```
rra, line 16: Extra guide heading

environments/templates/rubintv-application.yaml
```diff
@@ -0,0 +1,37 @@
+{{- if .Values.rubintv.enabled -}}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: "rubintv"
+spec:
+  finalizers:
+    - "kubernetes"
```
rra, line 8: Delete this; I cleaned this up already for all the other applications, since it isn't needed.

environments/templates/rubintv-application.yaml
```diff
@@ -0,0 +1,37 @@
...
+  finalizers:
+    - "resources-finalizer.argocd.argoproj.io"
+spec:
+  destination:
+    namespace: "rubintv"
+    server: "https://kubernetes.default.svc"
+  project: "default"
+  source:
+    path: "applications/rubintv"
+    repoURL: {{ .Values.repoURL | quote }}
```
rra, line 24: Must now be
```suggestion
    repoURL: {{ .Values.repoUrl | quote }}
```

rra (commented):
The way applications are configured for environments is changing in https://github.com/lsst-sqre/phalanx/pull/2426 so this will probably need a rebase against that if that PR is approved soon.

applications/rubintv/templates/deployment-frontend.yaml
athornton: I don't think so.  Based on my understanding of the app, the extra replicas will now be worker pods, and there is a single frontend pod serving images.  If I'm wrong, we'll change it then.

applications/rubintv/templates/deployment-frontend.yaml
athornton: Probably pasted from the existing rubintv?  And I don't know why.  But let's drop it.

applications/rubintv/templates/deployment-frontend.yaml
athornton: I don't know, actually.  We're doing this on the JupyterLab controller as a readiness probe and I thought it was how we were planning on doing FastAPI healthchecks.

applications/rubintv/templates/deployment-workers.yaml
athornton: These were left over from adapting rapid-analysis, where all of this is inside a loop over startup scripts.

applications/rubintv/templates/ingress.yaml
athornton: I'll change to `read:image` for now and ask about embargo data on SLAC.  Merlin thought it was "anyone who can get an RSP notebook" hence this value.

applications/rubintv/values-usdfdev.yaml
athornton: If they did I have no memory of it.  I've asked on Slack.

applications/rubintv/values-usdfdev.yaml
athornton: ...note that it's the _usdf tag, not merely c0030, and...beats me.  It's somewhere in the T&S Jenkins I assume.  I don't know where they push which images.  This particular part was on fire when I got here.  This is probably also why there's a pull secret in the config.

applications/rubintv/values-usdfprod.yaml
athornton: summit/usdfprod/tts all going away for now.  usdfdev is the only place we're actually running thus far.

applications/rubintv/values-tucson-teststand.yaml
athornton: No, TTS isn't there and has been deleted.

applications/rubintv/values.yaml
athornton: Those particular keys just need to vanish.

applications/rubintv/values.yaml
athornton: I think I got that right on this second pass?  Right-ish?

applications/rubintv/values-usdfdev.yaml
athornton: Injected.

applications/rubintv/values-usdfdev.yaml
athornton: Apparently they are not required but do speed up things?  I'll change the frontend to remove them, anyway.

athornton: And workers too.  We can put 'em back if we really need them.

applications/rubintv/templates/ingress.yaml
athornton: KT says `read:image` should be fine for the time being.

applications/rubintv/values-usdfdev.yaml
athornton: And that's not a safe assumption.  Maybe Merlin builds and pushes it by hand; I wouldn't be the least bit surprised.

applications/rubintv/templates/deployment-frontend.yaml
```diff
@@ -0,0 +1,102 @@
...
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "rubintv.selectorLabels" . | nindent 8 }}
+    spec:
+      automountServiceAccountToken: false
+      {{- with .Values.workers.imagePullSecrets }}
+      imagePullSecrets:
+      {{ toYaml . | nindent 8 }}
```
rra, line 24: This is the frontend but it's using workers.imagePullSecrets. I suspect the right answer here is to just move the pullSecrets configuration up a level and share it between the workers and the frontend on the grounds that we probably aren't going to maintain separate pull secrets, but if you want to keep them separate, this should be frontend.pullSecrets.

applications/rubintv/templates/deployment-frontend.yaml
rra: I double-checked and it looks like this is hitting a metadata endpoint, so we should be okay.

applications/rubintv/templates/deployment-frontend.yaml
```diff
@@ -0,0 +1,102 @@
...
+      initContainers:
+        - name: "secret-perm-fixer-frontend"
+          image: "busybox"
+          command:
+            - "/bin/sh"
+            - "-c"
+            - |
+              cp /secrets/* /etc/secrets && \
+              chown 1000:1000 /etc/secrets/* && \
+              chmod 0600 /etc/secrets/*
```
rra, line 79: If there's no postgres secret, you don't need to do any of this. This is only required for postgres, and I think you said elsewhere that the frontend doesn't need postgres credentials? If that's true, you can delete all of this; the default mode for mounted secrets is 0644, and only the postgres libraries care about the ownership.

rra (commented):
rubintv also needs a `secrets.yaml` specification for its secrets. See:

https://phalanx.lsst.io/developers/define-secrets.html
https://phalanx.lsst.io/developers/secrets-spec.html

applications/rubintv/templates/deployment-frontend.yaml
athornton: I'll take it out and check that the S3 creds are still readable.  I feel like I started there and something failed, but I might be misremembering.

applications/rubintv/templates/deployment-frontend.yaml
athornton: I should lift it.

applications/rubintv/templates/deployment-frontend.yaml
athornton: Works fine.

rra (approved):
Yeah, this seems fine, at least to get started with. Thank you!

PR #2424, opened 2023-08-18
DM-40343: Add API and schema documentation

rra:
Add an internals page that is available only via the top bar from the pull-down menu, since most people will never need to see it. Document the JSON schemas there (currently only one for secrets.yaml but more will come later) and add internal API documentation.

This is of arguable usefulness, but sometimes it can be nice to browse the API documentation when working on new code, and it has the nice benefit of checking all of the docstrings more thoroughly than a linter does.

### DIFF
--- a/docs/_templates/autosummary_core/pydantic_model.rst
+++ b/docs/_templates/autosummary_core/pydantic_model.rst
@@ -1,0 +1,11 @@
+{% if referencefile %}
+.. include:: {{ referencefile }}
+{% endif %}
+
+{{ objname }}
+{{ underline }}
+
+.. currentmodule:: {{ module }}
+
+.. auto{{ objtype }}:: {{ objname }}
+   :inherited-members: BaseModel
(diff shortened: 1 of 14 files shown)


docs/_templates/autosummary_core/pydantic_model.rst
```diff
@@ -0,0 +1,11 @@
+{% if referencefile %}
```
jonathansick, line 1: I'm just curious, how did you know that a `pydantic_model.rst` template would get used as the template by the autodoc_pydantic documenter? I can't follow from the docs I've seen how this works.

jonathansick (approved):

docs/_templates/autosummary_core/pydantic_model.rst
rra: Alas, the answer is "spending a bunch of time reading the source code while thinking about making a PR to improve support." I figured it out as part of my research for SQR-077.

PR #2425, opened 2023-08-18
Add Prompt Processing Environment and Prompt Proto Service Application

dspeck1:
Add in environment for prompt processing, prompt proto service applications, and associated documentation.  A shared chart application for prompt processing is a subchart for HSC, LATISS, LSSTComCam, LSSTCam.


environments/values-usdfdev-prompt-processing.yaml
```diff
@@ -0,0 +1,13 @@
+fqdn: k8s.slac.stanford.edu/usdf-prompt-processing-dev
```
rra, line 1: The other code does need to be a FQDN, not a URL. I think this environment is currently disabling all of the applications for which this would matter (although I think Argo CD's ingress may break?), but I don't think we want to support this syntax and try to teach other things how to handle it going forward. Can SLAC assign usdf-prompt-processing-dev.slac.stanford.edu or the like to this environment?

dspeck1: Thanks!  Adjusted to usdf-prompt-processing-dev.slac.stanford.edu

afausti (approved):
This looks good, but creating an application per instrument feels a bit repetitive. I think this is a perfect case for using the ArgoCD ApplicationSet controller, you can have a List generator that would pass the instrument parameter to your application template. https://argo-cd.readthedocs.io/en/stable/user-guide/application-set/

dspeck1:
Thanks!  I had looked at application sets, but didn't want to deviate too much from established patterns for this.

PR #2426, opened 2023-08-18
DM-40448: Restructure application config for environments

rra:
Previously, enabling and disabling environments was done via a top-level block for each application in values-_environment_.yaml, with one key (enabled) that controlled whether it was enabled. This meant changing the top-level keys every time a new application was added, and made it difficult to do schema checking for the values files for each environment.

Instead, move the application configuration under a top-level applications key, and adjust the Argo CD configuration accordingly. Simplify the configuration syntax so that the applications key is just a list of applications as keys and booleans as values. Enable some applications by default (cert-manager, gafaelfawr, ingress-nginx, vault-secrets-operator), and add argocd to the list of applications defined and enabled in values.yaml to make parsing easier and avoid needing that special case. The enabled applications are now represented by a merger of the values.yaml and values-_environment_.yaml file. Drop unnecessary overrides, so each environment no longer needs to list and explicitly disable every application that it doesn't want.

The current installer doesn't entirely understand the new structure. Hack around that for now by embedding assumptions about which applications are always enabled. The installer will be replaced with new code that's aware of this structure later.


rra:
This is mostly mechanical. The code changes are probably the interesting bits.

jonathansick (approved):
Looks good, I like that the environment configurations are merged with global defaults. As for keeping the toggle to enable an application under a `enabled` field — It looks like you want to continue that pattern, eh? Will applications potentially have keys other than "enabled" that we're future proofing for? Or could we instead streamline the datastructure so that the application name itself takes the boolean value. Like, `squareone = true`.

PR #2431, opened 2023-08-21
tickets/DM-40454: Correct path to serverPath for T&S site n3 volume defs

athornton:
I had specified "path" everywhere I meant "serverPath"


athornton:
Tested on base.  I'm going to merge, as it only affects things that are a) currently broken and b) not yet in use.

PR #2438, opened 2023-08-23
DM-40448: Add schema checking for environments

rra:
Now that we can fully model the schema for the values files for environments, add a command to generate that schema, store the current version in the schema directory, and enable schema checking for those files.

Add a Python test that parses the full Phalanx configuration as if we were about to generate documentation. This is redundant with actually generating documentation, but the Python test suite is much faster to run, so this catches problems earlier when iterating.

### DIFF
--- a/docs/extras/schemas/environment.json
+++ b/docs/extras/schemas/environment.json
@@ -1,0 +1,55 @@
+{
+  "title": "EnvironmentConfig",
+  "description": "Configuration for a Phalanx environment.\n\nThis is a model for the :file:`values-{environment}.yaml` files for each\nenvironment and is also used to validate those files. For the complete\nconfiguration for an environment, initialize this model with the merger of\n:file:`values.yaml` and :file:`values-{environment}.yaml`.",
+  "type": "object",
+  "properties": {
+    "name": {
+      "title": "Name",
+      "type": "string"
+    },
+    "fqdn": {
+      "title": "Fqdn",
+      "type": "string"
+    },
+    "vaultUrl": {
+      "title": "Vaulturl",
+      "type": "string"
+    },
+    "vaultPathPrefix": {
+      "title": "Vaultpathprefix",
+      "type": "string"
+    },
+    "applications": {
+      "title": "Applications",
+      "type": "object",
+      "additionalProperties": {
+        "type": "boolean"
+      }
+    },
+    "butlerRepositoryIndex": {
+      "title": "Butlerrepositoryindex",
+      "type": "string"
+    },
+    "onepasswordUuid": {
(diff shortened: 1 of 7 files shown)


docs/internals/json-schema.rst
```diff
@@ -16,5 +16,7 @@ A schema for the current model can be generated with the appropriate Phalanx com
 
    * - Schema
      - Command
-   * - `secrets.yaml files </schemas/secrets.json>`__
+   * - `environment values files <../schemas/environment.json>`__
+     - :command:`phalanx environment schema`
+   * - `secrets.yaml files <../schemas/secrets.json>`__
```
jonathansick, line 21: Doing external links to internal files is somewhat unsupported in Sphinx, which I think is part of the reason linkcheck is failing on these. The two options are:

1. Use the download role to indicate this is a download. I think linkcheck will handle this? e.g.

```rst
:download:`/schemas/secrets.json`
```

2. If you want the schema to be browseable, then create an rst page for the schema itself and use `literalinclude` to show the schema and include a separate download link to make it easy to download the file.

docs/extras/schemas/environment.json
```diff
@@ -0,0 +1,55 @@
...
+  "required": [
+    "name",
+    "fqdn",
+    "vaultUrl",
+    "vaultPathPrefix",
+    "applications"
+  ],
+  "additionalProperties": false,
+  "$id": "https://phalanx.lsst.io/schemas/environment.json"
+}
```
jonathansick, line 55: I think it'd be good to have newline added here.

jonathansick (approved):
Very nice thanks. I have a comment on how the schema file is presented which might tie into the linkcheck workaround.

docs/internals/json-schema.rst
rra: I think download is exactly what I want; I just didn't know that existed. (I don't think the schemas themselves are all that readable, although they could be made better if I tweaked the metadata in the schema, which is on my long list of follow-up stuff.) Thank you! I'll give that a try.

docs/extras/schemas/environment.json
rra: Yeah, good call, I'll make sure the file ends with a newline.

docs/internals/json-schema.rst
rra: That still isn't *exactly* what I wanted since I have to refer to it by the `/extras/schemas/environment.json` path for the download role and it makes a second copy of the schema, but when combined with html_extra_path it functionally it does everything I want, so it's a definite improvement. Thanks!

PR #2439, opened 2023-08-23
DM-40448: Remove the tap-schema application

rra:
The individual tap applications now deploy their own schema services, so this central tap-schema application is no longer used.


cbanek (approved):
LGTM!

PR #2442, opened 2023-08-24
DM-40343: Update developer documentation

rra:
Go through the about and developer documentation and try to update them to the current state of Phalanx. Flesh out hte developer documentation and try to provide a clearer flow between the steps needed to add a new application.

Start documenting the new secrets management approach, but add a lot of warnings that it's not currently live. Rework the developer documentation about secrets into a single page about defining secrets that discusses secrets.yaml as well as the current generate_secrets.py approach and adding a secret to 1Password.

Remove the chart overview page and include its contents where appropriate in other pages. Also move documentation about creating a chart from the add an application page to the one page about writing charts.

Add a new empty Helm chart starter and recommend it for new applications that aren't web services.


docs/developers/add-external-chart.rst
```diff
@@ -32,38 +32,47 @@ No published Helm chart
...
+If you are deploying a new service (and not, for instance, adding an external chart to an existing Phalanx application), start by creating a new chart.
+For charts designed primarily to deploy an external chart, normally you want to use the empty Helm starter:
 
-Configuration mostly involves carefully reading the documentation of the upstream Helm chart and building a ``values.yaml`` file that configures it appropriately.
-You may also need to add additional resources not created by the upstream Helm chart, particularly ``VaultSecret`` objects to create any secrets that it needs.
-See :doc:`add-a-onepassword-secret` for more about secrets.
+.. code-block:: shell
+
+   $ cd applications
+   $ helm create -p $(pwd)/../starters/empty <application>
```
jonathansick, line 44: If you want to include prompt characters to imply an interactive shell, use the `prompt` directive instead so that the prompt characters aren't included when the user copies the code cell.

```
.. prompt:: bash

   cd applications
   helm create -p $(pwd)/../starters/empty <application>
```

docs/developers/define-secrets.rst
```diff
@@ -0,0 +1,188 @@
...
+##############################
+
+To allow Phalanx to manage and verify the secrets for an application, every secret that the application uses must be configured.
+This configuration is used to generate secrets that can be randomly generated, copy secret values from an external source if appropriate, and ensure that all secrets required for the application are present in Vault.
+
+If the application doesn't have any secrets, you can skip this step and continue to :doc:`add-application`.
+
+If you have not already written a Helm chart for your application, do that first, following the instructions in either :doc:`write-a-helm-chart` or :doc:`add-external-chart`.
+
+Create the secret definitions
```
jonathansick, line 12: Although the schema reference is great, it'd still be good to have some examples for different scenarios that someone can look at and copy.

docs/developers/secrets-spec.rst
```diff
@@ -0,0 +1,68 @@
...
+
+    The ``>`` and ``|`` features of YAML may be helpful in keeping this description readable inside the YAML file.
+
+.. _reStructuredText: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html
+
+``if`` (string, optional)
+    If present, specifies the conditions under which this secret is required.
+    The value should be a Helm values key that, if set to a true value (including a non-empty list or object), indicates that this secret is required.
+    The Phalanx tools will look first in :file:`values-{environment}.yaml` and then in :file:`values.yaml` to see if this value is set.
+    If this condition evaluates to false, the secret is not used in that environment.
```
jonathansick, line 23: It might be good to explicitly list what values are considered falsey.

docs/developers/write-a-helm-chart.rst
```diff
@@ -0,0 +1,242 @@
...
+   **Be sure you are using Helm v3.**
+   Helm v2 is not supported
+
+If you are creating a Helm chart for a new web service, you can use the Helm chart starter that comes with Phalanx.
+Run:
+
+.. code-block:: shell
+
+   $ cd applications
+   $ helm create -p $(pwd)/../starters/web-service <application>
```
jonathansick, line 30: See other comment about using the `prompt` directive.

docs/developers/write-a-helm-chart.rst
```diff
@@ -0,0 +1,242 @@
...
+
+Replace ``<application>`` with the name of your new application, which will double as the name of the Helm chart.
+
+For any other type of chart, start with the empty Helm chart starter.
+You will then have to add all necessary resources.
+
+.. code-block:: shell
+
+   $ cd applications
+   $ helm create -p $(pwd)/../starters/empty <application>
```
jonathansick, line 40: And switch to `prompt` directive here too.

jonathansick (approved):
This is great, thanks. A nice overhaul of the docs in general.

For the secrets docs, I think the one thing missing is some examples and maybe a bit more demonstration of the `if` conditionals. Essentially, the reference is great, but we also need some "how-to" for the secrets with examples that show some common scenarios.

docs/developers/define-secrets.rst
rra: Examples are a great idea, thanks! Now added.

PR #2443, opened 2023-08-24
[DM-40477] Shared tap chart

cbanek:


.github/workflows/ci.yaml
```diff
@@ -146,4 +146,4 @@ jobs:
           argocd app wait -l "argocd.argoproj.io/instance=science-platform" \
             --port-forward \
             --port-forward-namespace argocd \
-            --timeout 300
+            --timeout 600
```
rra, line 149: My experience is that if it doesn't come up after five minutes, it's never going to come up.

applications/livetap/Chart.yaml
```diff
@@ -1,8 +1,13 @@
 apiVersion: v2
-name: cadc-tap
+name: live-tap
```
rra, line 2: This should probably be `livetap`, I think, to match the application name. I think elsewhere we've made the chart names match the application names.

applications/tap/Chart.yaml
```diff
@@ -1,8 +1,12 @@
 apiVersion: v2
-name: cadc-tap
+name: qserv-tap
```
rra, line 2: Here too I think we should make this match the application name.

rra (commented):
This looks great. This is a really nice simplification. Thank you!

I'm fairly sure that the reason why tests are failing is that the default resource requests are much too large for a GitHub Actions runner and in the restructuring the minikube instance of tap ended up with the production resource requests. If I'm right, it should be fixable by overriding the resource requests and the JVM size for minikube to set them back to where they were before the restructuring. It looks like that was no resource requests at all and a jvmMaxHeapSize of 4G.

rra (approved):

.github/workflows/ci.yaml
cbanek: Yeah, that was my thinking too until it turned right up on my laptop.  I'm going to try a few other things but this one will be reverted.  Just wanted to make sure it wasn't that

cbanek:
Yeah that was going to be my next try, just set all the resources to empty dicts and see if it works.  I think the jvmHeapSize should be fine because you can set that above what your memory is and it usually doesn't cause problems if I remember correctly

cbanek:
Well I've tried a bunch of ways to override it, but this seems to be a known and persistent problem in helm that the merging is always kind of busted for these cases.  Here's a bug I found that says it's still going on 

https://github.com/helm/helm/issues/5184#issuecomment-1434960073

"Still exists in 3.11.1
Reproducible only with subcharts"

Lucky.

For the record I've tried resources: {} and resources: null.

cbanek:
Also for anyone interested in the helm stuff, yes it's a subchart with resources and I'm running:

version.BuildInfo{Version:"v3.12.2", GitCommit:"1e210a2c8cc5117d1055bfaa5d40f51bbc2e345e", GitTreeState:"clean", GoVersion:"go1.20.5"}

rra:
Yeah, unfortunately I think you may have to explicitly override the requests and set them to something lower. I wonder if 0 does what you want. This was probably why we had the resources in the per-environment values files before, although I agree with you that it involves less repetition to have them in one place.

cbanek:
0 is what I'm trying now

cbanek:
looks like https://github.com/helm/helm/pull/12162 might fix this when it gets rolled out

PR #2444, opened 2023-08-24
DM-40343: Update administrator documentation

rra:
Break up the bootstrapping documentation into four pages. Flesh out the prerequisites for running Phalanx, and move the discussion of hostnames into its own document. Add a new document describing how to set up secrets for an environment, and document the new Phalanx command-line tools.

Add new procedures for syncing and auditing secrets for an environment.

Add more detailed documentation of how to set up a venv to use Phalanx tools to the about pages, and merge the instructions for contributing to the documentation with the instructions for setting up a local development environment so that we're not repeating ourselves.

Standardize on calling the Argo CD application `argocd`, not `argo-cd`.


jonathansick (approved):
Awesome!

PR #2447, opened 2023-08-26
DM-40502: Add support for creating new applications

rra:
Add a new phalanx application create command that creates a new application from a Helm starter template and adds the appropriate Argo CD configuration, environment values setting, and documentation stubs.

Simplify the starters to remove the comment before appVersion, since this won't be preserved by the template post-processing without using ruamel.yaml and probably isn't needed, and remove the appVersion from the empty starter since it often doesn't have an underlying application. Also remove sources from the empty starter, since an empty starter often has no sources.


src/phalanx/cli.py
```diff
@@ -133,6 +135,44 @@ def help(ctx: click.Context, topic: str | None, subtopic: str | None) -> None:
...
 
 
+@main.group()
+def application() -> None:
+    """Commands for Phalanx application configuration."""
+
+
+@application.command("create")
+@click.argument("name")
+@click.argument("starter")
```
jonathansick, line 145: Could the CLI document what starters are available?

I also wonder if the starter should be an option with `web-service` as the default? The nice thing about it being an option is that it could be a [`choices`](https://click.palletsprojects.com/en/8.1.x/options/#choice-options) type. To implement it we'd need to have a module-scoped service that looks up the directory names in the `starters` directory, so it isn't the cleanest, but it might be a better user experience.

docs/developers/add-application.rst
```diff
@@ -29,74 +30,18 @@ See :doc:`/applications/cert-manager/add-new-hostname` for more details.
...
-
-   The template values here come from the :file:`environments/values-{environment}.yaml` configuration file for each environment.
-   You should not need to change those values when adding a new application.
-
-#. Edit `environments/values.yaml <https://github.com/lsst-sqre/phalanx/blob/main/environments/values.yaml>`__ and add your new application to the ``applications`` key with an appropriate comment, similar to the other applications already listed.
-   Please maintain the alphabetical order of the applications.
-   Except in very unusual circumstances, the application should default to ``false`` (not installed).
-
-#. Finally, enable your application in one of the :file:`values-{enviornment}.yaml` files in `environments <https://github.com/lsst-sqre/phalanx/tree/main/environments/>`__.
+#. Enable your application in one of the :file:`values-{enviornment}.yaml` files in `environments <https://github.com/lsst-sqre/phalanx/tree/main/environments/>`__.
```
jonathansick, line 44:
```suggestion
#. Enable your application in one of the :file:`values-{environment}.yaml` files in `environments <https://github.com/lsst-sqre/phalanx/tree/main/environments/>`__.
```

jonathansick (approved):

src/phalanx/data/application-docs.tmpl
```diff
@@ -0,0 +1,16 @@
+.. px-app:: {{ name }}
+
+######################################
+{{ name }} - {{ description }}
+######################################
```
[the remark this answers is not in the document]
rra: Oh, this is great, thank you! I didn't even think about using that technique but the mismatched length was bothering me as well.

src/phalanx/cli.py
rra: So, the reason why this is tricky is that the location of the configuration can be specified with a command-line flag, so it's not known at the module level, which means I can't determine the available starters at the module level. I'm not sure there's any way around that while still supporting the `-c` flag. That's why I got stuck on runtime errors.

That said, I agree that it's a real downside to the interface, so I think what I'm going to do instead is just hard-code the available starters in an enum and then add a test so that tests fail if someone adds a new starter and forgets to add it to the enum. It's not as good as dynamic discovery, but it gets us better help output.

PR #2452, opened 2023-08-28
DM-40537: Support disabling Gafaelfawr

rra:
When generating the documentation, support environments that have Gafaelfawr disabled. We may have Phalanx environments that only run backend processes without any user interface, in which case Gafaelfawr is unnecessary.


athornton (approved):
This looks good.  I'm slightly dubious that this leaves anything coherent enough to be a Phalanx deployment--it feels like we're just agreeing to 'well, it's some stuff managed by argocd'--but if it meets a need...well, this is clearly also the model for rapid-analysis, isn't it?

PR #2457, opened 2023-08-29
DM-40502: Add onepassword-connect-dev application

rra:
This application will create the 1Password Connect API servers for all development and test 1Password vaults for SQuaRE-run Phalanx environments.

Because we need to create a separate subchart per 1Password vault and dependencies with aliases can't be templated using values.yaml, unfortunately we'll need to create two Phalanx applications, one that collects the API servers for the development and test vaults and runs on roundtable-dev, and one that collects the API servers for the production vaults and runs on roundtable. This is the one for development and test servers, starting with just IDF dev.

### DIFF
--- a/applications/onepassword-connect-dev/.helmignore
+++ b/applications/onepassword-connect-dev/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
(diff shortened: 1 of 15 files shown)


rra:
This doesn't work yet because I need permissions from Cristian to create the integration, but otherwise I think all the pieces are there. I will test on roundtable-dev before merging once that's available.

jonathansick (approved):

PR #2460, opened 2023-08-31
DM-40501: Add 1Password support for static secrets

rra:
Add new configuration options for environments that allow specifying a 1Password Connect server and Vault title. If those options are set for an environment, use that, via onepasswordconnectsdk, to retrieve static secrets provided that a separate static secrets YAML file was not given.

The layout used for 1Password is to have one 1Password item per application, with one field (without section nesting) for each secret.

### DIFF
--- a/docs/developers/define-secrets.rst
+++ b/docs/developers/define-secrets.rst
@@ -158,8 +158,8 @@ See the `vault-secrets-operator documentation <https://github.com/ricoberger/vau
 
 .. _dev-add-onepassword:
 
-Create static secrets in 1Password
-==================================
+Create static secrets in 1Password (current)
+============================================
 
 .. note::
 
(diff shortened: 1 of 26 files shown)


docs/developers/define-secrets.rst
```diff
@@ -261,6 +261,60 @@ To sync multiple environments at once:
...
+
+.. note::
+
+   This document only covers creating a 1Password-backed secret for the first time for an application.
+   If you want to update a secret, either by adding new 1Password secrets or by changing their secret values, you should follow the instructions in :doc:`/developers/update-a-onepassword-secret`.
+
+.. warning::
+
+   This is the new process for storing secrets in 1Password that is currently under development.
+   Most environments should continue to use the instrutions in :ref:`dev-add-onepassword`.
```
jonathansick, line 285:
```suggestion
   Most environments should continue to use the instructions in :ref:`dev-add-onepassword`.
```

jonathansick: More generally, I think this `.. warning::` should probably be rolled up with the `.. note::` at the beginning of the section. We don't want to overwhelm with extra admonitions.

jonathansick (approved):

PR #2468, opened 2023-09-05
prompt processing argocd users addition,  fix values and url

dspeck1:
Fix the mistake in the values file format in argocd applications, update url to fix link check, and add additional users to argocd.


rra (approved):
Looks good to me. Could you also create a values file for vault-secrets-operator? I suspect it needs to look like the ones for the other USDF environments.

dspeck1:
Thanks for reviewing!  USDF started installing the vault secrets operator in the vcluster setup so that is why left it out.

PR #2474, opened 2023-09-05
USDF Alert Stream Broker

dspeck1:
Add in alert stream broker at USDF and helm chart changes for install.


environments/templates/alert-stream-broker-application.yaml
```diff
@@ -18,6 +18,13 @@ spec:
   destination:
     namespace: alert-stream-broker
     server: https://kubernetes.default.svc
+  ignoreDifferences:
+  #- jsonPointers:
+    #- /spec/volumeName
+    #- /spec/storageClassName
```
bsmartradio, line 24: Is this being left in for future use?

bsmartradio (approved):
I looked at everything and it looks great. Just saw one commented out section that I was curious if it needed to stay in or not.

environments/templates/alert-stream-broker-application.yaml
dspeck1: Thanks!  Yes.  Will do separate PR to fix out of sync argocd issue.

PR #2478, opened 2023-09-05
DM-40501: Support encoding secrets in 1Password

rra:
Some of our secrets contain significant newlines, such as the AWS credentials and pgpass file for Butler. 1Password unfortunately does not allow embedded newlines in fields. The official solution to this problem is to use files, but files can no longer be attached to Server items and instead need to be separate Document items, which breaks our entire 1Password layout.

Instead, support encoding the secret in base64 when storing it in 1Password and decoding it when retrieving it. This should allow arbitrary values at the cost of some complexity and having to explicitly tag which secrets are encoded in base64.

### DIFF
--- a/docs/developers/define-secrets.rst
+++ b/docs/developers/define-secrets.rst
@@ -323,6 +323,13 @@ Phalanx requires all of the secret entries be top-level fields outside of any se
 
 Newlines will be converted to spaces when pasting the secret value.
 If newlines need to be preserved, be sure to mark the secret with ``onepassword.encoded`` set to ``true`` in :file:`secrets.yaml`, and then encode the secret in base64 before pasting it into 1Password.
+To encode the secret, save it to a file with the correct newlines, and then use a command such as:
+
+.. prompt:: bash
+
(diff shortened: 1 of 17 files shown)


docs/developers/define-secrets.rst
```diff
@@ -305,6 +321,9 @@ Change the field type to password so that the value isn't displayed any time som
 Do not use sections.
 Phalanx requires all of the secret entries be top-level fields outside of any section.
 
+Newlines will be converted to spaces when pasting the secret value.
+If newlines need to be preserved, be sure to mark the secret with ``onepassword.encoded`` set to ``true`` in :file:`secrets.yaml`, and then encode the secret in base64 before pasting it into 1Password.
```
jonathansick, line 325: Maybe include a example command that shows piping a file through base64 to encode a file?

jonathansick (approved):
Ah interesting. I used secure notes for this purpose when including things like private keys for GitHub apps and I forgot it'll be an issue when we switch to multiple values in the same 1Password item.

PR #2480, opened 2023-09-06
Change Strimzi loglevel to INFO

afausti:
- Strimzi loglevel was set to DEBUG in some environments, but logs are too chatty.


mareuter (approved):
Looks good to me.

PR #2481, opened 2023-09-06
[DM-40662] Fix Kapacitor alerts crosstalk at base

afausti:
Deploy a second instance of Kapacitor and point it to its corresponding InfluxDB instance.
Enable it at base.


mareuter (approved):
Looks good to me.

PR #2486, opened 2023-09-07
prompt processing next visit fan out app

dspeck1:
Add application and enable in dev


applications/next-visit-fan-out/values-usdfdev-prompt-processing.yaml
```diff
@@ -0,0 +1,23 @@
...
+  consumerGroup: test-group-3
+  nextVisitTopic: test.next-visit
+
+image:
+  repository: ghcr.io/lsst-dm/next_visit_fan_out
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: v1.0.1
+
+vaultSecretsPath: "secret/rubin/usdf-prompt-processing-dev/next-visit-fan-out"
```
rra, line 23: This should not be set here; instead, it should be injected by Argo CD as a global (see the other applications).

applications/next-visit-fan-out/values.yaml
rra, line 1: Needs Helm docs comments for each setting. Also needs the `globals` block to document the things passed in by Argo CD (see below).

environments/templates/next-visit-fan-out-application.yaml
```diff
@@ -0,0 +1,28 @@
...
+  namespace: argocd
+  finalizers:
+    - resources-finalizer.argocd.argoproj.io
+spec:
+  syncPolicy:
+    managedNamespaceMetadata:
+      labels:
+        argocd.argoproj.io/managed-by: argocd
+    syncOptions:
+    - CreateNamespace=true
```
rra, line 15: This is interesting and we probably should think about switching over to it, but to keep things consistent for now, could you explicitly manage the namespace as a resource the way that the other applications do?

environments/values-usdfdev-prompt-processing.yaml
rra, line 1: Default also needs to be added to `values.yaml`.

rra (commented):
Almost all of these comments would I think be handled for you automatically if you created the new application with `phalanx application create next-visit-fan-out --starter empty` to use the empty application starter. See the new documentation at https://phalanx.lsst.io/developers/write-a-helm-chart.html which I think covers everything I mentioned here.

applications/next-visit-fan-out/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
```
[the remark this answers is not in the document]
dspeck1: Phalanx rendered chart version sets it to 0.1.0

```
apiVersion: v2
appVersion: 0.1.0
description: Application to poll next visit events from kafka, fan out/duplicate the
  message, and send to knative.
name: next-visit-fan-out
type: application
version: 0.1.0
```

PR #2496, opened 2023-09-08
Add nublado major version upgrade doc

athornton:


docs/applications/nublado/major-upgrade.rst
```diff
@@ -0,0 +1,313 @@
...
+
+Note that in future the secret setup and management will all be handled
+as part of Phalanx deployment, and you will no longer need to do this
+manually.  However, at the present, you still do.
+
+Enable the application
+""""""""""""""""""""""
+
+In the values file for your installation in the top-level Phalanx
+``environments`` directory, set ``nublado: true``.
```
cbanek, line 81: This will probably always and already be set? Are you going to copy over the whole /applications/nublado around in phalanx?

This seems like something else that the user doesn't see, and moving these directories around seems like it could be really hard to back out a change if there's a problem?

docs/applications/nublado/major-upgrade.rst
```diff
@@ -0,0 +1,313 @@
...
+Note that if you do share persistent storage between nublado versions,
+the step about communicating with your users (below) goes from
+"important" to "critical".
+
+
+Create an alternate route
+"""""""""""""""""""""""""
+
+In the values file for your installation in ``applications/nublado``,
+set the key ``jupyterhub.hub.baseUrl`` to ``/n3``.
```
cbanek, line 101: yeah this kind of feels like it should be applications/nublado3 (which I assume already exists so you can test it on the environment?) /n3 seems like the right thing and the thing you would change to /nb when you want to roll it out to everyone?

docs/applications/nublado/major-upgrade.rst
```diff
@@ -0,0 +1,313 @@
...
+
+Let your users know about the new route and when it will be enabled
+(presumably, at the end of the next maintenance window).  Announce that
+that maintenance window will be disruptive and will destroy any existing
+user sessions.
+
+If you are using the same storage for users regardless of which spawner
+they're coming in as, which you very likely are, in addition to the new
+route, you need to do your utmost to make it extremely clear that if
+they are testing the new route, it is vital to make sure that their labs
```
cbanek, line 114: their labs from previous versions

docs/applications/nublado/major-upgrade.rst
```diff
@@ -0,0 +1,313 @@
...
+saved every five minutes by each environment, thus bafflingly reverting
+changes, and at worst it will cause full-scale corruption of the files
+the user has open.
+
+
+Silence Mobu complaints
+"""""""""""""""""""""""
+
+Turn off ``mobu`` if you're using it, so that it doesn't complain
+about the service outage.
```
cbanek, line 126: turn off the notebook runner of mobu, we still want TAP and other things to run

docs/applications/nublado/major-upgrade.rst
```diff
@@ -0,0 +1,313 @@
...
+
+Disable logins
+""""""""""""""
+
+Disable user logins to nublado v2.  This is most easily done by deleting
+the ``hub`` deployment from nublado v2.
```
cbanek, line 132: you could also just delete the ingresses to the hub

docs/applications/nublado/major-upgrade.rst
```diff
@@ -0,0 +1,313 @@
...
+
+Remove user sessions
+""""""""""""""""""""
+
+Remove any user sessions.  This is not, strictly speaking, necessary,
```
cbanek, line 137: I'd just say delete their sessions!  It seems like with the shared homedirs that it kind of is necessary, and either way it's probably safer to delete their session?  You already said above to tell them to turn it off

docs/applications/nublado/major-upgrade.rst
```diff
@@ -0,0 +1,313 @@
...
+#. Synchronize the ``science-platform`` application.  This will enable
+   the ``nublado`` application, which contains nublado v3.
+#. Synchronize the ``nublado`` application.  This will turn on the
+   nublado v3 spawner.
+#. Test that you can get a lab and that that lab allows use of the
+   command line and cell executions with the new spawner at ``/n3``.  If
+   you find errors, troubleshoot them.  Only proceed once you have the
+   new spawner correctly spawning and shutting down labs.
+#. Synchronize the ``nublado2`` application.  This will reenable the
+   default spawner.  At this point users will start returning (whether
```
cbanek, line 156: which one is the default?

docs/applications/nublado/major-upgrade.rst
cbanek: Or maybe you're saying delete it in argo.  At least that doesn't have to take a commit.  Whichever way you're saying to turn it off it'd be good to be specific

docs/applications/nublado/major-upgrade.rst
```diff
@@ -0,0 +1,313 @@
...
+disruptive and will destroy any existing user sessions.
+
+If you are using the same storage for users regardless of which spawner
+they're coming in as, which you very likely are, in addition to the new
+route, you need to do your utmost to make it extremely clear that if
+they intend to use both routes, it is vital to make sure that their labs
+are shut down before changing from one route to the other.  If they do
+not do this, it will at best create confusion as open notebooks will be
+saved every five minutes by each environment, thus bafflingly reverting
+changes, and at worst it will cause full-scale corruption of the files
```
cbanek, line 186: is it really necessary to tell the users they can use the old version?  seems like you really want to herd them and make sure they contact you first to report issues before saying they can try /n2.

cbanek: I'd really say it's better to really prevent the filesystem corruption issues considering we've hit that before

docs/applications/nublado/major-upgrade.rst
```diff
@@ -0,0 +1,313 @@
...
+In the values file for your installation in ``applications/nublado2``,
+set the key ``jupyterhub.hub.baseUrl`` to ``/n2``.
+
+Change nublado3 configuration
+"""""""""""""""""""""""""""""
+
+Remove the key ``jupyterhub.hub.baseUrl`` from the values file for your
+installation in ``applications/nublado``.
+
+This and the preceding step will make nublado v3 your default spawner on
```
cbanek, line 210: this paragraph being here seems kind of confusing to me.  why not just say in the section above "this will make nublado2 at /n2"

cbanek: Then you can say instead of "This and the preceding step will make nublado v3 your default spawner on" "removing the route to /n3" will make nubladov3 run at /nb

docs/applications/nublado/major-upgrade.rst
```diff
@@ -0,0 +1,313 @@
...
+the user has open.
+
+Change mobu configuration
+"""""""""""""""""""""""""
+
+If you are using ``mobu``, change the instance-specific values file.  In
+``config.autostart``, for each entry, set ``business.use_cachemachine`` to
+``false``.  This is necessary because nublado 3 uses its Jupyterlab
+Controller to carry out the functions formerly performed by
+``cachemachine``.
```
cbanek, line 196: and turn mobu back on and sync it, or however you're turning it on (I think you're implying that here)

docs/applications/nublado/major-upgrade.rst
```diff
@@ -0,0 +1,313 @@
...
+During the deployment window:
+
+#. Synchronize the ``science-platform`` application.  This will delete
+   the ``nublado2``, ``cachemachine``, and ``moneypenny`` applications.
+#. Remove any user sessions spawned by ``nublado2``.  This is harder to
+   do from the UI: you will need to find any users in ``nublado-users``
+   with a namespace called ``nublado2-<username>`` and remove only those
+   namespaces.  From the CLI, it's easier: ``kubectl get ns | grep
+   nublado2- | awk '{print $1}' | xargs kubectl delete ns``.
+#. Delete the values files for your installation from Phalanx's
```
cbanek, line 293: for your environment

cbanek: I actually wouldn't do any of this deleting until you know that the new version is good, maybe break that into some kind of cleanup section - the instructions make sense

docs/applications/nublado/major-upgrade.rst
```diff
@@ -0,0 +1,313 @@
...
+   DBA to do so), and remove the ``nublado2``, ``cachemachine``, and
+   ``moneypenny`` entries from your instance's Vault envlave.  In
+   practice we have not been doing this, because the resources consumed
+   are minimal.
+
+When Migration Is Complete For All Sites
+-----------------------------------------
+
+Once all sites are running nublado v3, and none are still running
+nublado v2, ``applications/nublado``, ``applications/moneypenny``, and
```
cbanek, line 307: applications/nublado2?

docs/applications/nublado/major-upgrade.rst
cbanek: yeah I'd put the deleting things into the everyone is running it section.  You never know when usdf is going to find some problem.

cbanek (approved):
The important thing is that it makes sense to you as this is a one time set of instructions.

I think my big point would be I'd rather have nublado v3 in a nublado3 directory and follow that pattern rather than trying to do a cup and balls routine to make nublado3 into nublado.

docs/applications/nublado/major-upgrade.rst
athornton: Yes, and you're right that we should have called this `/nublado3` but we didn't, and we're almost done with the migration.

docs/applications/nublado/major-upgrade.rst
```diff
@@ -0,0 +1,313 @@
+###############################
+Nublado Major Version Migration
+###############################
+
+Currently we only discuss nublado v2 to nublado v3 migration; when the
+next major version comes about, we'll add those instructions as well.
+
+In general, when a new version appears, it will be called ``nublado``
+and the previous version will be ``nubladoN`` where N is currently 2,
```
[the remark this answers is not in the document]
athornton: So the way it's been done so far (which is...everywhere we control except summit) is:

Bring up `nublado` (which probably should have been called `nublado3`) on `/n3`.  Encourage people to test it.

Once that seems OK, move `nublado2` to `/n2`.  Make `/nublado` `/nb`.  Now people *will* be using it but the first phase should have gotten the `it just plain doesn't work` pieces out of the way and bugs found will be more subtle.

Once that's been running a while without user complaints, turn off `/n2` and cachemachine/moneypenny.

I'll try to make this less confusing.

docs/applications/nublado/major-upgrade.rst
```diff
@@ -0,0 +1,313 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The following tasks can be performed at any time, and should be
+performed ahead of your maintenance window to actually turn on the
+application.
+
+Vault Secret Creation
+"""""""""""""""""""""
+
+The first thing you will need to do is to set up secrets for
```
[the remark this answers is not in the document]
athornton: It would have, but we're almost done with the migration at this point.  If we install the next one under `/nublado4` then we'll fix the docs then.

docs/applications/nublado/major-upgrade.rst
athornton: There are going to be Phalanx environments without nublado.  Roundtable, for instance.  Encoding the major version in the app name for future releases is probably a good idea, although I suspect we had some theory for why we shouldn't this time that I have forgotten.

docs/applications/nublado/major-upgrade.rst
athornton: See above.  As Cletus put it, "Shoulda, but didn't."

docs/applications/nublado/major-upgrade.rst
athornton: OK.  I meant just delete it in argo and then resync at the end of the deployment window.  I will make that explicit.

docs/applications/nublado/major-upgrade.rst
athornton: I'll make it explicit that it's a resync.

athornton:
Well, after discussion, it appears that Frossie may have been the catalyst, and is certainly a proponent of, the base app being called `nublado` in which case I think this is ready to merge.

PR #2508, opened 2023-09-12
[DM-40741] Enable MM2 auto-restart feature

afausti:
- When enabled, the Strimzi Cluster Operator watches the connector and its tasks for failures and if in case of failures it automatically restarts them.


mareuter (approved):
Looks good to me.

PR #2510, opened 2023-09-12
DM-37387: Update LATISS dev configs in prompt processing

hsinfang:


applications/prompt-proto-service-latiss/values-usdfdev-prompt-processing.yaml
```diff
@@ -8,10 +8,11 @@ prompt-proto-service:
     repository: ghcr.io/lsst-dm/prompt-proto-service
     pullPolicy: Always
     # Overrides the image tag whose default is the chart appVersion.
-    tag: AuxTel20230524-w_2023_20
+    tag: DM-41084-temporary-latest
 
   instrument:
-    calibRepo: s3://rubin-summit-users/
+    pipelines: (survey="SURVEY")=[${PROMPT_PROTOTYPE_DIR}/pipelines/${RUBIN_INSTRUMENT}/ApPipe.yaml]
+    calibRepo: s3://rubin-pp-users/central_repo/
```
kfindeisen, line 15: I wonder if this is why we could read Gaia DR3 on `-dev`, given that `-prod` uses `/app/butler`...

kfindeisen: Would it be possible to set the `calibRepo` in `values.yaml` to `""`, since this is something we always want to be different between `-dev` and `-prod`?

applications/prompt-proto-service-latiss/values-usdfdev-prompt-processing.yaml
```diff
@@ -2,7 +2,7 @@ prompt-proto-service:
 
   podAnnotations:
     # Update this field if using latest or static image tag in dev
-    revision: "1"
+    revision: "30"
```
kfindeisen: I think this change could have been done purely in Argo, once the changes in "Update LATISS dev configs" were pushed.

kfindeisen (approved):
Looks good!

applications/prompt-proto-service-latiss/values-usdfdev-prompt-processing.yaml
hsinfang: I don't think using the pod's own `/app/butler/butler.yaml` or directly using `/repo/embargo` would make a difference in retrieving a Gaia DR3 file existing on the disk that is not accessible from the pod.  The `butelr.yaml` config are the same. The registry records had paths like `file:///sdf/data/rubin/shared/the_monster/GAIA_DR3`  and retrieving unaccessible files from inside the pod would just never work. 

Anyhow, [DM-41171](https://jira.lsstcorp.org/browse/DM-41171) will disable using the workaround in `prod`.

PR #2512, opened 2023-09-12
[DM-40761] Add 2MASS color to the list of HiPS sets

cbanek:
This will tell the portal to put it in the list of selectable HiPS sets.  It should alread have its properties file as part of the list file.

### DIFF
--- a/applications/portal/templates/deployment.yaml
+++ b/applications/portal/templates/deployment.yaml
@@ -106,7 +106,8 @@ spec:
                               "temp://lsst/dp02_dc2/hips/images/band_r",
                               "temp://lsst/dp02_dc2/hips/images/band_i",
                               "temp://lsst/dp02_dc2/hips/images/band_z",
-                              "temp://lsst/dp02_dc2/hips/images/band_y"
+                              "temp://lsst/dp02_dc2/hips/images/band_y",
+                              "ivo://CDS/P/2MASS/color"
                           ],
                           "label": "Rubin Featured MOC"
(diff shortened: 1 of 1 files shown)


cbanek:
I'm still syncing the tiles, and then I'm going to have to restart datalinker and pull in a new version for the list file to be correct and contain the properties file.

applications/portal/templates/deployment.yaml
```diff
@@ -106,7 +106,8 @@ spec:
                               "temp://lsst/dp02_dc2/hips/images/band_r",
                               "temp://lsst/dp02_dc2/hips/images/band_i",
                               "temp://lsst/dp02_dc2/hips/images/band_z",
-                              "temp://lsst/dp02_dc2/hips/images/band_y"
+                              "temp://lsst/dp02_dc2/hips/images/band_y",
+                              "ivo://CDS/P/2MASS/color"
```
robyww, line 110: Have you confirm that `ivo` will resolve to the rubin server? I don't quite remember how to how the resolve the `ivo` in this context. I can look into i,t but if you have confirmed that it will indeed hit the rubin server then it looks good.

robyww: OK, I checked in to it. It will resolve to the rubin service if the rubin hips list service shows that it serves `/CDS/P/2MASS/color`. 

So this line is fine but you might want to confirm this with the hips list service.

robyww (approved):

PR #2513, opened 2023-09-13
Extend ts-salkafka user access control to groups

afausti:
In salobj, the ts-salkafka user also needs authorization to perform operations on kafka groups.


tribeiro (approved):
ok

PR #2514, opened 2023-09-13
DM-40697: Add pull-secret handling

rra:
Restructure the static secrets format and template to move the application secrets under an applications key and add a new pull-secret top-level key with details about the pull secret. This is in the form of a dict of registries to username and password pairs.  The encoded .dockerconfigjson secret key is generated during secret sync.

Add 1Password support for retrieving the pull secret. This 1Password item uses one section per registry with fields username and password specifying the HTTP Basic Auth credentials.

Change the phalanx secrets onepassword-secrets command to write a YAML file in the same syntax as is used for static secrets. Not only is this easier to compare, it makes it possible to download secrets once from 1Password and then perform multiple audit or sync operations without the 1Password credentials, if desired.


rra:
Tested by running `phalanx secrets onepassword-secrets idfdev` and inspecting the resulting output.

jonathansick (approved):
Looks great, thanks. It's good that we found this now because in retrospect I agree it's useful to namespace the application secrets so that other things can be added to the top-level of these secret configuration YAML files in the future if needed.

PR #2516, opened 2023-09-14
prompt processing: add production environment

dspeck1:
Add production environment and enable prompt proto service for latiss

### DIFF
--- a/charts/prompt-proto-service/values.yaml
+++ b/charts/prompt-proto-service/values.yaml
@@ -42,6 +42,9 @@ registry:
   ip: ""
   db: ""
   user: ""
+  centralRepoFile: false
+
+logLevel: lsst.resources=DEBUG
 
 knative:
   ephemeralStorageRequest: "20Gi"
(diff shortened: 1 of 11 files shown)


charts/prompt-proto-service/values.yaml
```diff
@@ -42,6 +42,9 @@ registry:
   ip: ""
   db: ""
   user: ""
+  centralRepoFile: false
+
+logLevel: lsst.resources=DEBUG
```
kfindeisen, line 47: In my experience, `lsst.resources` logs are _extremely_ obnoxious, and turning them on by default (for both `-dev` and `-prod`, and all instruments?) will make it impossible to discern anything we might be interested in at the `DEBUG` level (including our own diagnostics).

I see that this was copied from https://github.com/slaclab/rubin-usdf-prompt-processing/blob/main/kubernetes/overlays/prod/prompt-proto-service/prompt-proto-service.yaml#L102, but would it be possible to narrow the scope a bit?

dspeck1: What default would you like?

kfindeisen: If it's not possible to leave `SERVICE_LOG_LEVELS` undefined, then `""` would have the same effect.

dspeck1: Thanks.  Will set default to `""` in separate PR.

PR #2538, opened 2023-09-20
DM-36772: Migrate PostgreSQL credentials to mounted file

kfindeisen:
This PR does some refactoring and documentation cleanup of the `prompt-proto-service-*` applications, then rewrites its handling of PostreSQL credentials to mount them as a file instead of through environment variables. This change (with the corresponding code in lsst-dm/prompt_prototype#83) will greatly simplify `prompt-proto-service` configuration in the future.


dspeck1 (approved):
Looks great!  Thanks for adding comments to values files and additional chart fields.

PR #2545, opened 2023-09-21
DM-36772: Remove vault path overrides for prompt-proto-service.

kfindeisen:
This PR removes unwanted `vaultSecretsPath` overrides (see #2520) that should have been cleaned up on #2538.


dspeck1 (approved):
Looks good.  Thanks for removing.

PR #2557, opened 2023-09-22
DM-40887: Update prompt-proto-service to d_2023_09_22

hsinfang:


applications/prompt-proto-service-latiss/values-usdfprod-prompt-processing.yaml
```diff
@@ -8,7 +8,7 @@ prompt-proto-service:
     repository: ghcr.io/lsst-dm/prompt-proto-service
     pullPolicy: Always
```
kfindeisen, line 9: Pre-existing issue, but should this be `IfNotPresent` instead of `Always`? Possibly a question for @dspeck1.

kfindeisen (approved):
Change itself looks good, one question about the existing config.

applications/prompt-proto-service-latiss/values-usdfprod-prompt-processing.yaml
dspeck1: IfNotPresent is preferred for prod as long as new tags or revision annotation updated (if tag same) when a new docker image is built.

hsinfang: Thank you! This is changed to `IfNotPresent`

PR #2559, opened 2023-09-23
[DM-40888] Expose Schema Registry API in Sasquatch

afausti:
Add ingress configuration to give external access to the SR API it  can be used in the TTS. 
Authentication will be implemented later.


mareuter (approved):
Small typo but otherwise this looks fine.

PR #2570, opened 2023-09-25
alert stream broker: prometheus monitoring

dspeck1:
Add prometheus monitoring and JMX exporter in helm chart.


bsmartradio (approved):
It all looks good. Looking forward to using this!

PR #2572, opened 2023-09-25
DM-40889: Update secrets documentation for new method

rra:
The old 1Password approach isn't supported any more and some of the infrastructure has been turned off, so delete the old documentation. We're not going to follow it again; if we have to, we'll make temporary Vault changes until we can switch over.

Do an editing pass on the secrets management documentation and simplify for having only one approach. Remove a bunch of warnings that should hopefully now be irrelevant.


docs/admin/secrets-setup.rst
```diff
@@ -106,29 +108,19 @@ Static secret sources
...
-
-Currently, the installer scripts that populate Vault for a given Phalanx environment support either retrieving static secrets from 1Password_ or prompting interactively for them.
-For all SQuaRE-run Phalanx environments, the static secrets are stored in 1Password.
-
-This mechanism is being retired and therefore is not documented here in detail.
-For instructions on how to sync secrets using the installer scripts and using 1Password as a source of static secrets, see :ref:`dev-add-onepassword`, step three.
-
-Using the Phalanx command-line tool
------------------------------------
+Static secrets cannot be; the Phalanx administrator has to provide the values of static secrets in order to install a Phalanx environment.
```
jonathansick, line 111: The first sentence combined with the "Static secrets cannot be" is maybe a bit awkward. Consider defining what a static secret is, and then make the comparison to generated secrets. For folks like me who don't make much, if any, use of secrets generation, defining static secrets via comparison isn't as effective.

docs/developers/define-secrets.rst
```diff
@@ -128,25 +124,11 @@ Finally, here is an example of a static secret that needs an additional layer of
...
-Since this mechanism is currently being replaced with a new set of tooling, it is not documented here in detail.
-Adding new secrets to this mechanism will require reading the source and following the pattern used by other applications.
-
 Define VaultSecret resources
 ============================
 
-Once a secret is in Vault, you need to create or update a ``VaultSecret`` resource in your application's deployment (typically in its Helm chart).
-See :doc:`write-a-helm-chart` for more details on creating a Helm chart for an application.
+The Phalanx secrets tooling will ensure that the secret is in Vault, but you must still create or update a ``VaultSecret`` resource in your application's deployment (typically in its Helm chart) to tell `Vault Secrets Operator`_ how to create a ``Secret`` that your application can use.
+(See :doc:`write-a-helm-chart` for more details on creating a Helm chart for an application.)
```
jonathansick, line 131: Are these sorts of parenthetical sentences grammatically correct? I don't think I've seen them in technical writing I've read.

jonathansick (approved):

docs/developers/define-secrets.rst
rra: I think it's grammatically correct, in the sense that you can put whole sentences in parentheses and that's fine in English grammar, but you are entirely correct that I adore my parenthetical asides probably a little too much. :) In this case, reading this again, I don't think the parentheses add anything, and I'll drop them.

PR #2580, opened 2023-09-25
DM-40889: Improve developer documentation

rra:
Document the trick of adding a checksum annotation to force restarts of the deployment. Stop telling people to add development versions to the Chart.yaml appVersion; for development tags, just use overrides of image.tag. Tell people they can sync their application in the science-platform app to revert development changes. Clarify documentation about Chart.yaml and pull secrets.

### DIFF
--- a/docs/developers/deploy-from-a-branch.rst
+++ b/docs/developers/deploy-from-a-branch.rst
@@ -183,8 +183,7 @@ Once your branch is merged, remember to reset your application's Argo CD ``Appli
 
 #. In the application's page in Argo CD, click on the :guilabel:`Sync` button to redeploy the application from the default branch.
 
-You can also find your application in the ``science-platform`` Argo CD application and sync it from there to reset the default branch and any other settings you changed.
-Either approach works; use whichever you find more convenient.
+Alternatively, you can find the application in the ``science-platform`` Argo CD application and sync it from there to reset the default branch and any other settings you changed.
 
 Next steps
 ==========
--- a/docs/developers/deploy-from-a-branch.rst
+++ b/docs/developers/deploy-from-a-branch.rst
@@ -97,10 +97,9 @@ The fastest method for trying out changes to Kubernetes resources is to directly
 In your application's Argo CD page you can click on a specific resource (such as a ConfigMap_ or Deployment_) and click the :guilabel:`Edit` button on the live manifest.
 Make your changes, then click :guilabel:`Save`.
 
-Normally, these changes will immediately take effect.
-Sometimes if you change a ``ConfigMap`` you will need to restart the relevant deployments to pick up that change.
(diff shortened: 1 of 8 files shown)


docs/developers/deploy-from-a-branch.rst
```diff
@@ -100,6 +100,7 @@ Make your changes, then click :guilabel:`Save`.
 Normally, these changes will immediately take effect.
 Sometimes if you change a ``ConfigMap`` you will need to restart the relevant deployments to pick up that change.
 For instructions on how to do that, see :ref:`branch-deploy-restart`.
+(However, see :ref:`dev-deployment-restart` for a better way.)
```
jonathansick, line 103:
```suggestion
(However, see :ref:`dev-deployment-restart` for a better way.)
To configure the chart to make these restarts happen automatically, see :ref:`dev-deployment-restart`.
```

docs/developers/deploy-from-a-branch.rst
```diff
@@ -183,6 +184,9 @@ Once your branch is merged, remember to reset your application's Argo CD ``Appli
 
 #. In the application's page in Argo CD, click on the :guilabel:`Sync` button to redeploy the application from the default branch.
 
+You can also find your application in the ``science-platform`` Argo CD application and sync it from there to reset the default branch and any other settings you changed.
+Either approach works; use whichever you find more convenient.
```
jonathansick, line 188:
```suggestion
Alternatively, you can find the application in the ``science-platform`` Argo CD application and sync it from there to reset the default branch and any other settings you changed.
```

docs/developers/write-a-helm-chart.rst
```diff
@@ -5,7 +5,7 @@ Write a Helm chart for an application
 Argo CD manages applications in the Rubin Science Platform through a set of Helm charts.
 Which Helm charts to deploy in a given environment is controlled by the :file:`values.yaml` and :file:`values-{environment}.yaml` files in `/environments <https://github.com/lsst-sqre/phalanx/tree/main/environments/>`__.
```
jonathansick, line 6: If the intent is to standardize on not using `/` path prefixes, there may be other instances:

```suggestion
Which Helm charts to deploy in a given environment is controlled by the :file:`values.yaml` and :file:`values-{environment}.yaml` files in `environments <https://github.com/lsst-sqre/phalanx/tree/main/environments/>`__.
```

jonathansick (approved):

docs/developers/deploy-from-a-branch.rst
rra: Yes, this is better, but when applying it, I realized that my whole original change was wrong. The checksum is only computed during Helm operations, and this section is on making changes directly in Argo CD, so the restart always has to be done manually. I'll revise it a different way to make that clear.

PR #2585, opened 2023-09-27
DM-40889: Convert minikube to new secrets management

rra:
Convert the minikube environment to the new secrets management approach. Make the minimum changes to the installer to allow it to work with a Vault AppRole instead of a read token, which makes it usable with the new secrets management system.

Delete all of the old secret management scripts. Although we still have environments using the old secrets management approach, we're unlikely to use those scripts again to manage those environments, and any environment we touch from this point forward should be converted to the new system. In case of emergencies, we can poke changes directly into Vault for the time being.


jonathansick (approved):
Looks great!

PR #2593, opened 2023-09-28
update argocd sso secrets to fit new template

pav511:


rra (approved):
The other usdf environments will also need this change (alert-stream-broker, prompt-processing).

PR #2598, opened 2023-09-28
DM-40947: Better handling of missing 1Password secrets

rra:
Retrieve each 1Password item separately rather than using the bulk API so that we can do proper error reporting of all of the missing secrets we expected to be in 1Password but which weren't found. Throw that as an exception up through sync, but, in audit, catch that exception and properly report the missing secrets.

### DIFF
--- a/src/phalanx/exceptions.py
+++ b/src/phalanx/exceptions.py
@@ -116,6 +116,24 @@ class InvalidSecretConfigError(Exception):
         super().__init__(msg)
 
 
+class MissingOnepasswordSecretsError(Exception):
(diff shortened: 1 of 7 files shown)


jonathansick:
While we're thinking about the audit command, I wonder if we can clarify the labels in the audit report.

- "Missing secrets:" -> "Secrets in 1Password not yet synced to Vault:"
- "Incorrect secrets:" Not sure, but is "Secrets in 1Password and Vault differ:" the right idea? Or are there more possibilities for a secret being incorrect?

jonathansick (approved):

PR #2600, opened 2023-09-29
DM-40947: Properly diagnose 1Password encoding errors

rra:
Report a proper exception including the application and secret key names if the value of a secret in 1Password that is marked as encoded is either invalid base64 or decodes to something that cannot be represented as a Unicode string.


jonathansick:
This solves my problem, thanks. I was making a secret in secrets.yaml as being base64 encoded but ultimately it didn't need to be encoded in 1Password.

PR #2624, opened 2023-10-05
[DM-40944] Add pull secret to starter

cbanek:


rra (changes requested):
I don't think we should do this because the normal case of packages built with GitHub Actions don't need a starter. If we anticipate having multiple other applications that require a pull-secret in the future, we could add another starter and mention it in the documentation, though.

cbanek:
yeah I'm just keeping this open until I can make sure all the environments have the pull secrets that I need.  But I don't plan on merging this.

PR #2637, opened 2023-10-09
prompt processing: prompt proto service pod anti affinity and k8s requests

dspeck1:
Add pod anti affinity to prefer spreading pods on nodes.  Resource requests for CPU and memory.  Values may need to be adjusted later.


athornton (approved):
Looks good to me.

PR #2641, opened 2023-10-11
DM-40497: Overhaul how chart linting is done

rra:
We previously were using `helm lint` to lint our charts using the chart-tester GitHub Action, plus some code that attempted to expand a chart into copies of the chart for every environment so that it would detect problems with per-environment configuration. However, the latter code only handled environment configuration without a main `values.yaml`, an approach that we have since abandoned, and we weren't enabling strict mode, so this linting was mostly not doing anything. It was also incredibly slow, since it ran `helm repo update` for every chart.

Replace this framework with a new linting framework based on the Phalanx command-line tool, which understands our environment configuration and can pass in the appropriate `--values` and `--set` flags to `helm lint` to mimic what Argo CD will do. This new code adds the following:

- `phalanx application lint`, which lints a specific application chart either for all environments it is configured for or just one environment. This runs the correct sequence of `helm repo add`, `helm repo update`, `helm dependency update`, and `helm lint` to lint the whole application including third-party charts. I have confirmed that this catches YAML and Helm templating syntax errors.
- `phalanx application template`, which uses `helm template` similarly to print out all the Kubernetes resources that would be created by a given chart in a given environment.
- `phalanx application lint-all`, which lints all charts. This optimizes repo setup to only update repositories once, which makes this fast enough to allow it to be run on all application charts (previously this took tens of minutes). This also takes a `--git` flag, which uses GitPython to analyze the delta from a branch and lint only the relevant charts.
- `phalanx environment lint` and `phalanx environment template` equivalents for the top-level chart.

The Helm GitHub Action check has been replaced with a tox target (so that it can share a cache with Python testing) that runs `phalanx application lint-all --git` and `phalanx environment lint`. The latter is extremely fast so there's no reason not to run it every time. Helm installation is now handled by another GitHub Action provided by the Azure team. (This requires a GitHub token since it uses it to make a GraphQL query to GitHub to get the latest Helm release.)


src/phalanx/services/application.py
```diff
@@ -155,30 +155,42 @@ def lint(self, app_name: str, env_name: str | None) -> bool:
             success &= self._helm.lint_application(app_name, name, values)
         return success
 
-    def lint_all(self) -> bool:
+    def lint_all(self, *, git: bool = False) -> bool:
```
jonathansick: It might be a good to use a slightly more self-explanatory argument name for the Python API rather than just adopting the CLI option name (where it does make sense).

docs/about/local-environment-setup.rst
```diff
@@ -79,9 +79,7 @@ Install helm
 
 Some Phalanx commands require Helm (v3 or later) to be available on your PATH.
 Any version of Helm after v3 should be okay.
-You therefore must have it installed on your PATH.
-
-See the `Helm installation guide <https://helm.sh/docs/intro/install/>`__ for more details.
+See the `Helm installation guide <https://helm.sh/docs/intro/install/>`__ for installation instructions.
```
jonathansick, line 82: Did people really not see "Helm installation guide" as being "for installation instructions?" I guess we can repeat the word "installation" twice in a sentence to make a point... 😂

jonathansick (approved):
This looks really great! I'm excited to have this tooling when building out charts.

I gather the command for app linting is:

```
phalanx application lint <application> <environment>
```

What do you think about enabling linting on multiple applications or multiple specific environments in the same command invocation? Could we parse out a comma-separated listing of applications, like:

```
phalanx application lint times-square,noteburst --env idfdev
```

If we think that might be useful then turning the environment into a `-e/--env` type option might be a good idea.

src/phalanx/services/application.py
```diff
@@ -90,6 +123,164 @@ def create_application(
...
+            Environment whose globals should be injected.
+
+        Returns
+        -------
+        dict of str
+            Dictionary of Helm settings to their (string) values.
+
+        Notes
+        -----
+        This is a bit of a hack, since it hard-codes the injected values
```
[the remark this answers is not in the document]
rra: I do think we may eventually want to do something smarter. Right now, this is all going to be fairly opaque with no obvious error message, but doing better even on error reporting requires parsing the application templates, which feels like a big enough work item that I'm going to save it for another story. I added it to my long-term to-do list, though.

rra:
Good catch on the API. I have made the change you suggest to move the environment restriction to a flag and allow multiple environments to be given. I didn't make any corresponding change to template, since there it only makes sense to template a specific application and environment pair. (You didn't mention this, but since I made it inconsistent, it seemed worth calling out.)

The new code supports space-separated application names as separate arguments, which avoids having to introduce additional syntax for comma-separated lists.

PR #2650, opened 2023-10-12
temp fix: increase gafaelfawr redis memory limit

pav511:


athornton (approved):
It's your system's memory, but I wouldn't pin requests to 200Mi, just the limit.

pav511:
yeah good point, hit the merge queue...will repush branch

PR #2653, opened 2023-10-13
DM-41186: Add CADC auth support

rra:
Update to Gafaelfawr 9.5.0 and add appropriate configuration to enable the CADC-specific user information endpoint on IDF and USDF environments running TAP servers.


cbanek (approved):

cbanek:
Sorry I thought this already went in - I'm going to test on usdf-int next, and this worked for data-dev so I think we can just push this everywhere?

PR #2661, opened 2023-10-16
tickets/DM-41196: Add default ArgoCD application for nublado user objects

athornton:


rra (approved):
Thanks! Yeah, I forgot to update the chart after I made the last release.

PR #2664, opened 2023-10-18
Disable Summit EFD replication at the base

afausti:
Summit EFD replication is not currently working at the base. Disable MM2 and related source deployments for the moment. The need for having two SR deployments will be reviewed when we replace the InfluxDB Sink connectors with the telegraf-based connectors.


mareuter (approved):
Looks good to me.

PR #2685, opened 2023-10-26
DM-41350: Tell vault-secrets-operator to quote secrets

rra:
The templating in vault-secrets-operator apparently does support the quote function, so use it. This should fix problems with a Nublado database password that includes characters like ".


rra:
The theory here turns out to be entirely wrong. The secrets don't need to be quoted; that applies quote marks before base64 encoding, which isn't what is wanted at all.

PR #2686, opened 2023-10-27
DM-41424: Disable HTTP/2 in ingress-nginx

rra:
This interferes with keepalive for auth subrequests, and we can't use it effectively anyway for internal connections since we don't use TLS internal to the cluster.


rra:
This turns out to not work because auth-keepalive requires making the auth subrequets with Lua, which breaks a bunch of other things.

PR #2694, opened 2023-10-31
Change over to new secret framework for IDF prod

athornton:


athornton:
We'll test this during Patch and if everything goes well we can merge it then.

PR #2701, opened 2023-11-03
DM-40416: Add test alert topic

bsmartradio:
Adding a test alert topic to use for sending alerts from the pipeline to the broker. This is separate from the current alert stream which is receiving data from the simulator.


dspeck1 (approved):
Looks good.

PR #2719, opened 2023-11-13
Add initial config for exposurelog/narrativelog at USDF-dev/prod

athornton:


applications/narrativelog/values-usdfdev.yaml
```diff
@@ -0,0 +1,5 @@
+config:
+  site_id: usdfprod
```
ktlim, line 2:
```suggestion
  site_id: usdfdev
```

ktlim (approved):
This looks OK to me.  At some point we do want to be able to write to the database (which will likely move to a different database host), even if it's to a different table, but that can wait until we get something running.

PR #2725, opened 2023-11-14
DM-41705: Fix butler secrets and ingress

dhirving:


applications/butler/values.yaml
```diff
@@ -67,3 +67,7 @@ config:
   # -- URI to the file specifying the DirectButler configuration to be used
   # by the butler server
   configUri: ""
+  # -- The prefix of the path portion of the URL where the Butler service will
```
rra, line 70: Formatting nit: generally there should be a blank line between settings.

applications/butler/values.yaml
```diff
@@ -67,3 +67,7 @@ config:
...
+  # be exposed.  For example, if the service should be exposed at
+  # `https://data.lsst.cloud/api/butler`, this should be set to `/api/butler`
+  endpointPathPrefix: "/api/butler"
```
rra, line 73: We've used `pathPrefix` fairly uniformly elsewhere, so I would recommend that as well here.

rra (approved):
Looks good with a couple of nits.

PR #2726, opened 2023-11-15
Tickets/preops 4550: schedview app

ehneilsen:


rra:
Moved to a Phalanx ticket branch

PR #2751, opened 2023-11-28
tickets/PREOPS-4606: Add schedview-snapshot service

ehneilsen:
This is an initial attempt to add the schedview-snapshot survice, so far untested. But, at least some version needs to be checked into main before I can test and debug it, so this pull request is intended to get enough on main that I can begin testing and debugging.


applications/schedview-snapshot/.helmignore
```diff
@@ -0,0 +1,23 @@
+# Patterns to ignore when building packages.
```
athornton, line 1: We generally have not been using .helmignore files in our applications.  Is there something in particular that's been choking if you try to do a `helm template` on it in your development environment?

applications/schedview-snapshot/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+appVersion: tickets-preops-4603
```
athornton, line 2: Our appVersions are generally numeric, consisting of the GitHub release version of the application.  We use GitHub Actions to release a new version of the package on a tag.  Note that we use numeric versions, e.g. `6.1.1` rather than `v6.1.1`.

athornton (commented):
This is not usually the way we do appVersion, and historically we haven't had .helmignore files in applications.  Other than that, the content looks OK.

athornton:
I'm wrong about .helmignore, which is now part of our templates.  That's fine.

applications/schedview-snapshot/Chart.yaml
ehneilsen: Confirming our exchange on slack, it's okay to use the leading v here, and not worry about making multiple tags (with and without the leading v) for each release to support the conventions of both T&S and SQuaRE?

athornton (approved):
LGTM

applications/schedview-snapshot/Chart.yaml
athornton: Yes.

PR #2757, opened 2023-11-30
tickets/PREOPS-4619: enable schedview-snapshot in summit environment

ehneilsen:
In preparation for user testing with observing specialists next week


rra:
This seems stale and unlikely to be merged, and is also small enough to recreate if needed later.

PR #2760, opened 2023-11-30
DM-41950: Don't upload docs on PRs that don't modify docs/ itself.

jonathansick:
Now on pull requests, only upload the docs if the content in the `docs/` directory itself changes. This saves us CI time for changes that incidentally change the docs; we still CI that the docs will build and the docs themselves will update on the final merge to main.

In re-working the conditional, I've also found that we don't have a good overall strategy for tags; we'd need to generally build the docs for any tag. I've dropped "tags" from the `on` clause. But if we decide to start tagging Phalanx we'd want to revisit that.

This also adds `applications/argocd/values-*.yaml` to the paths considered as sources for docs.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -39,6 +39,7 @@ jobs:
               - "docs/**"
               - "applications/*/Chart.yaml"
               - "applications/*/values.yaml"
+              - "applications/argocd/values-*.yaml"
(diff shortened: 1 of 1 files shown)


jonathansick:
Example GitHub Actions run: https://github.com/lsst-sqre/phalanx/actions/runs/7053158409/job/19199669295

rra (approved):
This looks great. Thank you!

PR #2793, opened 2023-12-12
DM-41591: Include a visit with no template coverage in the upload.py test

hsinfang:


applications/prompt-proto-service-latiss/values-usdfdev-prompt-processing.yaml
```diff
@@ -13,7 +13,6 @@ prompt-proto-service:
   instrument:
     pipelines: >-
       (survey="SURVEY")=[${PROMPT_PROCESSING_DIR}/pipelines/LATISS/ApPipe.yaml,
-      ${PROMPT_PROCESSING_DIR}/pipelines/LATISS/SingleFrame.yaml,
```
kfindeisen: I'm not sure we will ever want to fall back on pipeline errors (especially if the "fallback" just means rerunning something that already ran!), but 👍 for something that's easy to roll back later.

kfindeisen (approved):

PR #2808, opened 2023-12-14
[DM-42069] Advance datalink to 2.1.6

cbanek:
tap schema release is already there.


gpdf (approved):
Looks good.  Tested on data-int.

PR #2814, opened 2023-12-18
Add the lsst.prompt namespace to Sasquatch

afausti:
Add the `lsst.prompt` namespace to Sasquatch, it assumes the messages have a `timestamp` column, and results will be recorded in the `lsst.prompt` database in InfluxDB. To start use the same set of tags used in `lsst.dm`.

### DIFF
--- a/applications/sasquatch/values-usdfdev.yaml
+++ b/applications/sasquatch/values-usdfdev.yaml
@@ -123,6 +123,12 @@ kafka-connect-manager:
         connectInfluxDb: "lsst.lf"
         topicsRegex: "lsst.lf.*"
         tags: benchmark_env,module,benchmark_type
+      lsstprompt:
+        enabled: true
+        timestamp: "timestamp"
+        connectInfluxDb: "lsst.prompt"
+        topicsRegex: "lsst.prompt.*"
+        tags: dataset_tag,band,instrument,skymap,detector,physical_filter,tract,exposure,patch,visit,run
 
 kafdrop:
   ingress:
@@ -146,6 +152,7 @@ rest-proxy:
       - lsst.camera
       - lsst.verify
       - lsst.lf
+      - lsst.prompt
 
 chronograf:
   ingress:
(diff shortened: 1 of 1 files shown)


applications/sasquatch/values-usdfdev.yaml
```diff
@@ -123,6 +123,12 @@ kafka-connect-manager:
         connectInfluxDb: "lsst.lf"
         topicsRegex: "lsst.lf.*"
         tags: benchmark_env,module,benchmark_type
+      lsstprompt:
+        enabled: true
+        timestamp: "timestamp"
```
kfindeisen, line 128: Do we need to do anything special to ensure there is a timestamp column, or do `analysis_tools` and `SasquatchDatastore` already enforce that?

applications/sasquatch/values-usdfdev.yaml
```diff
@@ -123,6 +123,12 @@ kafka-connect-manager:
...
+        connectInfluxDb: "lsst.prompt"
+        topicsRegex: "lsst.prompt.*"
+        tags: dataset_tag,band,instrument,skymap,detector,physical_filter,tract,exposure,patch,visit,run
```
kfindeisen, line 131: IIRC tags should be values that are frequently reused. In Prompt Processing, every visit-detector is a different run, so I recommend removing `run` from the list of tags.

Prompt Processing also organizes its processing by `group`, so it may be worth including that as a tag. However, this is also an undbounded set; in general, there is a 1-1 correspondence between groups and visits (there will be slightly more groups because some visits are cancelled).

kfindeisen (commented):
A few questions about the new config; I'm not sure what's the best choice for the tags.

applications/sasquatch/values-usdfdev.yaml
afausti: `analysis_tools` already add the timestamp column and a timestamp value with the right format.

applications/sasquatch/values-usdfdev.yaml
kfindeisen: Actually, wait -- if `run` is the name of the Butler run collection, then there is only one per night per (Gen 3) pipeline per service deployment. That might be small enough?

afausti: That sounds good. I'll remove the `run` tag and add the new `group` tag.

kfindeisen (approved):
Thanks! Let's go with this.

PR #2821, opened 2023-12-27
DM-42189: Update Butler for multi-repo server

dhirving:
Butler server is now able to serve multiple repositories from a single service.  Updated configuration to point to a repository index file instead of a single Butler configuration file.  Added ingresses for each of the repositories.

From the client's perspective, the server it is connecting to is "/api/butler/repo/dp02".  The theory is that in the future we may need separate Butler services for each repository, but at the moment it is excessive to have a separate Butler server for each.  

Also went ahead and added Butler to idfint, as well as a disabled configuration for idfprod.

### DIFF
--- a/applications/butler/templates/ingress-anonymous.yaml
+++ b/applications/butler/templates/ingress-anonymous.yaml
@@ -20,7 +20,6 @@ template:
       - host: {{ required "global.host must be set" .Values.global.host | quote }}
         http:
           paths:
-            {{- range $repositoryLabel := .Values.config.repositoryLabels }}
             # For direct end-user use of the Butler client library, the
             # Butler() convenience constructor must be able to load a
             # configuration file via unauthenticated HTTP.  This exists for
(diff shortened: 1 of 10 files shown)


applications/butler/templates/ingress-anonymous.yaml
```diff
@@ -20,25 +20,26 @@ template:
...
         http:
           paths:
+            {{- range $repositoryLabel := .Values.config.repositoryLabels }}
             # For direct end-user use of the Butler client library, the
             # Butler() convenience constructor must be able to load a
             # configuration file via unauthenticated HTTP.  This exists for
             # compatibility with the way Butler instances were configured prior
             # to the existence of the Butler server -- they are passed the URI
             # for a repository root on the filesystem or HTTP, from which a
             # configuration file is loaded.
```
rra, line 30: Very minor nit, but I would put the comment outside of the range, since there's no need to repeat it for each repository.

rra (approved):
Looks good to me with one very minor comment.

PR #2837, opened 2024-01-08
DM-42405: Update kafka IPs after sasquatch maintenance

hsinfang:


hsinfang:
My mistake here. It's https://github.com/lsst-sqre/phalanx/pull/2836

PR #2859, opened 2024-01-15
[DM-42456] Add the ghcr.io write token to Sasquatch secrets

afausti:
- Review Sasquatch secret definitions and conditionals. I've noticed that secret conditionals don't work if a component is enabled at the subchart level, for example, in `sasquatch/charts/strimzi-kafka/values.yaml`:

```
users:
  replicator:
    enabled: true
```
and  `sasquatch/secrets.yaml`:

```
replicator-password:
  description: >-
    replicator KafkaUser password.
  if: strimzi-kafka.users.replicator.enabled
```
A way around this was disabling the component by default at the subchart level and explicitly enabling it at the top level,  in `sasquatch/values-idfdev.yaml` in this case.

This is related to #2839,  where we made the `telegraf-password` secret unconditional in Sasquatch.

- Add ghcr.io write token to Sasquatch secrets, it is required by Strimzi to build custom Kafka Connect images.

### DIFF
--- a/applications/sasquatch/README.md
+++ b/applications/sasquatch/README.md
@@ -84,7 +84,7 @@ Rubin Observatory's telemetry service.
 | source-kapacitor.resources.requests.cpu | int | `1` |  |
 | source-kapacitor.resources.requests.memory | string | `"1Gi"` |  |
 | squareEvents.enabled | bool | `false` | Enable the Square Events subchart with topic and user configurations. |
-| strimzi-kafka | object | `{}` | Override strimzi-kafka configuration. |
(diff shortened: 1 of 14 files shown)


rra (approved):
Oh, yes, good catch, that is indeed an issue since the Phalanx tooling isn't smart enough to know that these subcharts are local and look at their defaults.

Note that you can move all the trues into the top-level `values.yaml` instead of the per-environment values files if you want. The tooling will pick that up.

afausti:
Thank you Russ! that's right, I can enable some components in `values.yaml` instead.

PR #2867, opened 2024-01-18
DM-42190: Add delegated token for datalinker links

dhirving:
In order to work with Butler client/server, Datalinker will need to get a delegated token for making a request to the Butler server on behalf of the caller.


rra (approved):
Looks good.

PR #2875, opened 2024-01-22
tickets/SITCOM-728

mareuter:


src/phalanx/models/environments.py
```diff
@@ -235,6 +235,34 @@ class EnvironmentConfig(EnvironmentBaseConfig):
         ),
     )
 
+    control_system_app_namespace: str | None = None
```
jonathansick, line 238: The default value for these need to be Pydantic fields. The Pydantic Field contains the default value, and docstring goes into the Field's `title` and `description` arguments. There is no docstring for Pydantic fields.

jonathansick: Additionally, I think we should move all these control system fields into a new Pydantic model class and then add that new `ControlSystemConfig` class as a `control_system` attribute on the `EnvironmentConfig` as an optional field. I.e.,

```
class EnvironmentConfig(EnvironmentBaseConfig):

    control_system: ControlSystemConfig | None = None
```

That'll help keep the control system application's global configuration fairly contained and clearly optional from Kubernetes/Argo CD environment configuration.

rra (commented):
I haven't looked at this in detail yet, but this is a lot less sprawling than I was afraid it was going to be, and it looks like a lot of the applications are now nested as subcharts. Thank you for doing all of that work! That will really help in keeping everything maintainable.

src/phalanx/models/environments.py
mareuter: The tests kept failing when trying to add the control system config object in the requested location. I had to move it into the `EnvironmentBaseConfig` definition to get things to pass.

rra: Yup, that's good. The difference between the two is whether the schema is also defined for the per-environment values files or just the merged settings. You want it defined for both, so putting it in base is correct.

I haven't seen a new push to this PR branch with that change, just FYI in case you'd thought you'd pushed it.

mareuter: The change was more extensive than I realized and I've just pushed it now. The way the control system configuration parameters are documented in `environments/values.yaml`, I've lost the ability to specify project wide settings and now have to do everything on a site level which isn't ideal. If you look at what I've done, can you tell me a way to handle project wide settings?

environments/values.yaml
```diff
@@ -181,9 +202,25 @@ applications:
...
+# If this section is used, it can have the following parameters:
+# _appNamespace_ (Set the namespace for the control system components)
+# _imageTag_ (The image tag to use for control system images)
+# _siteTag_ (The tag that tells the control system component where it is running)
+# _topicName_ (The Kafka identifier for control system topics)
+# _kafkaBrokerAddress_ (The Kafka broker address)
+# _kafkaTopicReplicationFactor_ (The Kafka topic replication factor)
+# _schemaRegistryUrl_ (The Schema Registry URL)
+# _s3EndpointUrl_ (The S3 URL for the environment specific LFA)
+controlSystem: null
```
rra, line 226: This is where you should set all of the project-wide defaults, by replicating the same structure for `controlSystem` here. In other words, every one of these options should be set here with a default and a documentation string, and then the individual environments can override those options as needed.

This means there are some defaults exposed to environments that don't run a control system, but that's fine since presumably only the control system applications will use those defaults and none of them are enabled. The tricky part will be figuring how what to do with the things that must be set, since there's no reasonable default. The right approach is to add validators, but I don't think all of that work has to be done up-front before this is merged since it will be a lot easier to improve once we can get this landed.

pyproject.toml
```diff
@@ -24,6 +24,7 @@ requires-python = ">=3.11"
 
 [project.scripts]
 phalanx = "phalanx.cli:main"
+update-shared-chart-version = "phalanx.control_system.update_shared_chart_version:run"
```
rra, line 27: Please add any new command-line functionality to the existing Phalanx CLI rather than adding a new CLI entry point (in this case, probably under `phalanx application`).

src/phalanx/control_system/update_shared_chart_version.py
```diff
@@ -0,0 +1,87 @@
...
+        with chart.open() as ifile:
+            values = yaml.safe_load(ifile)
+
+        dependencies = values["dependencies"]
+        for dependency in dependencies:
+            if dependency["name"] == DIR_MAP[opts.app_type]:
+                dependency["version"] = opts.chart_version
+
+        with chart.open("w") as ofile:
+            yaml.dump(values, ofile, sort_keys=False)
```
rra, line 66: This code should go into `phalanx.storage.config`. It's fairly similar to other code already provided by that module.

src/phalanx/control_system/update_shared_chart_version.py
```diff
@@ -0,0 +1,87 @@
+"""Script for updating shared chart versions."""
+
+import argparse
+import pathlib
+
+import yaml
+
+APPS_DIR = "applications"
+
+DIR_MAP = {"csc": "csc", "collector": "csc_collector"}
```
rra, line 10: I'm not really a fan of having two different names for the same thing and then doing a mapping. Can we find a way to consistently use the same name?

environments/templates/auxtel-application.yaml
```diff
@@ -0,0 +1,53 @@
...
+        - name: "global.controlSystem.topicName"
+          value: {{ .Values.controlSystem.topicName | quote }}
+        - name: "global.controlSystem.kafkaBrokerAddress"
+          value: {{ .Values.controlSystem.kafkaBrokerAddress | quote }}
+        - name: "global.controlSystem.kafkaTopicReplicationFactor"
+          value: {{ .Values.controlSystem.kafkaTopicReplicationFactor | quote }}
+        - name: "global.controlSystem.schemaRegistryUrl"
+          value: {{ .Values.controlSystem.schemaRegistryUrl | quote }}
+        - name: "global.controlSystem.s3EndpointUrl"
+          value: {{ .Values.controlSystem.s3EndpointUrl | quote }}
```
rra, line 49: These injections will also need to be added to `phalanx.services.ApplicationService._build_injected_values` so that `phalanx application lint` and `phalanx application template` work correctly.

applications/nublado/templates/vault-secrets.yaml
```diff
@@ -59,3 +59,13 @@ spec:
...
+---
+apiVersion: ricoberger.de/v1alpha1
+kind: VaultSecret
+metadata:
+  name: kafka-secret
+  labels:
+    {{- include "nublado.labels" . | nindent 4 }}
+spec:
+  path: "{{- .Values.global.vaultSecretsPath }}/ts/software/ts-salkafka"
+  type: Opaque
```
rra, line 71: This needs to be in some sort of conditional or it will break in other environments.

Note that we will need to move all of your secrets into the new secrets management framework in the future, which will require merging this with the nublado secret, but we can handle this as a later change. (I think the plan is to give you access to the relevant 1Password vault.)

rra (commented):
This looks good to me except for the changes noted. I didn't review all of the control system charts; I think we should merge them however they are and we can review them afterwards in the name of getting this all landed, since you've been using them and testing their functionality already.

applications/nublado/templates/vault-secrets.yaml
mareuter: Fixed.

environments/values.yaml
mareuter: OK, that's what I had originally done, but took it out to hide things more. I put back what I originally did now.

src/phalanx/models/environments.py
rra: To be clear, they should all still be nested under `controlSystem`, which I think was Jonathan's point. But `controlSystem` can be specified in `values.yaml`.

environments/templates/auxtel-application.yaml
mareuter: OK. Should I just inject them into the dictionary or trap them going in? Trapping means specifying a long list of apps to check which seems unsatisfactory.

mareuter: Eh, the more I dig at this, the more confused I get. I think I'm going to need some help getting this sorted out.

pyproject.toml
mareuter: I realized that this script is predicated on using the link mechanism for the shared charts rather than what is currently used. I'll have to redo it or just remove it.

environments/templates/auxtel-application.yaml
rra: Something like this:
```python
        if environment.control_system:
            extra = {
                "appNamespace": environment.control_system.app_namespace,
                "imageTag": environment.control_system.image_tag,
                "siteTag": environment.control_system.site_tag,
                "topicName": environment.control_system.topic_name,
                "kafkaBrokerAddress": (
                    environment.control_system.kafka_broker_address
                ),
                "kafkaTopicReplicationFactor": (
                    environment.control_system.kafka_topic_replication_factor
                ),
                "schemaRegistryUrl": (
                    environment.control_system.schema_registry_url
                ),
                "s3EndpointUrl": environment.control_system.s3_endpoint_url,
            }
            values.extend(
                {
                    f"global.controlSystem.{k}": v
                    for k, v in extra.items()
                    if v is not None
                }
            )
```
It's okay to inject them for every application.

pyproject.toml
mareuter: I've moved everything into the phalanx cli now.

environments/templates/auxtel-application.yaml
rra: It looks like `control_system` moved back into `EnvironmentConfig` somewhere along the line. It needs to be in `EnvironmentBaseConfig`.

mareuter: Arg. That was me not paying attention. Sigh.

mareuter: OK. I think I now have this working. The only thing that doesn't "work" is `appNamespace` in something like `phalanx application template`. That particular parameter is filled in the application template over in `environments/templates` which that command ignores. As the docs say, this does't quite reproduce Argo CD behavior, so I think this is good enough.

src/phalanx/control_system/update_shared_chart_version.py
mareuter: This differentiation is gone in the phalanx cli implementation.

src/phalanx/control_system/update_shared_chart_version.py
mareuter: Moved.

mareuter:
I think I've addressed all of the outstanding requests.

applications/sasquatch/charts/strimzi-kafka/values.yaml
```diff
@@ -8,6 +8,8 @@ kafka:
   version: "3.5.1"
   # -- Number of Kafka broker replicas to run.
   replicas: 3
+  # -- Number of down brokers that the system can tolerate.
+  disruption_tolerance: 0
```
afausti, line 14: Looks good, thanks for adding this configuration.

afausti (approved):
I reviewed the changes in Sasquatch, good job in extending the configuration and limiting the changes to TTS while keeping the default behavior elsewhere. Also thanks for the work on enabling the Kafka metrics.

PR #2876, opened 2024-01-22
[DM-42596] Add secret to push Strimzi Kafka Connect images to ghcr.io

afausti:
- Strimzi requires a secret of type `kubernetes.io/dockerconfigjson` to push images to ghcr.io. Add a secret to sasquatch with the `.dockerconfigjson` key and a `VaultSecret` resource to create the `sasquatch-connect-push-secret` secret.

### DIFF
--- a/applications/sasquatch/secrets.yaml
+++ b/applications/sasquatch/secrets.yaml
@@ -67,5 +67,5 @@ ts-salkafka-password:
   if: strimzi-kafka.users.ts-salkafka.enabled
 connect-push-secret:
   description: >-
-    Write token for pushing generated kafka-connect image to GitHub container registry.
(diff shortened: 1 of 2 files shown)


rra (approved):
This is a bit fiddly, but could you keep the `connect-push-secret` name in `secrets.yaml` since it's more human-readable and makes it clearer what this secret is for, and instead rename it in the template? For an example of how to do that, see:

https://github.com/lsst-sqre/phalanx/blob/main/applications/argocd/templates/vault-secrets.yaml

(That example doesn't actually rename anything, but hopefully it's clear how you would go about doing so. The quoted string after `.Secrets` is the name of the secret in Vault, which in this case would be `connect-push-secret`.)

afausti:
Thanks for the suggestion, Russ! That worked.

PR #2877, opened 2024-01-22
tickets/DM-42599: Add filestore-backup application

athornton:


applications/filestore-backup/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v2
+name: filestore-backup
+version: 1.0.0
+description: Tool to manage Google Filestore backups
```
rra, line 4: While this is a description of the underlying repository, I think it's not quite a description of the Kubernetes service being defined here. It's a `CronJob` so it's not really a tool. Maybe something like:

```suggestion
description: Create and manage lifecycle of Google Filestore backups
```

applications/filestore-backup/templates/cronjobs.yaml
```diff
@@ -0,0 +1,97 @@
...
+  successfulJobsHistoryLimit: 1
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          restartPolicy: Never
+          automountServiceAccountToken: true
+          {{- with .Values.tolerations }}
+          tolerations:
+{{ toYaml . | indent 12 }}
```
rra, line 16:
```suggestion
          tolerations:
            {{ toYaml . | nindent 12 }}
```
and similarly everywhere else you have this construct, so that it will match all of our other charts.

applications/filestore-backup/templates/cronjobs.yaml
```diff
@@ -0,0 +1,97 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: create-backup
```
rra, line 4: All the normal metadata is missing.

applications/filestore-backup/templates/cronjobs.yaml
```diff
@@ -0,0 +1,97 @@
...
+              {{- with .Values.tool.backup.debug }}
+              - name: "RUBIN_GOOGLE_FILESTORE_TOOLS_DEBUG"
+                value: "true"
+              {{- end }}
+            command: [ "create_backup" ]
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: purge-backup
```
rra, line 52: Normal metadata is missing.

I'm not really a fan of putting multiple resources in the same file unless they make no sense at all in isolation. I know we've done this in a few places, but for resources this long, I think it would make them more readable to break them into two files, one to create the backups and one to purge them.

applications/filestore-backup/values.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+# -- Affinity rules for the filestore-backup pods
+affinity: {}
+
+# The following will be set by parameters injected by Argo CD and should not
+# be set in the individual environment values files.
+image:
+  # -- Filestore-Backup image to use
+  repository: ghcr.io/lsst-sqre/rubin-google-filestore-tool
+  # -- Pull policy for the filestore-backup image
```
rra, line 29: Missing blank line before the next comment, throughout.

applications/filestore-backup/values.yaml
```diff
@@ -0,0 +1,67 @@
...
+  pullPolicy: "IfNotPresent"
+  # -- Tag of filestore-backup image to use
+  # @default -- The appVersion of the chart
+  tag: ""
+
+tool:
```
rra, line 35: I'm not really a fan of `tool` as the top-level key, since it doesn't communicate anything but implies that it does (like there's some part of this chart that deploys something that *isn't* the tool). For cases like this where you're just collecting settings, I would use `config`, which is consistent with lots of other Phalanx application charts.

applications/filestore-backup/values.yaml
```diff
@@ -0,0 +1,67 @@
...
+  # -- Filestore instance (e.g. "fshare-instance-dev")
+  # @default -- Must be overridden in environment-specific values file
```
rra, line 37: I have no strong opinion on this but most other places we use `None, must be set` for this case, which is a bit shorter.

applications/filestore-backup/values.yaml
```diff
@@ -0,0 +1,67 @@
...
+  # -- Zone for Filestore instance (e.g. "b" from "us-central1-b")
+  # @default -- Must be overridden in environment-specific values file
+  zone: ""
+  # -- File Share name for filestore instance.  Always "share1" unless
+  # storage is on an Enterprise tier
+  fileShare: "share1"
+  backup:
+    # -- Turn on debugging mode
+    debug: false
+    # -- Backup schedule
```
rra, line 48: For the schedules, it would be nice to add a `# @default -- ` comment that explains the default schedule in English rather than just showing the somewhat opaque cron value. (Also probably worth mentioning in the description of the setting that this is a UNIX cron expression.)

applications/filestore-backup/values.yaml
```diff
@@ -0,0 +1,67 @@
...
+global:
+  # -- Base path for Vault secrets
+  # @default -- Set by Argo CD
+  vaultSecretsPath: ""
+  # -- GCP Project ID
+  # @default -- Set by Argo CD
+  gcpProjectId: ""
+  # -- GCP Region
+  # @default -- Set by Argo CD
+  gcpRegion: ""
```
rra, line 67: Injecting these is a good idea since we know them already, but if you add any new injections, you need to also update `phalanx.services.application.ApplicationService._build_injected_values` to add the additional values.

applications/filestore-backup/templates/cronjobs.yaml
```diff
@@ -0,0 +1,97 @@
...
+              runAsGroup: 1000
+              capabilities:
+                drop:
+                  - all
+              readOnlyRootFilesystem: true
+            env:
+              - name: "RUBIN_GOOGLE_FILESTORE_TOOLS_PROJECT"
+                value: {{ .Values.global.gcpProjectId | quote }}
+              - name: "RUBIN_GOOGLE_FILESTORE_TOOLS_ZONE"
+                value: "{{ .Values.global.gcpRegion }}-{{ .Values.tool.zone }}"
```
rra, line 86: These values will only be set for GCP deployments which is going to produce syntactically correct but invalid configurations if someone mistakenly tries to deploy this on a non-GCP deployment. I think this should use the Helm `required` macro instead so that Helm throws an error if these values are not set.

environments/templates/filestore-backup-application.yaml
```diff
@@ -0,0 +1,38 @@
+{{- if (index .Values "applications" "filestore_backup") -}}
```
rra, line 1: Should be a dash, not an underscore.

rra (approved):

environments/templates/filestore-backup-application.yaml
athornton: Got it.

applications/filestore-backup/templates/cronjobs.yaml
athornton: Done (for gcpProjectId), on the grounds that it's unlikely someone would specify one and not the other and the templating is uglier for a concatenated value.

PR #2889, opened 2024-01-25
DM-42660: Use Butler server for Datalinker on IDF

dhirving:
Upgraded datalinker and Butler server to versions that are able to work with each other.  Configured datalinker deployments in all IDF environments to use Butler server instead of connecting directly to the Butler database.  Enabled Butler server on idfprod.

Butler repositories for services using Butler server are now configured by putting the list of repositories in an environment variable, instead of a URI pointing to a file containing the list of repositories.  The file is currently hosted on S3, which services will no longer have credentials for in the near future.

### DIFF
--- a/applications/datalinker/values-idfdev.yaml
+++ b/applications/datalinker/values-idfdev.yaml
@@ -1,3 +1,3 @@
 config:
   separateSecrets: true
-  overrideButlerRepositoryIndex: "s3://butler-us-central1-repo-locations/data-dev-repos-remote.yaml"
+  useButlerServer: true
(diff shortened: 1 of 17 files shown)


applications/datalinker/values-idfdev.yaml
```diff
@@ -1,2 +1,3 @@
 config:
   separateSecrets: true
+  overrideButlerRepositoryIndex: "s3://butler-us-central1-repo-locations/data-dev-repos-remote.yaml"
```
dhirving, line 3: There's some minor stupidity here -- downloading this index file requires the S3 credentials we're trying to get rid of. 🤦 

The new butler factory thing no longer requires this configuration to be in a file -- you can just pass in a dict of labels and URIs.  So I think it makes sense for the content of this file to be in phalanx instead.  When we go in to get rid of all the credentials I'll add an environment-level global configuration for this and pass the contents in an environment variable instead of using a URI.

rra (approved):
This looks good. Go ahead and merge this; I'm about to release 1.7.0 and will make some follow-up changes to Phalanx to reflect changes in the 1.7.0 release.

PR #2910, opened 2024-02-01
DM-42191: Support Butler client-server for vo-cutouts

dhirving:
Updated vo-cutouts to use Butler client/server instead of a direct postgres connection.  It now requests a delegated token from Gafaelfawr to pass to Butler server.

### DIFF
--- a/applications/vo-cutouts/Chart.yaml
+++ b/applications/vo-cutouts/Chart.yaml
@@ -4,7 +4,7 @@ version: 1.0.0
 description: "Image cutout service complying with IVOA SODA"
 sources:
   - "https://github.com/lsst-sqre/vo-cutouts"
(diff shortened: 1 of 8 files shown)


applications/vo-cutouts/Chart.yaml
```diff
@@ -4,7 +4,7 @@ version: 1.0.0
 description: "Image cutout service complying with IVOA SODA"
 sources:
   - "https://github.com/lsst-sqre/vo-cutouts"
-appVersion: 1.0.0
+appVersion: tickets-dm-42191
```
dhirving, line 7: Will update this before merging, once the new release is available.

rra (approved):
Looks great.

PR #2933, opened 2024-02-09
tickets/DM-42730: Move checkerboard to Phalanx roundtable-dev

athornton:
Move checkerboard to Phalanx roundtable-dev

Checkerboard has been rewritten as a FastAPI application, and moved into Phalanx.


applications/checkerboard/templates/ingress.yaml
```diff
@@ -0,0 +1,46 @@
...
+                  name: {{ template "checkerboard.fullname" . }}
+                  port:
+                    number: 8080
+            - path: "{{ .Values.ingress.path }}/redoc"
+              pathType: "Exact"
+              backend:
+                service:
+                  name: {{ template "checkerboard.fullname" . }}
+                  port:
+                    number: 8080
```
jonathansick, line 39: I don't think these are needed? The Prefix path type should handle all these.

jonathansick (approved):

applications/checkerboard/templates/ingress.yaml
athornton: Oh, right, I had those in there because they'd be anonymous, but that was before I realized that the whole thing (because it's using Basic auth rather than Gafaelfawr) is effectively anonymous.

athornton: I'm going to leave them here for now, because I'm going to retool checkerboard to use gafaelfawr auth, in which case I'm going to want each of those unauthenticated and the main route authed.

PR #2934, opened 2024-02-09
Added vault token authentication capability to install script

stvoutsin:
### Pull Request Description: 

This pull request introduces the capability to run the installation script using either approle, or token-based authentication.
It also introduces various refinements a summary of which follows:

### Changes Made:

**Modified argument parsing:**
- Added capability to the script to be run using named arguments, while maintaing backwards compatibility, where the script defaults to using the existing approle authentication method if run the way it was before
- For example install can be now run with the following ways:

        ./install.sh myenv VAULT_ROLE_ID=your-vault-id VAULT_SECRET_ID=your-secret
        ./install.sh myenv your-vault-id your-secret # Uses VAULT_ROLE_ID and VAULT_SECRET_ID
        ./install.sh myenv VAULT_TOKEN=your-vault-token
        ./install.sh myenv VAULT_TOKEN=your-vault-token VAULT_TOKEN_LEASE_DURATION=31536000
        ./install.sh ENVIRONMENT=myenv VAULT_TOKEN=your-vault-token
        ./install.sh ENVIRONMENT=myenv VAULT_TOKEN=your-vault-token VAULT_TOKEN_LEASE_DURATION=31536000
- Brief explanation: Environment argument can be passed either named or positional, for the auth arguments, if not passed in as named, default to using VAULT_ROLE_ID & VAULT_SECRET_ID. For VAULT_TOKEN it can only be passed in as a named argument, otherwise it would be tricky as approle and token auth cases both can be run using 2 parameters. (VAULT_TOKEN_LEASE_DURATION can also be optionally passed in, otherwise it defaults to one year) 

**Added Usage Guidance documentation:**

- Enhanced the usage message to provide clear instructions on how to use the script with environment variables and named arguments.
- Added documentation at the top of the script to explain and demonstrate how it can be used
 
**Dependency Checks:**

- Added a function to check for dependencies (yq, vault, kubectl, helm) at the beginning of the script. If any required command is not found, the script exits with an error message.

**Cleanup**
- Used functions where possible to avoid duplication
- Modified script based on suggestions from the shellcheck tool so that it validates

## Testing Performed:
- Tested the script with various scenarios, including different combinations of environment variables and arguments.
- Tested the various auth mechanisms (using VAULT_TOKEN and approle). In both cases vault-secrets-operator got the right secrets passed in, and the other apps were also run successfully

## Validation/Linting Performed:
- Validation of the script was done using Shellcheck
https://github.com/koalaman/shellcheck

## Additional Notes:

If you think this adds to much complexity to the bash script or were planning on going a different direction with it feel free to ignore, or suggestions on modifications are also welcome.
There is potential to also transform into a python script where named params would be more natural and easier to use.
Might be worth adding some unit tests to the script as I'm adding some logic with the argument parsing? If you think these changes make sense I can do that

### DIFF
--- a/docs/developers/local-development.rst
+++ b/docs/developers/local-development.rst
@@ -96,7 +96,7 @@ Finally, run the installer for the minikube environment.
 
 .. prompt:: bash
 
-  ./install.sh minikube $VAULT_TOKEN
+  ./install.sh minikube VAULT_TOKEN=$VAULT_TOKEN
 
 Access the Argo CD UI
 =====================
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -124,6 +124,27 @@ if [ -z "$VAULT_ROLE_ID" ] || [ -z "$VAULT_SECRET_ID" ]; then
     fi
 fi
 
+# Get environment configuration
+config="../environments/values-${ENVIRONMENT}.yaml"
+
+echo "Getting Git branch and remote information..."
+GIT_URL=$(git config --get remote.origin.url)
+# Github runs in a detached head state, but sets GITHUB_REF,
+# extract the branch from it.  If we're there, use that branch.
+# git branch --show-current will return empty in deatached head.
+GIT_BRANCH=${GITHUB_HEAD_REF:-$(git branch --show-current)}
+
+echo "Logging on to Vault..."
+
+VAULT_ADDR=""
+if grep '^vaultUrl:' "$config" >/dev/null; then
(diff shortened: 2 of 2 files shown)


rra (approved):
This looks great, thanks. Please go ahead and merge.

The long-term plan is indeed to incorporate the installer into the phalanx command-line tool so that we can use the Python models and libraries to do a lot of what this script is currently doing, but the work hasn't been a very high priority since we very rarely run the installer apart from using it for CI testing.

PR #2952, opened 2024-02-15
Add conditional to create VaultSecret resource

afausti:


applications/sasquatch/templates/vault-secrets.yaml
```diff
@@ -14,6 +14,7 @@ metadata:
 spec:
   path: "{{ .Values.global.vaultSecretsPath }}/pull-secret"
   type: kubernetes.io/dockerconfigjson
+{- if .Values.strimzi-kafka.connect.enabled }}
```
rra, line 17:
```suggestion
{- if (index .Values "strimzi-kafka" "connect" "enabled") }}
```

PR #2964, opened 2024-02-20
Modify mobu for the roe environment to test TAP only with list of queries

stvoutsin:
## What is this PR for?
This PR modifies the mobu chart config for the roe environmnent, to use the TAPQueryRunner to run three TAP Queries for health-checking purposes
 
## What type of PR is it?
Improvement

## How was this tested?
- Tested in dev environment
- Formatting tested with pre-commit


rra (approved):
Looks good, merge away.

PR #2986, opened 2024-02-26
add myself as admin especially in base

womullan:
Frossie says send this to Russ so i can see apps deployed in base when logged in as me ..

### DIFF
--- a/applications/argocd/values-base.yaml
+++ b/applications/argocd/values-base.yaml
@@ -23,3 +23,4 @@ argo-cd:
       policy.csv: |
         g, lsst-sqre:friends, role:admin
         g, lsst-sqre:square, role:admin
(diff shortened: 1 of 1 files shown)


womullan:
ok but argocd uses the github groups not individual names (at least in base) and I am in the sqre group I think - i added my id
..

rra (commented):
You're already a member of lsst-sqre:friends, so I don't think this will do anything. You should already have access. Are you not seeing applications when you go to base-lsp.lsst.codes/argo-cd and log in as yourself via GitHub? (Also, I think this would add a team named "womullan." At sites like the base where we use GitHub teams, I think this has to be a team.)

A line more like this would be used at USDF, where we list individuals, but the base (and summit and Tucson test stand) use GitHub team membership and I don't think there's any way to add individual users.

PR #2989, opened 2024-02-28
tickets/SITCOM-1251: Add LSSTCam CSCs to producers and connectors

mareuter:


afausti (approved):
Looks good. Note that `kafka-connect-manager` at USDF is not being used at the moment, and this section in the configuration will go away on Thursday when we'll remove InfluxDB OSS.

PR #2987, opened 2024-02-27
DM-42712: Create Prompt Processing service for Ops Rehearsal

hsinfang:
This will be the Prompt Processing service for LSSTComCamSim data.

PR #2993, opened 2024-02-29
DM-43090: next-visit-fan-out for LSSTComCamSim for dev

hsinfang:
Some were started in https://github.com/lsst-sqre/phalanx/pull/2987  but I'm now separating out the next-visit-fan-out part to here.

### DIFF
--- a/applications/next-visit-fan-out/values.yaml
+++ b/applications/next-visit-fan-out/values.yaml
@@ -2,6 +2,7 @@ knative:
   hscUrl: 'http://prompt-proto-service.prompt-proto-service-hsc/next-visit'
   latissUrl: http://prompt-proto-service.prompt-proto-service-latiss/next-visit
   lsstcomcamUrl: http://prompt-proto-service.prompt-proto-service-lsstcomcam/next-visit
+  lsstcomcamsimUrl: http://prompt-proto-service.prompt-proto-service-lsstcomcamsim/next-visit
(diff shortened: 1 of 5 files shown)


applications/next-visit-fan-out/values.yaml
```diff
@@ -2,6 +2,7 @@ knative:
   hscUrl: 'http://prompt-proto-service.prompt-proto-service-hsc/next-visit'
   latissUrl: http://prompt-proto-service.prompt-proto-service-latiss/next-visit
   lsstcomcamUrl: http://prompt-proto-service.prompt-proto-service-lsstcomcam/next-visit
+  lsstcomcamsimUrl: http://prompt-proto-service.prompt-proto-service-lsstcomcamsim/next-visit
   lsstcamUrl: http://prompt-proto-service.prompt-proto-service-lsstcam/next-visit
```
kfindeisen, line 6: I don't think I understand these URLs. Are they really _exactly the same_ for `usdfdev` and `usdfprod`?

If they're supposed to be the same, then I suggest removing those blocks from the per-environment files.

kfindeisen (approved):

applications/next-visit-fan-out/values.yaml
hsinfang: You are right. They are exactly the same in `usdfdev` and `usdfprod`. I'm removing this from the per-environment files.

PR #2995, opened 2024-02-29
Update sdm_schemas to 2.4.0 release

JeremyMcCormick:
Update tap image and datalinkPayloadUrl to 2.4.0 release of sdm_schemas.


charts/cadc-tap/README.md
gpdf, line 1: `tapSchema.image.tag` was not updated to 2.4.0 as it should have been.

Was this file hand-edited or did the pre-commit fail?  Why did the self-consistency check in the CI not see this error?

gpdf (changes requested):
Self-consistency between `values.yaml` and `README.md` must be fixed.

charts/cadc-tap/README.md
gpdf: This looks fine now.

gpdf (approved):
Looks good

PR #2997, opened 2024-02-29
tickets/DM-42884: move Vault to Phalanx

athornton:
Move Vault from old Roundtable to Phalanx Roundtable.


athornton:
We had a lot of the merging conversation during co-work 2024-Feb-29.  I broke up the vault docs and added configuration examples, as well as canonicalizing the name and underscore lengths for my other recently-added services, and both Russ and Jonathan agreed it would be fine to merge once I did that.

PR #3028, opened 2024-03-07
DM-43218: Add consdb KafkaUser

jonathansick:
This user has broad permissions as a Kafka consumer.

The app can authenticate as user "consdb" and password corresponding to the consdb-password from the sasquatch secret / vault item.

The password is automatically provisioned in vault with Phalanx secrets.


ktlim (approved):
Looks great, thanks!

PR #3048, opened 2024-03-12
DM-41040 Prompt Processing Resource Requests

dspeck1:
Add resource requests for queue proxy and user container.


applications/prompt-proto-service-hsc/values.yaml
```diff
@@ -7,6 +7,12 @@ prompt-proto-service:
     autoscaling.knative.dev/max-scale: "100"
     autoscaling.knative.dev/target-utilization-percentage: "60"
     autoscaling.knative.dev/target-burst-capacity: "-1"
+    queue.sidecar.serving.knative.dev/cpu-resource-request: "1"
+    queue.sidecar.serving.knative.dev/cpu-resource-limit: "2"
+    queue.sidecar.serving.knative.dev/memory-resource-request: "1Gi"
+    queue.sidecar.serving.knative.dev/memory-resource-limit: "2Gi"
+    queue.sidecar.serving.knative.dev/ephemeral-storage-resource-request: "400Mi"
+    queue.sidecar.serving.knative.dev/ephemeral-storage-resource-limit: "450Mi"
     # Update this field if using latest or static image tag in dev
```
kfindeisen, line 16: Why do the annotations not match the requests further down in the file (e.g., there the CPU limit is 1 and the memory limit is 8 GiB)?

dspeck1: The queue proxy is a separate container that queues requests for prompt processing in the user-container.  The queue proxy does not need as much memory as the prompt processing container since it only processing http requests and not performing other work.  Setting request values so it does not get starved out.

kfindeisen: Thanks for the clarification -- I missed that this was for `queue-proxy`.

PR #3049, opened 2024-03-13
DM-43205: Calibs fail to locally certify in Prompt Processing

kfindeisen:
This PR adds a flag to the Prompt Processing service that toggles caching calibs. While this is a useful feature, it causes the DM-41915 workaround to break in certain development environments. The flag is kept out of the published API as much as possible to make it easy to remove once calib handling is properly fixed.


kfindeisen:
Apologies for the late commit -- cleaning up the initContainers was a longstanding bit of technical debt that I'd meant to do the next time I was working on Phalanx, and then I forgot. 😳

hsinfang (approved):
Looks good and thanks for cleaning up other pieces too

PR #3050, opened 2024-03-13
DM-43223: Add Summit rubintv deployment

sebastian-aranda:


mareuter (commented):
A good chunk of the ``workers`` configuration is for the USDF and won't work on the summit.

sebastian-aranda:
> A good chunk of the `workers` configuration is for the USDF and won't work on the summit.

Worker pods configurations will be set on a different PR. See [DM-44243](https://rubinobs.atlassian.net/browse/DM-44243).

PR #3064, opened 2024-03-14
DM-40415: Add the alert distribution pipelineTask to Prompt Processing

bsmartradio:
Add alert stream credentials to prompt processing environment.

### DIFF
--- a/charts/prompt-proto-service/values.yaml
+++ b/charts/prompt-proto-service/values.yaml
@@ -71,6 +71,15 @@ apdb:
   # @default -- None, must be set
   namespace: ""
 
+alerts:
+  # -- Username for sending alerts to the alert stream
+  username: "kafka-admin"
+  # -- Server address for the alert stream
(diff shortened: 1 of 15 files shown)


charts/prompt-proto-service/values.yaml
```diff
@@ -71,6 +71,15 @@ apdb:
   # @default -- None, must be set
   namespace: ""
 
+alerts:
+  # -- Username for sending alerts to the alert stream
+  username: "kafka-admin"
```
hsinfang, line 76: Do you really want to set a default for `username`?  Maybe better leaving it unset and setting it in each instrument's service? 

Same for `server`.

hsinfang (approved):

charts/prompt-proto-service/values.yaml
bsmartradio: The server will be the same for all services, and at the moment we only have one admin account. This may change in the future but for now the setup is that everything but the topic is the same for each service. We can do it that way, but it isn't needed right now.

PR #3070, opened 2024-03-15
DM-43342: Add new infrastructure Argo CD project

rra:
Create an Argo CD project named infrastructure and put the core Phalanx infrastructure applications in it. Use this to write new RBAC rules for idfdev that divide users into admins and developers and block write access to infrastructure applications by developers.


athornton (approved):
This looks good.  All the applications you've tagged are definitely infrastructure, but it feels weird that that's all of them.  Probably right, though.

PR #3102, opened 2024-03-22
DM-43430: Add Butler server at USDF

dhirving:
Added a Butler server at USDF providing access to the `embargo` Butler repository, and configured datalinker to use it.  This fixes an issue where Firefly would display a 500 server error when trying to load images via datalinker for Live ObsCore TAP searches.

Previously Datalinker was not properly configured at USDF -- it was pointed at a Butler repository index file containing mostly repositories hosted as POSIX filesystems unreachable from the service, and was missing postgres configuration necessary to access the Butler DB.

### DIFF
--- a/applications/butler/values.yaml
+++ b/applications/butler/values.yaml
@@ -64,13 +64,9 @@ global:
   vaultSecretsPath: ""
 
 config:
-  # -- URI to the DirectButler repository index file listing the configurations
-  # for each repository to be hosted by this server.
-  indexUri: ""
-
-  # -- List of Butler repository labels which will be hosted by this server,
-  # matching those from the index file.
-  repositoryLabels: []
+  # -- Mapping from Butler repository label to Butler configuration URI for
+  # repositories which will be hosted by this server.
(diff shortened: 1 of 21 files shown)


applications/butler/values.yaml
```diff
@@ -64,13 +64,16 @@ global:
...
 
-  # -- List of Butler repository labels which will be hosted by this server,
-  # matching those from the index file.
-  repositoryLabels: []
+  # -- Postgres username used to connect to the Butler DB
+  # @default -- Use values specified in per-repository Butler config files.
+  pguser: ""
+
+  # -- URL for the S3 service where files for datasets are stored by Butler.
+  s3_endpoint_url: ""
```
rra, line 76: The Helm convention is to use lowerCamelCase for all settings, so s3EndpointUrl. It produces a bit of unfortunate mismatch between Helm and Python in order to follow the conventions of both ecosystems in their relevant code.

rra (approved):
Looks good apart from one parameter naming nit. Thank you!

PR #3120, opened 2024-03-26
Prevent merging branches with merges from main

rra:
Adopt the shared workflow used elsewhere in DM to ensure that we don't merge Phalanx branches that have merges back from main. DM style is to always rebase instead.


rra:
Abandoning this, since the underlying action doesn't support merge queues at all and appears to be orphaned.

PR #3123, opened 2024-03-26
Let Prompt Processing quietly accept empty surveys.

kfindeisen:
This PR adds an entry for the `""` survey to the LSSTComCamSim Prompt Processing config. This survey is likely to be associated with calibrations and other off-sky observations.


kfindeisen:
Closed in favor of #3104.

PR #3138, opened 2024-03-29
DM-43584: Add an option to turn off caching entirely

kfindeisen:
This PR adds a flag to the Prompt Processing service that toggles caching all inputs. This allows large-scale deployments that won't run out of storage space, but hurts preload performance. The flag is kept out of the published API as much as possible to make it easy to remove once the local repo is self-managing (DM-40193).


ktlim (approved):
Looks fine, including the reduction to 5GB, which I think still leaves a considerable margin of safety.

PR #3148, opened 2024-04-01
Turn off force source handling in prompt processing default

hsinfang:
I'm turning it off as the `dev` default too, so it's getting tested by default.

### DIFF
--- a/applications/prompt-proto-service-latiss/values-usdfdev-prompt-processing.yaml
+++ b/applications/prompt-proto-service-latiss/values-usdfdev-prompt-processing.yaml
@@ -11,7 +11,7 @@ prompt-proto-service:
 
   instrument:
     pipelines: >-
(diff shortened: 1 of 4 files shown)


kfindeisen (approved):
I'm not sure about turning it off in `dev`, but it's untestable either way (LATISS is too small, and ComCamSim doesn't have templates). As long as we don't forget to remove those later...

hsinfang:
Do it more properly with a ticket in https://github.com/lsst-sqre/phalanx/pull/3150

hsinfang:
Actually I've added templates for ComCamSim in dev.  

I'll edit and not change in `dev` then.  

Sorry for the change of mind about ticket.

hsinfang:
> Actually I've added templates for ComCamSim in dev.
But the templates I added don't seem to cover the raw.  

... I will refresh the dev raw then. But probably not today.

PR #3150, opened 2024-04-01
DM-43608: Turn off force source handling in prompt processing default

hsinfang:
I'm turning it off as the dev default too, so it's getting tested by default.


kfindeisen (approved):
I'm not sure about turning it off in `dev`, but it's untestable either way (LATISS is too small, ComCamSim doesn't have templates, and HSC rarely revisits the same field). As long as we don't forget to remove those later...

hsinfang:
Thanks! I'm leaving them on in `dev` then.

PR #3159, opened 2024-04-02
 Set APP version 1.0.0 and set S3_ENDPOINT in usdf_prod

womullan:


ktlim (changes requested):
I think this isn't going to work as is because nothing in the templates is reading the `environment:` clause.  We either need to make these specific items in the values or add a generic environment inclusion under `env:` in the template (like this, from another application):
```
       env:
{{- range $key, $value := .Values.environment }}
          - name: {{ $key | quote }}
            value: {{ $value | quote }}
{{- end }}
```
I'll work on this after lunch.

womullan:
This is what i did in obsloctap .. but then i guess i did not actually access S3 ..

ktlim (approved):

PR #3164, opened 2024-04-04
[DM-43580] Change Sasquatch Kafka min.insync.replicas setting to 2 at USDF

afausti:
@mareuter planning on rolling this out today at USDF during patch


mareuter (approved):
Looks good to me.

PR #3165, opened 2024-04-04
DM-43555: Add phalanx environment install command

rra:
Port the installer/install.sh script to Python as the new phalanx environment install command. Add a strongly worded confirmation message before proceeding, which can be overrriden with a command line flag for CI testing.

Expand the Helm storage layer and add Argo CD and Kubernetes storage layers to support the installer. Use kubectl rather than the Python Kubernetes libraries to manipulate Kubernetes objects, since it's simpler and more straightforward for the type of actions the installer needs to take.

Introduce a new way of passing around Vault credentials, and support authenticating to Vault with an AppRole, at least in the context of the installer. (The secrets and vault commands still require tokens.)

Remove the old installer and switch GitHub Actions CI over to the new command.


src/phalanx/storage/command.py
athornton, line 1: Line 36 still says "Helm", when it's a generic command processor.

This is a thing I keep reinventing over and over.  My current design--well, you'll see it in the runlab PR--just captures and decodes stdout and stderr and returns a structure with the returncode and those--effectively, your "capture()" method.

We need to standardize it and move it into Safir-or-something, and give it both "run" and "run_async" methods (the second using asyncio.subprocess).

src/phalanx/storage/config.py
```diff
@@ -312,6 +314,67 @@ def get_environment_chart_path(self) -> Path:
...
+        repo = Repo(str(self._path))
+        try:
+            origin = repo.remote("origin")
+        except ValueError as e:
+            msg = 'Current repository has no remote named "origin"'
+            raise GitRemoteError(msg) from e
+        if not origin.url:
+            raise GitRemoteError('Remote "origin" has no URL')
+
+        # If the URL is not an https URL, accept a few forms of github.com
```
athornton, line 357: Nice.  Also something that probably ought to be lifted into a shared library.

src/phalanx/cli.py
```diff
@@ -334,6 +353,93 @@ def environment() -> None:
...
+    else:
+        msg = (
+            "Either VAULT_TOKEN or both VAULT_ROLE_ID and VAULT_SECRET_ID"
+            " must be set"
+        )
+        raise click.UsageError(msg)
+
+    # Prompt the user unless they specifically said not to.
+    if not force_noninteractive:
+        print(_INSTALL_WARNING.format(environment=environment))
```
athornton, line 433: I don't know if click.confirm will let you do this, and maybe it's being TOO aggressive...

...but given my tendency to speed through installers without thinking about them, I feel like it might be a good idea to generate a nonce -- maybe the classical six-digit random PIN -- display it, and make the user type it or at least copy-paste it, rather than just "y-Return".

src/phalanx/storage/kubernetes.py
```diff
@@ -0,0 +1,80 @@
+"""Storage layer for direct Kubernetes operations."""
+
+from __future__ import annotations
+
+from .command import Command
+
+__all__ = ["KubernetesStorage"]
+
+
+class KubernetesStorage:
```
athornton, line 10: On the one hand, simpler...but on the other hand it feels like we already HAVE a lot of it mocked out for testing if we were using the Safir driver like we do in the Nublado controller.  I mean...we already have secrets and namespaces in there, right?

athornton (approved):
LGTM.

I have some questions about "can we make this a shared library" around the Command class mostly, and...while I am certainly not someone who should throw stones about just shelling out via Command or its moral equivalent (`git` versus GitPython in particular), your kubectl driver feels like, except for waiting for rollouts, it's stuff we already have in Safir.

src/phalanx/storage/kubernetes.py
rra: Yes, the thing we don't have is waiting for a rollout to complete. I don't think that's a trivial thing to implement with timeout and error handling, etc., and we're doing everything else with command-line tools (Helm and Argo CD), so it didn't seem worth the effort to reimplement it in Python.

The testing code isn't that useful for Phalanx since Phalanx already tests the installer with a good integration test, making unit tests (particularly of this stuff) fairly pointless.

src/phalanx/cli.py
rra: I'm going to merge this in its current form, but I added to my to-do list determining the current configured Kubernetes cluster and displaying that as part of the prompt to make it easier for people to realize when they're pointing to the wrong cluster.

PR #3181, opened 2024-04-08
Adjust resource request and limits fork Kafka at TTS

afausti:
We added default resource rquests and limits for Kafka in Sasquatch. That doesn’t seem enough for TTS which runs the Control System.
For TTS, we are requesting 80G of memory and 4 CPUs for Kafka.


mareuter (approved):
Looks good to me.

PR #3194, opened 2024-04-11
Update PANDA values for nublado/idfint

athornton:


timj (approved):
I didn't make the original request but looking at the ticket it seems like this change is consistent with that. It would have been easier to read the diff if IDDS_CONFIG had not been moved.

cc/ @wguanicedew

athornton:
Agreed, but I also wanted all the PANDAs config alphabetically sorted and together, which is why I moved `IDDS_CONFIG`.  Canonicalizing quotation of values in this section was mostly just to satisfy my OCD.

PR #3199, opened 2024-04-12
DM-43837: Flesh out Helm docs for Sasquatch values.yaml

rra:
Add documentation of more individual settings and add @default markers to elide complex defaults. This should make the generated documentation less wide and thus easier to browse.


afausti (approved):
Awesome! thanks for fixing the helm docs, especially the work documenting the InfluxDB Enterprise subchart. The generated docs look great.

PR #3200, opened 2024-04-12
DM-43837: Add more documentation pointers for Sasquatch

rra:
Link to the Sasquatch documentation and add references to the sources of various Sasquatch components to its Chart.yaml.


afausti (approved):
Thanks for doing that!

PR #3236, opened 2024-04-22
IDF demo.lsst.cloud

athornton:


applications/tap/values-idfdemo.yaml
jonathansick, line 1: Is tap being used in the demo? Perhaps delete if its stale?

applications/portal/values-idfdemo.yaml
jonathansick, line 1: Is portal being used in the demo? Perhaps delete if this file is stale?

environments/values-idfdemo.yaml
```diff
@@ -0,0 +1,23 @@
...
+butlerRepositoryIndex: "s3://butler-us-central1-repo-locations/data-dev-repos.yaml"
+butlerServerRepositories:
+  dp02: "https://demo.lsst.cloud/api/butler/repo/dp02/butler.yaml"
+gcp:
+  projectId: "science-platform-demo-9e05"
+  region: "us-central1"
+  clusterName: "science-platform-demo"
+onepassword:
+  connectUrl: "https://roundtable-dev.lsst.cloud/1password"
+  vaultTitle: "Phalanx data-demo.lsst.cloud"  # FIXME?  Wrong but consistent
```
jonathansick, line 13: We'd need to change it though Chile IT, so probably no use for the comment here.

environments/values-idfdemo.yaml
```diff
@@ -0,0 +1,23 @@
...
+appOfAppsName: "science-platform"
+butlerRepositoryIndex: "s3://butler-us-central1-repo-locations/data-dev-repos.yaml"
+butlerServerRepositories:
+  dp02: "https://demo.lsst.cloud/api/butler/repo/dp02/butler.yaml"
+gcp:
+  projectId: "science-platform-demo-9e05"
+  region: "us-central1"
+  clusterName: "science-platform-demo"
+onepassword:
+  connectUrl: "https://roundtable-dev.lsst.cloud/1password"
```
jonathansick, line 12: Shouldn't this use production? No need to have exposure to a test cluster.

jonathansick (commented):
Looks good. I see some stale values files that we might want to just delete rather than keep around. I also think would should use the production 1password connect server.

environments/values-idfdemo.yaml
athornton: My reasoning is that this cluster is closer to idf-dev than it is to anything else.  But I also don't really care.  Prod's fine.

PR #3242, opened 2024-04-25
DM-43934: USDF Alert Stream Broker Dev Kafdrop

dspeck1:
Add in kafdrop helm subchart for alert-stream-broker and enable values for usdf-dev-alert-stream-broker. Add kafdrop kafa user template.  The usdf-dev-alert-stream-broker is connecting to the external bootstrap because TLS is disabled for that listener.


afausti:
Sasquatch also uses Kafdrop. I would suggest making it a shared chart in Phalanx to avoid repetition.

rra:
Is this branch something you're still working on?

PR #3243, opened 2024-04-25
DM-44070: Remove alert stream simulator

bsmartradio:


afausti:
Sasquatch also uses Kafdrop. I would suggest making it a shared chart in Phalanx to avoid repetition.

bsmartradio:
> Sasquatch also uses Kafdrop. I would suggest making it a shared chart in Phalanx to avoid repetition.

We won't be adding kafdrop in this ticket. I was temporarily rebasing to ensure any kafdrop testing would not break the alert stream broker.

dspeck1 (approved):
LGTM

PR #3254, opened 2024-04-29
tickets/DM-43999: Fix label on sasquatch pod monitors

mareuter:


afausti (commented):
This looks very interesting, so adding namespace labels per application seems possible. I let @rra review it, as I'm not familiar with the changes in the CLI. But looks great.

environments/values.yaml
```diff
@@ -240,6 +240,9 @@ applications:
   # -- Enable the vo-cutouts application
   vo-cutouts: false
 
+# -- Add labels for an application namespace.
+namespaceLabels: {}
```
rra, line 244: The placement of this setting implies that it is being added to every namespace, but it's only added to sasquatch. Is the intent to apply it to other namespaces later and this is just for testing? If we're going to eventually need to pick which namespaces it's applied to and which it isn't, I think this needs different structure (perhaps a mapping of namespace names to labels).

src/phalanx/models/environments.py
```diff
@@ -205,6 +205,15 @@ class EnvironmentBaseConfig(CamelCaseModel):
         ),
     )
 
+    namespace_labels: dict[str, Any] | None = Field(
```
rra, line 208: Any shouldn't be used in types. Right now this is `dict[str, dict[str, str]]`.

environments/values.yaml
mareuter: It was supposed to be generic, but I see your point. If another app uses the config parameter, they'll pick up labels they probably don't want. Let me think about how this could be done with you mapping suggestion.

src/phalanx/models/environments.py
mareuter: Linter didn't like that one. Likes `dict[str, str]` instead, but not sure that's right either.

environments/values.yaml
mareuter: OK, I went a different route by using applications to hide the labelling from others that don't care. Let me know if this is going in the right direction.

rra (approved):
I don't think we need the intermediary `applications` layer, or at least I can't think of any other types of namespaces that we would have. Other than that, looks good to me. (Ideally we should retrofit all existing applications to support this, and also add this to the template for new applications, but that can be future work.)

src/phalanx/models/environments.py
rra: Oh, right, sorry, I had one too many layers there.

mareuter:
I managed to take out the need for the `application` subspace. Can you have a look once more please?

rra (approved):
Perfect, looks great. Thank you!

PR #3263, opened 2024-04-29
Rework fastapi-bootcamp app with newer starter

athornton:


athornton:
Ready for review.

applications/fastapi-bootcamp/templates/ingress.yaml
```diff
@@ -22,7 +21,7 @@ template:
       - host: {{ required "global.host must be set" .Values.global.host | quote }}
         http:
           paths:
-            - path: {{ .Values.config.pathPrefix | quote }}
+            - path: "/fastapi-bootcamp"
```
athornton, line 24: Sorry, Jonathan.  Take it up with Russ.

rra: ...that's an odd way of saying "hey, I think there's a bug in the starter." I'll go fix the starter; I added the config setting but then forgot to use it.

rra (approved):
That all looks good except for the ingress path change, which you can revert (the starter should be fixed shortly).

applications/fastapi-bootcamp/templates/ingress.yaml
athornton: Oh, the history behind that is that on the first iteration of that, I had "/fastapi-bootcamp", and Jonathan caught it.  I figured this was a disagreement over how opinionated we wanted to be.

PR #3264, opened 2024-04-29
DM-43581: Add pull-secret to rubintv

ctslater:


applications/rubintv/templates/vault-secrets.yaml
```diff
@@ -7,3 +7,13 @@ metadata:
...
+---
+apiVersion: ricoberger.de/v1alpha1
+kind: VaultSecret
+metadata:
+  name: pull-secret
+  labels:
+    {{- include "rubintv.labels" . | nindent 4 }}
+spec:
+  path: "{{- .Values.global.vaultSecretsPath }}/pull-secret"
+  type: kubernetes.io/dockerconfigjson
```
sebastian-aranda, line 19: This looks good, we are not going to use it for the frontend deployment, but we should use it for the workers deployment. We need to first update the worker images though, from: `lsstts/rubintv-broadcaster:c0030_usdf` to `ts-dockerhub.lsst.org/rapid-analysis:c0036`.

`lsstts/rubintv-broadcaster:c0030_usdf` is too old! from almost 1 year ago: https://hub.docker.com/r/lsstts/rubintv-broadcaster/tags.

sebastian-aranda (changes requested):
Looks good, this should allow us pulling images from `ts-dockerhub.lsst.org`. I left a couple comments though so you can take a look :pray:.

sebastian-aranda (approved):
Looks good, thanks for going through changes :+1: .

PR #3267, opened 2024-04-30
Add RSP_SITE_TYPE to nublado values

athornton:


applications/nublado/values-tucson-teststand.yaml
```diff
@@ -24,6 +24,7 @@ controller:
         LSST_SCHEMA_REGISTRY_URL: "http://sasquatch-schema-registry.sasquatch:8081"
         PGPASSFILE: "/opt/lsst/software/jupyterlab/secrets/postgres-credentials.txt"
         PGUSER: "oods"
+        RSP_SITE_TYPE: "telescope"
```
rra, line 27: This is probably not telescope since it uses Kafka instead, but I don't think we know what it is yet, and we can cross that bridge later.

rra (approved):
Seems fine to me.

applications/nublado/values-tucson-teststand.yaml
athornton: OK.  The immediately-important one (which I'm using at usdf-dev) is "staff", and (after some correction) I seem to be getting reasonable guidance from the terminal launch message (and the env var is set).

In the longer run we probably want to lift this setting higher anyway, as we were discussing.

PR #3275, opened 2024-05-02
Remove bucketmaker from sasquatch (it lives in monitoring now)

athornton:


afausti (approved):
Good catch, thank you!

PR #3292, opened 2024-05-06
tickets/PREOPS-5118: Update the schedview snapshot dashboard running in phalanx to support access to an S3 bucket

ehneilsen:
This change requires not just updating the schedview version, but also requires configuring phalanx to provide it with the secret it needs.
For this, I was guided by a combination of the phalanx documentation, the configuration of plot-navigator (from which some changes were cut-and-pasted), and [this thread on slack](https://lsstc.slack.com/archives/C2JP8GGVC/p1714426470147829).
It seems to work like this, but I'm not 100% sure copying the secret from nublado using the butler's vault is the right thing to do, in that I think it give this app write access it doesn't need (although KT [says this is okay](https://lsstc.slack.com/archives/C2JP8GGVC/p1714580872053049?thread_ts=1714426470.147829&cid=C2JP8GGVC)).

### DIFF
--- a/applications/schedview-snapshot/secrets.yaml
+++ b/applications/schedview-snapshot/secrets.yaml
@@ -1,0 +1,9 @@
+# Follow the example of plot-navigator, but only getting the aws-credentials
+
+"aws-credentials.ini":
+  description: >-
+    Cloud storage for access to the LFA, from which scheduler snapshots
+    can be loaded.
+  copy:
+    application: nublado
+    key: "aws-credentials.ini"
--- a/applications/schedview-snapshot/templates/_helpers.tpl
+++ b/applications/schedview-snapshot/templates/_helpers.tpl
@@ -1,29 +1,4 @@
 {{/*
-Expand the name of the chart.
-*/}}
-{{- define "schedview-snapshot.name" -}}
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
-{{- end }}
-
-{{/*
-Create a default fully qualified app name.
-We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
-If release name contains chart name it will be used as a full name.
-*/}}
-{{- define "schedview-snapshot.fullname" -}}
-{{- if .Values.fullnameOverride }}
-{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
-{{- else }}
-{{- $name := default .Chart.Name .Values.nameOverride }}
-{{- if contains $name .Release.Name }}
-{{- .Release.Name | trunc 63 | trimSuffix "-" }}
-{{- else }}
-{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
-{{- end }}
-{{- end }}
-{{- end }}
-
-{{/*
 Create chart name and version as used by the chart label.
 */}}
 {{- define "schedview-snapshot.chart" -}}
--- a/applications/schedview-snapshot/templates/_helpers.tpl
+++ b/applications/schedview-snapshot/templates/_helpers.tpl
@@ -1,4 +1,29 @@
 {{/*
+Expand the name of the chart.
+*/}}
+{{- define "schedview-snapshot.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "schedview-snapshot.fullname" -}}
+{{- if .Values.fullnameOverride }}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- $name := default .Chart.Name .Values.nameOverride }}
+{{- if contains $name .Release.Name }}
+{{- .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
+{{- end }}
+
+{{/*
 Create chart name and version as used by the chart label.
 */}}
 {{- define "schedview-snapshot.chart" -}}
--- a/applications/schedview-snapshot/templates/deployment.yaml
+++ b/applications/schedview-snapshot/templates/deployment.yaml
@@ -34,6 +34,8 @@ spec:
           env:
             - name: BOKEH_ALLOW_WS_ORIGIN
               value: {{ .Values.global.host }}
+            - name: AWS_SHARED_CREDENTIALS_FILE
+              value: /opt/schedview/aws-credentials.ini
           ports:
             - name: "http"
               containerPort: 8080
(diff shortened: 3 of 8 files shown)


applications/schedview-snapshot/secrets.yaml
```diff
@@ -0,0 +1,9 @@
+# Follow the example of plot-navigator, but only getting the aws-credentials
+
+"aws-credentials.ini":
+  description: >-
+    Cloud storage for access to the LFA, from which scheduler snapshots
+    can be loaded.
+  copy:
+    application: nublado
+    key: "aws-credentials.ini"
```
rra, line 9: Does this application need all of the credentials that a user would have, to all of the S3 buckets, or does it only need access to a specific subset of S3 buckets? I'm worried that this is too many permissions and it would be better to create more scoped credentials specific to the application, but I'm not sure what the application does. If it should be able to access anything a user can access from a notebook (read and write), then this is correct.

applications/schedview-snapshot/templates/_helpers.tpl
```diff
@@ -1,3 +1,28 @@
...
+{{- else }}
+{{- $name := default .Chart.Name .Values.nameOverride }}
+{{- if contains $name .Release.Name }}
+{{- .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
+{{- end }}
+
```
rra, line 25: We're no longer using this pattern with any new applications because it's unnecessary complexity that's never used in Phalanx. I recommend deleting this all again and just putting schedview-snapshot literally in your YAML files.

applications/schedview-snapshot/templates/deployment.yaml
```diff
@@ -77,3 +82,7 @@ spec:
           emptyDir: {}
         - name: slashdatcache
           emptyDir: {}
+        - name: schedview-snapshot-secrets
+          secret:
+            secretName: {{ include "schedview-snapshot.fullname" . }}
```
rra, line 87: Here, for example, this should just be "schedview-snapshot" instead of this more complex include template.

applications/schedview-snapshot/templates/vault-secrets.yaml
```diff
@@ -0,0 +1,11 @@
+# Follow the example of plot-navigator
+
+apiVersion: ricoberger.de/v1alpha1
+kind: VaultSecret
+metadata:
+  name: {{ template "schedview-snapshot.fullname" . }}
+  labels:
+    {{- include "schedview-snapshot.labels" . | nindent 4 }}
+spec:
+  path: "{{ .Values.global.vaultSecretsPath }}/butler-secret"
```
rra, line 10: The butler-secret pattern is no longer supported and we're trying to move away from it. You'll see that plot-navigator has an if conditional on separateSecrets. Since this is a new secret, you should use only the separateSecrets approach, which is the new way to do this.

In other words, path should instead be:
```
path: "{{ .Vault.global.vaultSecretsPath }}/schedview-snapshot"
```

applications/schedview-snapshot/templates/vault-secrets.yaml
```diff
@@ -0,0 +1,11 @@
+# Follow the example of plot-navigator
+
+apiVersion: ricoberger.de/v1alpha1
+kind: VaultSecret
+metadata:
+  name: {{ template "schedview-snapshot.fullname" . }}
```
rra, line 6: Here too, change this to just "schedview-snapshot"

rra (commented):
Note that only usdfint has switched to the new secrets management system, so for usdfdev you (or one of the USDF administrators) will need to manually put the AWS credentials that your application needs into the aws-credentials.ini key under the schedview-snapshot Vault path for your application.

applications/schedview-snapshot/secrets.yaml
ehneilsen: The application needs read access to the LFA (is this just one bucket, or many?): all of URIs for objects I need access to begin `s3://rubin:rubinobs-lfa-cp/Scheduler`. (There will need to be a deployment at the summit as well, with access to the corresponding summit URIs, but I'm not there yet, and summit staff will probably be the ones to configure phalanx for that environment, and it will be handled in a separate ticket and pull request.)
I don't know how to get credentials that limit the access to this subset. When I asked on slack (see the thread I linked to in the original comment), it sounded like trying this would make things more complicated, and it would be acceptable and simpler to use the nublado credentials. If I can get help getting something more restricted, that would be great.

applications/schedview-snapshot/templates/vault-secrets.yaml
ehneilsen: How do I create (or who do I ask to create) an entry in the vault for schedview-snapshot? How do I create (or who do I ask to create) the secrets that need to go into the vault?

applications/schedview-snapshot/templates/_helpers.tpl
ehneilsen: Change made in commit ca02a336.

applications/schedview-snapshot/templates/deployment.yaml
ehneilsen: Change made in commit ca02a336.

applications/schedview-snapshot/secrets.yaml
rra: Well, my security advice would be to get a more restricted token that only has read access to the objects or bucket that you are using, put that token in the value for the `aws-credentials.ini` secret in the Vault path for this application, and remove the copy directive (eveything else about this `secrets.yaml` is file). Using correctly-scoped tokens is generally better, in part because it will protect against mistakes.

But I'm not involved in running the USDF cluster and it's their security decision, so if they really want you to use the more-powerful tokens that are injected into user notebooks, it's their call. @ktlim, are you the person who handles this?

applications/schedview-snapshot/templates/vault-secrets.yaml
rra: You can ask on #ops-usdf on Slack for help with Vault. I'm not sure if they'll want to give you access so that you can make the change yourself or make it for you. Where you get the secret depends on the answer to the above question: either someone should give you a new secret, or you (or someone helping you) will get it from the nublado Vault secret.

applications/schedview-snapshot/secrets.yaml
ktlim: The rubin-summit-users credential is currently used for read-only access to all buckets that staff users will need access to on an everyday basis so that multiple profiles don't need to be used.  (The credential also has read/write access to the buckets containing Butler repo user collections.)

In the near future (within the next few weeks) when we move to the embargo rack, which necessitates multiple profiles anyway due to having multiple endpoint URLs, I have been considering having an additional LFA-specific read-only credential that would be used with `s3://lfa@rubinobs-lfa-cp/Scheduler` URIs.  That profile would also be installed in the nublado `butler-secret`, but a more limited custom LFA-only profile could also be used.

rra: It sounds like it'll be a couple of weeks before things are ready to switch to more limited secrets, so this copy directive is correct for now. But you'll still need to change the reference to butler-secret to schedview-snapshot instead, down below where you create the `VaultSecret` resource, since we're switching everything over to per-app secrets. usdfint is done, and usdfdev and maybe even usdfprod should hopefully be done next Thursday.

Once the more restrictive secrets have been implemented, the more restricted secret should be the only one included in the schedview-snapshot secret, and at that point the copy directive here should be removed.

rra (approved):

PR #3302, opened 2024-05-08
Make lsst-sqre/friends and lsst/Ops argocd admins for idfdemo

athornton:


athornton:
self-merging.  It's only Demo if I got it wrong.

PR #3324, opened 2024-05-14
DM-44244: Create rubintv-dev application and add it to Summit environment

sebastian-aranda:


applications/rubintv-dev/values-summit.yaml
```diff
@@ -0,0 +1,40 @@
...
+  siteTag: "summit"
+
+  imagePullSecrets:
+  - name: pull-secret
+
+  frontend:
+    debug: true
+    env:
+    - name: S3_ENDPOINT_URL
+      value: "https://s3.rubintv.cp.lsst.org"
```
mareuter, line 11: Tip: You can use some YAML tricks to not duplicate information that's used elsewhere in the file. See: https://github.com/lsst-sqre/phalanx/blob/main/applications/simonyitel/values-tucson-teststand.yaml at the `ccoods` and `mtaos` entries.

docs/applications/rubintv-dev/index.rst
```diff
@@ -0,0 +1,32 @@
+.. px-app:: rubintv-dev
+
+#################################################################
+rubintv-dev — Real-time display front end development application
+#################################################################
+
```
mareuter, line 6: Can we put a bit of text that flags this as a development app? I see there is something in the title, but a little more in the main body would be good.

mareuter (approved):
Some minor things to think about, but otherwise this looks good.

applications/rubintv-dev/values-summit.yaml
sebastian-aranda: Nice!! I didn't know about this trick, thank you one more time Michael :pray:, I'll make the change.

PR #3329, opened 2024-05-14
[DM-44337] Configure timestamp precision in the Telegraf Kafka Consumer input plugin 

afausti:
The precision configuration setting determines how much timestamp precision is retained in the points received from input plugins. All incoming timestamps are truncated to the given precision.
We are doing that in the Telegraph Kafka input plugin configuration to make sure timestamps are truncated to microseconds (same behaviour of the InfluxDB Sink connectors)


mareuter (approved):
Looks good to me.

PR #3334, opened 2024-05-16
tickets/DM-44373

athornton:
Adopt service-discovery-endpoint capable Giftless build.

In fact, we don't need this yet, because `LEGACY_ENDPOINTS` currently defaults to `true` but I want to make sure that in the future, when the default changes, nothing breaks.


rra (approved):
LGTM

PR #3353, opened 2024-05-20
DM-44141: Migrate APDB/Butler credentials from .pgpass to db-auth

kfindeisen:
This PR has Prompt Processing download a `db-auth.yaml` file from the Vault and configure it for use by the service. The deletion of `PGUSER` is conditional on the assumption that `db-auth.yaml` will be used for Butler and Cassandra authentication, with `.pgpass` needed only for connecting to SQL APDBs.

It might be useful to make these blocks conditional on the Vault secret actually containing the `pgpass_file` or `db-auth_file` keys, but I don't know how to do so.


dspeck1 (approved):
LGTM

PR #3363, opened 2024-05-22
Add default resource limits/requests to Phalanx core apps

athornton:
For our core applications, add reasonably-sane (but not tight) requests/limits for all of them.

Site-specific overrides to follow eventually.

### DIFF
--- a/applications/strimzi/values.yaml
+++ b/applications/strimzi/values.yaml
@@ -1,0 +1,8 @@
+strimzi-cluster-operator:
+  resources:
+    limits:
+      cpu: "1"
+      memory: "1Gi"
+    requests:
+      cpu: "20m"
+      memory: "450Mi"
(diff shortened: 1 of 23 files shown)


applications/strimzi/values.yaml
```diff
@@ -0,0 +1,8 @@
+strimzi-cluster-operator:
```
rra, line 1: Missing documentation.

rra (commented):
There should be a blank line before the `# --` line everywhere for consistency. Please also add:

```
```
before every resource setting so that we don't blow up the generated Helm values documentation with the serialized dictionary.

This is missing several Argo CD components (I noticed at least the applicationset controller and the dex server). Were you going to go back and add those later?

I suspect you will need to lower these settings for the minikube environment.

athornton:
I think I've done all that, and made minikube resources empty for all the minikube-deployed apps.

PR #3369, opened 2024-05-24
DM-44537: Enable minimum token lifetime for Nublado

rra:
Use the Gafaelfawr functionality we added some time back and require a minimum token lifetime when spawning a lab. This lifetime can be explicitly configured or taken from the maxAge lab culler setting.

Change the culler defaults to match the settings we're using for most environments so that the default maxAge is not longer than the default Gafaelfawr token lifetime.


athornton (approved):
Looks good, although the reauthentication feels like it might turn out to be obnoxious in practice.

athornton:
Do we have any safeguards such that if a site has the cull timeouts very long (but not infinite), and the token the user gets doesn't live that long, we don't get stuck in a reauthentication loop?

rra:
Gafaelfawr rejects such requests with an error.

PR #3373, opened 2024-05-27
fix for nublado sync with CC-IN2P3 branch

gabrimaine:


applications/nublado/values-ccin2p3.yaml
```diff
@@ -64,11 +69,22 @@ controller:
         - containerPath: "/sps"
           volumeName: "sps"
 jupyterhub:
+  prePuller:
+    continuous:
+      enabled: true
```
rra, line 74: This won't do anything of use so far as I can tell. I think it will try to pre-pull a stock JupyterLab image, which Nublado doesn't provide a mechanism to launch, so I think it will just waste disk space without accomplishing anything.

applications/nublado/values-ccin2p3.yaml
```diff
@@ -64,11 +69,22 @@ controller:
...
   hub:
+    baseUrl: "/nb"
+    config:
+      ServerApp:
+        shutdown_no_activity_timeout: 432000
```
rra, line 79: This doesn't do anything any more, which is why we deleted it.

rra (commented):
This is deleting the human-readable comments translating numbers of seconds into something more familiar. I think that may have been unintentional.

applications/nublado/values-ccin2p3.yaml
gabrimaine: OK, I removed useless lines.

rra (approved):

rra:
Looks good, feel free to merge whenever you're ready, although note that you'll need to merge further changes made on main to delete now-unnecessary culler configuration that matches the new defaults.

PR #3379, opened 2024-05-28
tickets/DM-35106: add requests/limits for RSP/Roundtable/Monitoring-minus-sasquatch apps

athornton:


applications/nublado/values.yaml
```diff
@@ -348,6 +348,16 @@ hub:
   # controls how the database password is managed.
   internalDatabase: true
 
+  # -- Resource limits and requests for the Hub
+  # @default -- See `values.yaml`
```
athornton, line 352: I don't understand what I did wrong here such that I'm getting the actual resources rather than "See `values.yaml`" in the `README.md`.

PR #3392, opened 2024-05-30
DM-44461: Mobu notebook refresh

fajpunk:
* Add a `github-webhook-secret` secret to mobu.
  * This has been added to 1password and sync'd for `idfdev`, but nowhere else yet.
* Add an anonymous ingress for mobu to receive GitHub webhooks


applications/mobu/values-idfdev.yaml
```diff
@@ -1,6 +1,21 @@
...
+        - "exec:notebook"
+      business:
+        type: "NotebookRunner"
+        options:
+          repo_url: "https://github.com/lsst-sqre/dfuchs-test-mobu.git"
+          repo_branch: "main"
+        restart: true
     - name: "recommended"
       count: 1
       users:
```
fajpunk, line 21: This is only for testing, I'll ditch it before merging.

applications/mobu/secrets.yaml
```diff
@@ -11,3 +11,6 @@ ALERT_HOOK:
   description: >-
     Slack web hook to which mobu should report failures and daily status.
   if: config.slackAlerts
+GITHUB_WEBHOOK_SECRET:
+  description: >-
+    Github webhook secret for integrated repos
```
rra, line 16: Could you flesh out the description here a little? This wouldn't tell me where that secret is, what I would need to do to change it, etc. In particular, I think it's important to say where in GitHub this secret is also stored, and to note that the same secret is shared by all environments (which I think is the right thing to do).

Also, I know the current secret naming is all over the place, but I'm trying to slowly move things towards the lowercase-with-dashes naming since I think it's a bit more human-readable and makes it clear that the environment variable may not match the secret (because usually we're prepending the application name). So github-webhook-secret here perhaps?

rra (approved):
Only had some comments about the secret naming and description. Everything else looks great; feel free to merge after changing that (well, once we have a new mobu release).

PR #3397, opened 2024-05-31
Enable CloudSQL UWS database for TAP (Initially just for the idfdev SSOTAP service)

stvoutsin:
## PR Summary:

This PR addresses Jira issue DM-44524: Move the UWS database for SSO-TAP to CloudSQL
A summary of the changes introduced to enable the feature:

### Changes to cadc-tap chart:
  - Add optional cloudsql configuration to values
  - Add optional UWS database / username configuration to values
  - Optional serviceaccount.yaml / Used when cloudsql is enabled
  - Make uws optional (add enabled: true | false).
  - Make UWS (uws database that get's deployed as part of cadc-tap until now) resources optional, only deployed when uws enabled
  - Add cloudsql proxy container option to tap-deployment / Deployed if cloudsql is enabled
  - Add Helm unit tests that use: https://github.com/helm-unittest/helm-unittest/
  
### Changes to applications/ssotap:
  - Set serviceAccount in values to default to "ssotap"
  - Enable cloudsql / Disable uws for idfdev. Set parameters so that CloudSQL is used for UWS (using Proxy) / Use tap-postgres image 1.16.0 (which includes auto-initialisation)
  - Add secrets-idfdev.yaml (new secret for UWS CloudSQL database for idfdev)
  
### Changes to github CI workflow:
  - Enable Helm unittest to be run

## Other relevant changes:
- A Postgres database has been created in GCP CloudSQL, named "ssotap" for idfdev
- A secret to store the password for this database has been stored in OnePass   
  
## How was this tested:
- Helm chart unit tests for cadc-tap
- Tested sync & async queries to ssotap with the following scenarios:
   - Empty cloudsql ssotap UWS database
   - Non-empty cloudsql ssotap UWS database
   - cloudsql disabled / uws enabled 
  
## Detailed Description

Detailed description of steps taken, design decisions, as well as some question and potential issues found can be found here:
https://github.com/stvoutsin/rub-notes/wiki/DM%E2%80%9044427-%E2%80%90-Enable-CloudSQL-UWS-database-for-SSO-TAP-service 
  
  
## Other info:
The changes should be backwards compatible, i.e. no changes should be required to any of the other tap or ssotap applications. If there are I've missed something and should correct.

### DIFF
--- a/applications/ssotap/secrets-idfdev.yaml
+++ b/applications/ssotap/secrets-idfdev.yaml
@@ -1,11 +1,3 @@
-"google_creds.json":
-  description: >-
-    Google service account credentials used to write async job output to
-    Google Cloud Storage.
-pgpassword:
-  description: >-
-    Password to external PostgreSQL server that contains the Solar System
-    Objects data.
-uwspassword:
+uws-db-password:
   description: >-
     Password to external UWS PostgreSQL server
--- a/applications/ssotap/secrets-idfdev.yaml
+++ b/applications/ssotap/secrets-idfdev.yaml
@@ -1,0 +1,11 @@
+"google_creds.json":
+  description: >-
+    Google service account credentials used to write async job output to
+    Google Cloud Storage.
+pgpassword:
+  description: >-
+    Password to external PostgreSQL server that contains the Solar System
+    Objects data.
+uwspassword:
+  description: >-
+    Password to external UWS PostgreSQL server
--- a/applications/ssotap/values-idfdev.yaml
+++ b/applications/ssotap/values-idfdev.yaml
@@ -1,4 +1,21 @@
 cadc-tap:
+  config:
+    uwsDatabaseUrl: "postgresql://localhost:5432/ssotap"
+    uwsDatabaseUsername: "ssotap"
+    pg:
+      image:
+        repository: "ghcr.io/lsst-sqre/tap-postgres-service"
+        pullPolicy: "Always"
+        tag: "1.16.0"
+
+  uws:
+    enabled: false
+
   tapSchema:
     image:
       repository: "lsstsqre/tap-schema-idfdev-sso"
+
+  cloudsql:
+    enabled: true
+    instanceConnectionName: "science-platform-dev-7696:us-central1:science-platform-dev-e9e11de2"
+    serviceAccount: "ssotap@science-platform-dev-7696.iam.gserviceaccount.com"
--- a/applications/ssotap/values-idfdev.yaml
+++ b/applications/ssotap/values-idfdev.yaml
@@ -1,15 +1,10 @@
 cadc-tap:
   config:
-    uwsDatabaseUrl: "postgresql://localhost:5432/ssotap"
-    uwsDatabaseUsername: "ssotap"
     pg:
       image:
         repository: "ghcr.io/lsst-sqre/tap-postgres-service"
         pullPolicy: "Always"
         tag: "1.16.0"
-
-  uws:
-    enabled: false
 
   tapSchema:
     image:
@@ -19,3 +14,4 @@ cadc-tap:
     enabled: true
     instanceConnectionName: "science-platform-dev-7696:us-central1:science-platform-dev-e9e11de2"
     serviceAccount: "ssotap@science-platform-dev-7696.iam.gserviceaccount.com"
+    database: "ssotap"
--- a/charts/cadc-tap/templates/serviceaccount.yaml
+++ b/charts/cadc-tap/templates/serviceaccount.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: {{ .Values.serviceAccount.name }}
+  name: {{ required "serviceAccount.name must be set" .Values.serviceAccount.name | quote }}
   labels:
     {{- include "cadc-tap.labels" . | nindent 4 }}
   annotations:
--- a/charts/cadc-tap/templates/serviceaccount.yaml
+++ b/charts/cadc-tap/templates/serviceaccount.yaml
@@ -1,0 +1,15 @@
+{{- if or .Values.serviceAccount.create .Values.cloudsql.enabled -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Values.serviceAccount.name }}
+  labels:
+    {{- include "cadc-tap.labels" . | nindent 4 }}
+  annotations:
+    {{- if .Values.cloudsql.enabled }}
+    iam.gke.io/gcp-service-account: {{ required "cloudsql.serviceAccount must be set to a valid Google service account" .Values.cloudsql.serviceAccount | quote }}
+    {{- end }}
+    {{- with .Values.serviceAccount.annotations }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+{{- end }}
--- a/charts/cadc-tap/templates/tap-deployment.yaml
+++ b/charts/cadc-tap/templates/tap-deployment.yaml
@@ -72,7 +72,7 @@ spec:
                 {{- if .Values.cloudsql.enabled }}
                 -Duws.password=$UWS_DB_PASSWORD
                 -Duws.username={{ required "cloudsql.database must be specified" .Values.cloudsql.database }}
-                -Duws.url=jdbc:postgres://localhost:5432/{{ .Values.cloudsql.database }}
+                -Duws.url=jdbc:postgresql://localhost:5432/{{ .Values.cloudsql.database }}
                 {{- else }}
                 -Duws.username=postgres
                 -Duws.url=jdbc:postgresql://{{ template "cadc-tap.fullname" . }}-uws-db/
--- a/charts/cadc-tap/templates/tap-deployment.yaml
+++ b/charts/cadc-tap/templates/tap-deployment.yaml
@@ -70,7 +70,7 @@ spec:
                 -Duws.driverClassName=org.postgresql.Driver
                 -Duws.maxActive=100
                 {{- if .Values.cloudsql.enabled }}
-                -Duws.password=$uwspassword
+                -Duws.password=$UWS_DB_PASSWORD
                 -Duws.username={{ required "cloudsql.database must be specified" .Values.cloudsql.database }}
                 -Duws.url=jdbc:postgres://localhost:5432/{{ .Values.cloudsql.database }}
                 {{- else }}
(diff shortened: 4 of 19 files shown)


athornton:
So first thing, that's a fantastic PR summary and stop making me look as lazy as I actually am, please.

charts/cadc-tap/values.yaml
```diff
@@ -158,14 +164,7 @@ tapSchema:
...
   # -- Resource limits and requests for the TAP schema database pod
-  # @default -- See `values.yaml`
-  resources:
-    limits:
-      cpu: "1"
-      memory: "1Gi"
-    requests:
-      cpu: "5m"
-      memory: "400Mi"
+  resources: {}
```
athornton, line 167: Please don't remove the resources.  If these are not the right default values then fix them.

athornton:
I like the helm unit-testing sorcery.

athornton:
One other thing: is there https://github.com/lsst/idf_deploy work that has to go along with the ssotap SA to get WI functioning?

applications/ssotap/secrets-idfdev.yaml
```diff
@@ -0,0 +1,11 @@
+"google_creds.json":
+  description: >-
+    Google service account credentials used to write async job output to
+    Google Cloud Storage.
+pgpassword:
+  description: >-
+    Password to external PostgreSQL server that contains the Solar System
+    Objects data.
```
rra, line 8: `secrets.yaml` and `secrets-<environment>.yaml` files are merged, so you don't need to duplicate the values from `secrets.yaml` here. You only have to add the new one.

applications/ssotap/secrets-idfdev.yaml
```diff
@@ -0,0 +1,11 @@
...
+uwspassword:
+  description: >-
+    Password to external UWS PostgreSQL server
```
rra, line 11: The existing naming scheme for secrets is all over the map, but I'm trying to standardize on dash-separated names since they're a bit more human-readable. How about `uws-db-password` for this? (Or `uws-database-password` if you want to be more verbose.)

applications/ssotap/values-idfdev.yaml
```diff
@@ -1,4 +1,21 @@
 cadc-tap:
+  config:
+    uwsDatabaseUrl: "postgresql://localhost:5432/ssotap"
```
rra, line 3: If CloudSQL is in use, the only part of this that really is configurable is the database name. I wonder if it would make sense to say that the database name and user will always be the same for CloudSQL and move that, as a single setting, into the CloudSQL section, and then just set the database URL to this if CloudSQL is enabled. That would simplify the interface here a bit.

applications/ssotap/values-idfdev.yaml
```diff
@@ -1,4 +1,21 @@
...
+    uwsDatabaseUsername: "ssotap"
+    pg:
+      image:
+        repository: "ghcr.io/lsst-sqre/tap-postgres-service"
+        pullPolicy: "Always"
+        tag: "1.16.0"
+
+  uws:
+    enabled: false
+
```
rra, line 13: Similarly here, we can assume that if CloudSQL is enabled, UWS is always disabled.

charts/cadc-tap/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if or .Values.serviceAccount.create .Values.cloudsql.enabled -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Values.serviceAccount.name }}
```
rra, line 5: I don't think there's a great default value for the service account name, so I would be inclined to do the same `required` thing here. So:
```suggestion
  name: {{ required "serviceAccount.name must be set" .Values.serviceAccount.name | quote }}
```

charts/cadc-tap/templates/tap-deployment.yaml
```diff
@@ -39,18 +67,23 @@ spec:
...
                 -Dtapschemauser.url=jdbc:mysql://{{ .Values.config.tapSchemaAddress }}/
                 -Dtapschemauser.maxActive=100
-                -Duws.username=postgres
                 -Duws.driverClassName=org.postgresql.Driver
-                -Duws.url=jdbc:postgresql://{{ template "cadc-tap.fullname" . }}-uws-db/
                 -Duws.maxActive=100
+                {{- if .Values.cloudsql.enabled }}
+                -Duws.password=$uwspassword
+                -Duws.username={{ .Values.config.uwsDatabaseUsername }}
+                -Duws.url=jdbc:{{ .Values.config.uwsDatabaseUrl }}
```
rra, line 75: If my suggestion above sounded good, this would then be something like:
```suggestion
                -Duws.username={{ .Values.config.cloudsql.database }}
                -Duws.url=jdbc:postgres://localhost:5432/{{ .Values.config.cloudsql.database }}
```

charts/cadc-tap/templates/tap-deployment.yaml
```diff
@@ -74,6 +107,13 @@ spec:
                   name: "cadc-tap"
                   key: "pgpassword"
             {{- end }}
+            {{- if .Values.cloudsql.enabled }}
+            - name: "uwspassword"
+              valueFrom:
+                secretKeyRef:
+                  name: "cadc-tap"
+                  key: "uwspassword"
```
rra, line 115: This would change to `uws-db-password` if my previous suggestion sounded good. The internal `$uwspassword` environment variable is fine and doesn't need to change, although I probably would pedantically change the environment variables used in this chart to use all-caps names like `TAP_UWS_DB_PASSWORD` and `TAP_DB_PASSWORD` (for `pgpassword`). But this is just me being perfectionist and you can ignore that. :)

charts/cadc-tap/templates/uws-db-deployment.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if .Values.uws.enabled -}}
```
rra, line 1: These could then be
```suggestion
{{ if (not .Values.cloudsql.enabled) -}}
```
just so that there's one less configuration setting to keep in sync.

charts/cadc-tap/templates/uws-db-networkpolicy.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if .Values.mockdb.enabled -}}
```
rra, line 1: I don't think you meant mockdb here.

charts/cadc-tap/values.yaml
```diff
@@ -212,6 +213,51 @@ uws:
...
+  resources:
+    limits:
+      cpu: "100m"
+      memory: "20Mi"
+    requests:
+      cpu: "5m"
+      memory: "7Mi"
+
+serviceAccount:
+  # -- Specifies whether a service account should be created.
```
rra, line 251: It's worth adding to the documentation here that a service account is always created if CloudSQL is enabled.

charts/cadc-tap/values.yaml
```diff
@@ -212,6 +213,51 @@ uws:
...
+  create: false
+
+  # -- Annotations to add to the service account
+  annotations: {}
+
+  # The name of the service account to use.
+  # @default -- Generated using the fullname template
+  name: "cadc-tap"
```
rra, line 259: It looks like this `@default` note is incorrect; I'm guessing you tried doing that and it didn't work? It would be great if we could have it pick up the deployment name that way, but passing in the service account name explicitly is fine. I would set this to null and say something like `@default -- None, must be set if service account creation is enabled`, though, as noted above.

rra (commented):
I agree with Adam: fantastic PR summary. The Helm tests look great; I am really looking forward to trying that out for my services.

rra:
The idf_deploy changes were already done in https://github.com/lsst/idf_deploy/pull/523.

charts/cadc-tap/values.yaml
stvoutsin: Didn't mean to remove those, they should be back now thank for noticing

applications/ssotap/secrets-idfdev.yaml
stvoutsin: Ok I see, I've modified this accordingly

applications/ssotap/secrets-idfdev.yaml
stvoutsin: Sure that makes sense. I've changed it to uws-db-password. Was wondering if I should change "pgpassword" from the ssotap secrets while I'm here, but maybe I'll leave that for now

applications/ssotap/values-idfdev.yaml
stvoutsin: I think that makes sense. One reason i had those as separate from the cloudsql configuration, was that I was considering the possibility that someone will want to use the chart with an external UWS database that is not on cloudsql. 

Having said that my proposed changes do not quite allow that, as these were used only when cloudsql is enabled. So I've gone with your suggestion for now, and if we want to add this capability we can reconsider which structure makes it easier and more transparent.

applications/ssotap/values-idfdev.yaml
stvoutsin: Yep, removed this extra param. Should make the values more concise

charts/cadc-tap/templates/serviceaccount.yaml
stvoutsin: Okay i've made this required

charts/cadc-tap/templates/tap-deployment.yaml
stvoutsin: I agree and gone with this for now. Perhaps later if we want to open the door for external UWS that is not on cloudsql, we can revisit

charts/cadc-tap/templates/tap-deployment.yaml
stvoutsin: Sure all that makes sense, changes applied (uws-db-password and the all-caps env names)

charts/cadc-tap/templates/uws-db-networkpolicy.yaml
stvoutsin: Oops, thought I had fixed this one, good catch thank you

charts/cadc-tap/values.yaml
stvoutsin: For this one there may be a better solution that I was missing. Basically I wanted the serviceaccount template which is in the cadc-tap chart to end up with the name of the application that is using the cadc-tap as a dependency, e.g. "ssotap" for the ssotap application. Didn't find a better solution for this, so ended up just having it passed in as a name value like this. (The default note is indeed incorrect and I've changed that)

stvoutsin:
> One other thing: is there https://github.com/lsst/idf_deploy work that has to go along with the ssotap SA to get WI functioning? EDIT the linked notes say not, which feels weird to me. I mean, clearly the service is functioning, but I don't understand why there isn't a need for WI bindings over in the idf_deploy stuff.

I think I need to correct the links, as I may have missread that part of the documentation. I did create a SA in using Terraform, which I assume is all that is needed, as that get's passed in and used in the templates to connect via the proxy.

rra (approved):
Looks great. Thank you!

PR #3398, opened 2024-05-31
DM-44544: Add ir2 repo to idfdev Butler server

dhirving:
We want to use the ir2 repo to test the "hybrid model" where the Butler's database is at Google but the data is at USDF.

Added a configuration option to Butler to allow it to use two different sets of S3 credentials simultaneously.  Added the ir2 repo to data-dev.

### DIFF
--- a/applications/butler/secrets.yaml
+++ b/applications/butler/secrets.yaml
@@ -18,3 +18,9 @@
   copy:
     application: nublado
     key: "postgres-credentials.txt"
+"additional-s3-profile":
+  description: >-
+    Credentials and endpoint for a second S3 profile to use, in addition to the
+    default endpoint.  For docs on format see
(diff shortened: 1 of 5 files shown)


applications/butler/secrets.yaml
```diff
@@ -18,3 +18,9 @@
   copy:
     application: nublado
     key: "postgres-credentials.txt"
+"additional-s3-profile":
+  description: >-
+    Credentials and endpoint for a second S3 profile to use, in addition to the
+    default endpoint.  For docs on format see
+    https://github.com/lsst/resources/blob/a34598e125919799d3db4bd8a2363087c3de434e/python/lsst/resources/s3utils.py#L201
```
athornton, line 25: Gross URL.

athornton (approved):
Looks OK.

PR #3409, opened 2024-06-05
tickets/PREOPS-5192: Update phalanx config for schedview-snapshot to use its own vault secret

ehneilsen:
There were a few "policy violation" events that got shown by ArgoCD on usdf-int that commits here address (or attempt to address) as well. See the slack thread [here](https://lsstc.slack.com/archives/C027902HL8H/p1717442595254499).

### DIFF
--- a/applications/schedview-snapshot/templates/deployment.yaml
+++ b/applications/schedview-snapshot/templates/deployment.yaml
@@ -29,7 +29,6 @@ spec:
               drop:
                 - "all"
             readOnlyRootFilesystem: true
-          privileged: false
           securityContext:
             privileged: false
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
--- a/applications/schedview-snapshot/templates/deployment.yaml
+++ b/applications/schedview-snapshot/templates/deployment.yaml
@@ -29,6 +29,7 @@ spec:
               drop:
                 - "all"
             readOnlyRootFilesystem: true
+          privileged: false
           securityContext:
             privileged: false
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
--- a/applications/schedview-snapshot/templates/deployment.yaml
+++ b/applications/schedview-snapshot/templates/deployment.yaml
@@ -29,7 +29,6 @@ spec:
               drop:
                 - "all"
             readOnlyRootFilesystem: true
-          securityContext:
             privileged: false
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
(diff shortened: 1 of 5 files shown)


applications/schedview-snapshot/templates/deployment.yaml
```diff
@@ -29,6 +29,7 @@ spec:
               drop:
                 - "all"
             readOnlyRootFilesystem: true
+          privileged: false
```
rra, line 32: This isn't a valid setting at this level of a container specification (it would go under `securityContext`, so I think Argo CD will either reject this manifest or will silently delete this setting. I'm not sure what this addition was intended to do, though. Containers are unprivileged by default so if this were under `securityContext` it would just be reasserting the default and would have no effect.

applications/schedview-snapshot/values-usdfint.yaml
```diff
@@ -1,3 +1,11 @@
...
   # -- Overrides the image tag whose default is the chart appVersion.
   tag: "v0.12.0"
+
+resources:
+  limits:
+    cpu: 2.0
+    memory: "8Gi"
+  requests:
+    cpu: 1.0
+    memory: "4Gi"
```
rra, line 11: In general I recommend putting this into `values.yaml` instead, since resource consumption is usually fairly similar across environments, and only overriding it in the per-environments file if there's something unusual about that environment.

rra (approved):
There's one syntax error that should be fixed before merging (I think you can just delete that `privileged` line).

applications/schedview-snapshot/templates/deployment.yaml
ehneilsen: When I didn't have it there, I was getting the following "event" from ArgoCD:

> policy disallow-privileged-containers/privileged-containers fail: validation error: Privileged mode is disallowed. The fields spec.containers[*].securityContext.privileged and spec.initContainers[*].securityContext.privileged must be unset or set to `false`. . rule privileged-containers failed at path /spec/

Originally, I didn't have a privileged keyword at all. When I added it as I did, it made the event go away. But, I did indeed miss the `securityContext` element in the complaint. I'm not clear on why what I did worked, or why it was necessary at all. I'll go ahead and move it under `securityContext` and check whether the complaint comes back.

ehneilsen: When I move it under `securityContext`, the event reappears.

ehneilsen: When I put it both next to and under `securityContext`, the event goes away again.

ehneilsen: When I take it out from the place where it should **not** go (next to, not under, `securityContext`), a second policy complaint also shows up:
`policy disallow-host-path/host-path fail: validation error: HostPath volumes are forbidden. The field spec.volumes[*].hostPath must be unset. . rule host-path failed at path /spec/`
I don't think I'm using hostpath volumes anywhere (and find/grep doesn't find any).
Maybe there's a bug in the policy configuration for usdf-inf somewhere?

PR #3410, opened 2024-06-06
DM 44461: Mobu GitHub resource/secret toggle

fajpunk:
## Changes
Taking some cues from @stvoutsin with this PR description and helm unit tests :grinning: 

### Mobu
* Add mobu chart value (`config.githubIntegration`) to conditionally provision resources and secrets to support environments that are behind a VPN and can't get github webhook requests.

### Repo-wide
* Use latest version of helm unittest plugin ([the repo moved](nittest/blob/254c582c18d52e07541dbe0185eb64befa20a822/README.md?plain=1#L7))
* Run `helm unittest` in all `charts/*` and `applications/*` dirs with tests.

## Testing
* Set `config.githubIntegration` to `false` for mobu in `idfdev`
  * Verified ingress was gone and secrets were not injected into container env
 * New helm unit tests

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -64,7 +64,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Install Helm Unittest Plugin
-        run: helm plugin install https://github.com/quintush/helm-unittest
+        run: helm plugin install https://github.com/helm-unittest/helm-unittest
 
       - name: Run Helm Unittest
         run: helm unittest charts/cadc-tap
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -66,9 +66,8 @@ jobs:
       - name: Install Helm Unittest Plugin
         run: helm plugin install https://github.com/helm-unittest/helm-unittest
 
-      - name: Run Helm Unittest
-        run: helm unittest charts/cadc-tap
-
+      - name: Run Helm Unittest for all applications and charts with a `tests` directory
+        run: helm unittest {applications,charts}/*/tests/..
       - uses: lsst-sqre/run-tox@v1
         with:
           python-version: "3.12"
--- a/applications/mobu/README.md
+++ b/applications/mobu/README.md
@@ -13,6 +13,7 @@ Continuous integration testing
 | affinity | object | `{}` | Affinity rules for the mobu frontend pod |
 | config.autostart | list | `[]` | Autostart specification. Must be a list of mobu flock specifications. Each flock listed will be automatically started when mobu is started. |
 | config.debug | bool | `false` | If set to true, include the output from all flocks in the main mobu log and disable structured JSON logging. |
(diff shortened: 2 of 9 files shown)


.github/workflows/ci.yaml
```diff
@@ -66,8 +66,9 @@ jobs:
       - name: Install Helm Unittest Plugin
         run: helm plugin install https://github.com/quintush/helm-unittest
 
+      # Runs all tests/*_test.yaml files by default
       - name: Run Helm Unittest
-        run: helm unittest charts/cadc-tap
+        run: helm unittest charts/*/ applications/*/
```
fajpunk, line 71: @stvoutsin, any reasons you can think of why we shouldn't do this?

fajpunk: I scoped it down to just the dirs with `tests/` subdirs, that makes the output  cleaner.

.github/workflows/ci.yaml
```diff
@@ -64,10 +64,12 @@ jobs:
...
       - name: Install Helm Unittest Plugin
-        run: helm plugin install https://github.com/quintush/helm-unittest
+        run: helm plugin install https://github.com/helm-unittest/helm-unittest
 
-      - name: Run Helm Unittest
-        run: helm unittest charts/cadc-tap
+      - name: Run Helm Unittest for all applications and charts with a `tests` directory
+        run: |
+          dirs=$(find {applications,charts} -maxdepth 2 -mindepth 2 -type d -name tests -exec dirname {} \;)
+          helm unittest $dirs
```
rra, line 72:
```suggestion
          helm unittest {applications,charts}/*/tests
```
will work now that you're adding tests for mobu so there's at least one application with tests and the first wildcard will be expanded, I think.

rra (approved):
This looks great. Thank you!

.github/workflows/ci.yaml
fajpunk: That doesn't work because the dir with `Chat.yaml` is the dir that needs to passed to `helm unittest`, which is the _parent_ of the tests directory.  There very well may be some `bash` wizardry to get this to work that's nicer than my pages-long find command.

Another option is to pass all of the application and chart dirs with `helm unittest {applications,charts}/*/`, but that has output for all dirs, even if they don't have subdirs, and the results for individual dirs are not moved to the bottom :(

fajpunk: ...but `helm unittest {applications,charts}/*/tests/..` works!

.github/workflows/ci.yaml
stvoutsin: Yes agree with this, great stuff thanks

.github/workflows/ci.yaml
rra: Oh, that's what I missed! How annoying. That's a nice workaround; I probably would have gone for:
```
    helm unittest $(echo {applications,charts}/*/tests} | sed 's%/tests%%g')
 ```
 but that's still pretty ugly and the find is honestly probably better than that.

PR #3423, opened 2024-06-11
DM-44523: run cm-service under k8s

fritzm:
Updates to cm-service chart
* Add redis and cnpg postgress to the app
* Add secrets management
* Move env vars from configmap into deployment
* Remove gafaelfawr ingress for time being (until gf gets deployed to usdf cm vclusters)
* Add volume configuration support for service output dir
* Adjust to run service as user `lsstsvc1`
* Add condor-capable worker deployment
* Move deployment from `idfdemo` to `usdf-cm`


applications/cm-service/values.yaml
```diff
@@ -16,36 +36,112 @@ image:
...
+  # -- Annotations for the worker pod
+  podAnnotations: {}
+
+  # -- Tolerations for the worker pod
+  tolerations: []
+
+  htcondor:
+    config:
+      # -- If specified, location for htcondor config file to be injected into
+      #    worker pod
```
ctslater, line 109: These hanging indents puts a weird whitespace in the auto-generated README.

applications/cm-service/templates/pvc.yaml
```diff
@@ -0,0 +1,34 @@
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: cm-service-output
+  labels:
+    {{- include "cm-service.labels" . | nindent 4 }}
+spec:
+  {{- with .Values.config.outputVolume.storageClassName }}
```
ctslater, line 9: What's the advantage of doing a `with` instead of just `storageClassName: {{ .Values.config.outputVolume.storageClassName }}` ?

applications/cm-service/templates/deployment.yaml
```diff
@@ -40,22 +61,32 @@ spec:
...
-      {{- with .Values.tolerations }}
+      {{- with .Values.frontend.tolerations }}
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       securityContext:
         runAsNonRoot: true
-        runAsUser: 1000
-        runAsGroup: 1000
+        runAsUser: 17951
```
ctslater, line 91: Break out to values file?

applications/cm-service/values-usdf-cm.yaml
```diff
@@ -0,0 +1,31 @@
...
+  logLevel: "INFO"
+  logProfile: "development"
+  databaseEcho: true
+  outputVolume:
+    storageClassName: "sdf-data-rubin"
+    subPath: "shared/campaigns/users/usdf-cm-prod"
+worker:
+  htcondor:
+    config:
+      mountPath: "/opt/lsst/software/stack/conda/envs/lsst-scipipe-9.0.0/etc/condor/config.d"
```
ctslater, line 11: Suspicious looking path; is rubin-env actually mounted there?

applications/cm-service/values.yaml
```diff
@@ -2,8 +2,28 @@
...
+  outputVolume:
+    # -- If specified, name of storage class requested in service output area PVC
+    storageClassName: null
+    # -- Minimum storage requested in service output area PVC
+    storage: 1Gi
+    # -- If specified, sub-path within bound PV to be mounted at service output area
+    subPath: null
+
+  # -- URL path prefix
+  pathPrefix: "/cm-service/v1"
```
ctslater, line 26: I think we're going to want the v1 to be internal to the container so that the GUI can live at /cm-service/.

ctslater (approved):
Just a few minor thoughts added, all looks good.

applications/cm-service/templates/pvc.yaml
fritzm: The body of the `with` is not expanded if the variable is `null`, so you don't get a dangling/empty yaml key this way.

applications/cm-service/templates/deployment.yaml
fritzm: Good point -- USDF (or other site) specific -- will do.

applications/cm-service/values-usdf-cm.yaml
fritzm: Yup. This nests our condor config within the config defaults that come along with the stack.  Otherwise, setting up the stack in the container wipes out our config when it repoints `CONDOR_CONFIG`.

applications/cm-service/values.yaml
fritzm: Agreed. I will coordinate with Eman to rotate the paths on a subsequent ticket.

applications/cm-service/values.yaml
fritzm: Hunh -- I copied that style from some other applications in phalanx.  I see what you are saying, though.  They don't show up when rendered in-browser at `lsst.io` because of the vagaries of HTML so I missed them!

applications/cm-service/templates/deployment.yaml
fritzm: Actually, upon inspection, this user/group is already specified as the default in the container image.  I will just remove these.

PR #3427, opened 2024-06-11
DM-44783: Prepare for OR4 Prompt Processing

hsinfang:


applications/prompt-proto-service-lsstcomcamsim/values-usdfprod-prompt-processing.yaml
```diff
@@ -10,30 +10,33 @@ prompt-proto-service:
...
+    endpointUrl: https://sdfembs3.sdf.slac.stanford.edu
 
   imageNotifications:
     kafkaClusterAddress: prompt-processing-2-kafka-bootstrap.kafka:9092
-    topic: rubin-prompt-processing-prod
+    topic: rubin-summit-notification
 
   apdb:
-    config: s3://rubin-summit-users/apdb_config/sql/pp_apdb_lsstcomcamsim.py
+    config: s3://rubin-summit-users/apdb_config/cassandra/pp_apdb_lsstcomcamsim_pre_or4.py
```
isullivan, line 36: Will you create a new config and keyspace without "pre" before OR4 begins?

applications/prompt-proto-service-lsstcomcamsim/values-usdfprod-prompt-processing.yaml
```diff
@@ -10,30 +10,33 @@ prompt-proto-service:
...
+
+  alerts:
+    topic: alerts-simulated
```
isullivan, line 39: Just checking that this should not also be `alert-stream-test` as in dev.

isullivan (approved):
Looks good! Just a couple things for you to double-check.

applications/prompt-proto-service-lsstcomcamsim/values-usdfprod-prompt-processing.yaml
hsinfang: I believe that it has been created. And we need to remember to switch to it right before OR4 begins.

applications/prompt-proto-service-lsstcomcamsim/values-usdfprod-prompt-processing.yaml
hsinfang: That's my understanding from @bsmartradio  from  https://lsstc.slack.com/archives/C0345PUU4N9/p1718147016397699

applications/prompt-proto-service-lsstcomcamsim/values-usdfprod-prompt-processing.yaml
isullivan: I can confirm that `pp_apdb_lsstcomcamsim_or4` has been created, and is what we should use for OR4. I agree we should wait to switch until just before the rehearsal, in case they send more test data.

PR #3430, opened 2024-06-13
DM-40193: Disk space leak in Prompt Processing

kfindeisen:
This PR adds service configuration options for the caching system introduced in lsst-dm/prompt_processing#169, and cuts ephemeral storage requests to a quarter.


erinleighh (approved):
Everything looks good!

PR #3434, opened 2024-06-14
DM-42924: Test Prompt Processing with Cassandra APDB

kfindeisen:
This PR updates the LATISS-dev and ComCamSim-dev instances of Prompt Processing to use Cassandra-backed APDBs. HSC-dev is left on SQL for test coverage, but the URI has been updated to reflect the config files' reorganization. We are not deploying Cassandra in production just yet.


isullivan (approved):
Looks good, I assume you have updated the configs in the s3 bucket to match.

kfindeisen:
Other way around -- these updates are needed to fit the contents of the bucket.

PR #3435, opened 2024-06-16
Make configured secret name match resource

athornton:


rra (approved):
Thanks, sorry about that, It was hidden in the `values.yaml` file instead of in `templates` where I was searching, so I forgot about it, and then the order with which I did the deployment hid the problem until the Redis server was restarted again.

PR #3444, opened 2024-06-19
Upgrade TAP images / Setup TAP uws on CloudSQL / New cadc libs

stvoutsin:
**Description**
This PR addresses the tasks of creating & enabling the use of CloudSQL (Postgres) as the backend database for UWS, for both the tap & ssotap applications. It also upgrades the versions of the lsst-tap-service & tap-postgres images, to newer ones which include upgrades to the cadc libraries, enabling automated UWS creation through the application on initialisation of the tomcat app & switching to using cadc-tap tomcat as the base image.
This change for the cadc upgrades also come with the use of a configmap to pass in some properties, which is mounted on /config which the cadc-tap expects to find for the properties.

**Backwards Compatibility**
Changes should not require any changes from the users of phalanx. Changes to the idf values for tap & ssotap are included in PR

**Related Jira Issues**
- https://rubinobs.atlassian.net/browse/DM-44524
- https://rubinobs.atlassian.net/browse/DM-44607
- https://rubinobs.atlassian.net/browse/DM-44746

**How was this tested**
- Manual validation (sync & async queries to data & TAP_SCHEMA via pyvo & firefly portal)
- taplint validator
- helm unittest

**Related changes (Not in PR)**
- Changed idf_deploy to enable creation of ssotap & tap CloudSQL databases on int & dev
- Updated lsst-tap-service & tap-postgres (Upgrading cadc-tap libs & enabling automated uws table create, switch to using cadc-tap tomcat base image)
- phalanx secrets sync on int & dev (PENDING)

**Notes for review**
- We have different setups for phalanx installations (cloudsql, non-cloudsql uws, qserv / pg databases for data). Is this likely to break anything for other envs that I haven't tested?


charts/cadc-tap/templates/tap-deployment.yaml
```diff
@@ -82,23 +82,20 @@ spec:
...
-                -Dca.nrc.cadc.auth.Authenticator=org.opencadc.tap.impl.AuthenticatorImpl
                 {{- end }}
-                -Dca.nrc.cadc.reg.client.RegistryClient.local=true
                 -Dgafaelfawr_url={{ .Values.global.baseUrl }}/auth/api/v1/user-info
                 -Dgcs_bucket={{ .Values.config.gcsBucket }}
                 -Dgcs_bucket_url={{ .Values.config.gcsBucketUrl }}
                 -Dgcs_bucket_type={{ .Values.config.gcsBucketType }}
                 -Dbase_url={{ .Values.global.baseUrl }}
-                -Dca.nrc.cadc.util.PropertiesReader.dir=/etc/creds/
+                -Dca.nrc.cadc.util.PropertiesReader.dir=/config/
```
athornton, line 98: I dislike creating new top-level directories.  But that's probably just me being a Unix dinosaur.

charts/cadc-tap/templates/tap-deployment.yaml
```diff
@@ -142,6 +139,8 @@ spec:
               readOnly: true
             - name: "tmp"
               mountPath: "/tmp"
+            - name: "config-volume"
+              mountPath: "/config"
```
athornton, line 143: As above.  `/etc/tap/config` maybe?

applications/ssotap/secrets-idfint.yaml
```diff
@@ -0,0 +1,3 @@
+uws-db-password:
+  description: >-
+    Password to external UWS PostgreSQL server
```
athornton, line 3: "Password for" would sound a bit better to my ears, here and elsewhere.

athornton:
I don't know whether this is likely to break other environments.  It doesn't particularly look dangerous to me, but I also have effectively no domain knowledge here.

Remember to validate that memory requests/limits are still adequate, since we're changing images, and I don't have any feel as to how drastic those changes are.

athornton (approved):
A couple minor naming and location things, which if you choose to ignore would be fine.

The one thing that might not be: if you don't already have credentials for usdf-dev, you should probably ask Peter for some and test on usdf-dev (if that wouldn't be spectacularly disruptive but ehhhh I mean it's dev...).  I'm slightly worried that since you've changed the qserv image and USDF has qserv parameters, there's room for breakage.

charts/cadc-tap/templates/tap-deployment.yaml
stvoutsin: I did try a different directory which doesn't seem to work. I think because the cadc TAP service expects to find the config there. It may be possible to convince it to use a different directory, but not sure how complicated it will be.
Any objections to going with this for now, and perhaps revisit later as a separate task?

applications/ssotap/secrets-idfint.yaml
stvoutsin: Yep sounds right, I've changes it to match your suggestion

PR #3453, opened 2024-06-24
DM-44635: Mobu in GitHub CI

fajpunk:
* Config to support Mobu CI and Mobu Refresh GitHub apps (https://github.com/lsst-sqre/mobu/pull/352)
* Enabling both apps in the `idfdev` env

## Testing
* No github integrations enabled
* Both integrations enabled
* Only refresh integration enabled
* Only CI integration enabled


applications/mobu/README.md
```diff
@@ -13,7 +13,9 @@ Continuous integration testing
 | affinity | object | `{}` | Affinity rules for the mobu frontend pod |
 | config.autostart | list | `[]` | Autostart specification. Must be a list of mobu flock specifications. Each flock listed will be automatically started when mobu is started. |
 | config.debug | bool | `false` | If set to true, include the output from all flocks in the main mobu log and disable structured JSON logging. |
-| config.githubIntegration | bool | `true` | Whether to enable the GitHub app integration. This should be `false` in envs that are behind a VPN because the integration depends on being able to receive webhook requests from github.com. |
+| config.github | object | `{}` | GitHub app integration options. |
```
athornton, line 16: Put an @default in values.yaml for this so we don't have an inlined object; sure, since in this case it's empty it's no big deal, but it's good to get into the habit.

applications/mobu/templates/deployment.yaml
```diff
@@ -82,23 +111,33 @@ spec:
...
         runAsNonRoot: true
         runAsUser: 1000
         runAsGroup: 1000
       volumes:
-        {{- if .Values.config.autostart }}
-        - name: "autostart"
-          configMap:
-            name: {{ include "mobu.fullname" . }}-autostart
-        {{- end }}
+        - name: "config"
```
athornton, line 124: Am I reading this right, that we're collapsing the two configmaps into one?

The only reason we might not want to do this is that on-the-fly editing of the autostart configmap is reasonably common (to modify or disable flocks) so this might be a little more dangerous.

applications/mobu/values.yaml
```diff
@@ -26,6 +26,9 @@ config:
   # Each flock listed will be automatically started when mobu is started.
   autostart: []
 
+  # -- GitHub app integration options.
+  github: {}
```
athornton, line 30: Add "# @default -- See `values.yaml`" here to prevent injecting the object into the docs.

athornton (approved):
A couple minor things.  One is just a documentation thing that you will learn that Russ cares a lot about, and the other is a slight concern about collapsing the two configmaps into one.

applications/mobu/templates/deployment.yaml
fajpunk: Nope, this is [mounting the two separate ConfigMaps as two separate files](https://kubernetes.io/docs/concepts/storage/projected-volumes/) in to the same `/etc/mobu` directory:

```
appuser@mobu-6f947f55cc-jrrfk:/$ ls -l /etc/mobu/
total 0
lrwxrwxrwx 1 root root 21 Jun 25 15:39 autostart.yaml -> ..data/autostart.yaml
lrwxrwxrwx 1 root root 25 Jun 25 15:39 github-config.yaml -> ..data/github-config.yaml
appuser@mobu-6f947f55cc-jrrfk:/$
```
The autostart CM can still be manually edited independently, and the changes do make it into the running mobu pods (though we'd have to create new mobu pods to pick up the changes anyway, right?)

PR #3483, opened 2024-07-03
TAP async results redirect fix (TAP Image upgrades)

stvoutsin:
**Description**

Changes how TAP async results are handled. Introduce a results servlet, so that the uws/results/result endpoint links to that instead of the external Google bucket/CDN. The previous approach lead to an issue when trying to use pyvo to query asynchronously using the "CredentialStore" class for authentication, as the auth headers were sent to the Google CDN at the step of fetching the results, leading to an error HTTP response.

The changes in phalanx are upgrades to the Docker images for tap-postgres & lsst-tap-service, where these changes are applied.
The TAP service image upgrades also include upgrading the Capabilites generating code, to match the CADC upstream changes.

One additional change is to add a path_prefix, which is used by both the results and the capabilities servlet.

**How was this tested**

Deployed & tested on dev
- pyvo async & sync queries
- taplint validation

**Potential issues**

Has not been tested on environments that use S3 bucket subdirectories, I've added a JUnit test for this case, but that's one type of environment I'm less confident will not break.


athornton (approved):
LGTM (assuming the referenced container versions behave as stated, anyway)

PR #3498, opened 2024-07-09
DM-44509: Upgrade Butler server, datalinker, and vo-cutouts

dhirving:
The new release of Butler server includes backwards-incompatible changes to the REST API, so datalinker and vo-cutouts need new versions with an updated client library simultaneously.


dhirving:
Russ is going to do this in a different PR because there are also some `datalinker` changes that need to be addressed.

PR #3509, opened 2024-07-11
DM-44635: mobu: nicer config

fajpunk:
The GitHub CI app and Refresh app functionality are now completely separated. They each have their own:
* Config dicts in chart values
* Anonymous Ingresses
* ConfigMaps that get mapped to yaml files

Having separate config dicts and ingresses makes it a bit more clear in Phalanx that these are independently toggle-able functionalities, and having config come from completely separate ConfigMaps makes the logic to toggle them in the Mobu app much cleaner.

The mobu changes are in https://github.com/lsst-sqre/mobu/pull/354.


applications/mobu/values-base.yaml
```diff
@@ -1,3 +0,0 @@
-config:
-  # Mobu here is behind a VPN and is not reachable by github.com
-  githubIntegration: false
```
fajpunk, line 3: All GitHub integration is disabled by default.

applications/mobu/values-tucson-teststand.yaml
```diff
@@ -1,3 +0,0 @@
-config:
-  # Mobu here is behind a VPN and is not reachable by github.com
-  githubIntegration: false
```
fajpunk, line 3: All GitHub integration is disabled by default.

applications/mobu/templates/configmap-github-ci-app.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.config.githubCiApp -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "mobu.fullname" . }}-github-ci-app
+  labels:
+    {{- include "mobu.labels" . | nindent 4 }}
+data:
+  yaml: |
```
rra, line 9: Nit: I would use the filename here rather than `yaml` as a key. Either way works, but the file name is a bit more self-documenting.

applications/mobu/templates/configmap-github-refresh-app.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.config.githubRefreshApp -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "mobu.fullname" . }}-github-refresh-app
+  labels:
+    {{- include "mobu.labels" . | nindent 4 }}
+data:
+  yaml: |
```
rra, line 9: Likewise here.

applications/mobu/values-idfdev.yaml
```diff
@@ -1,15 +1,24 @@
+image:
+  pullPolicy: "Always"
+  tag: "tickets-DM-44635-nicer-github-integration-config"
 config:
   debug: true
-  githubRefreshAppEnabled: true
-  githubCiAppEnabled: true
-  github:
+  githubRefreshApp:
+    accepted_github_orgs:
```
rra, line 7: If you add that `to_caml` setting, you can now use `acceptedGithubOrgs` here instead, which looks more natural in Helm.

applications/mobu/values.yaml
```diff
@@ -26,9 +26,13 @@ config:
...
-  # -- GitHub app integration options.
-  # @default -- See `values.yaml`
-  github: {}
+  # -- Configuration for the GitHub refresh app integration.
+  # @default -- disabled.`
+  githubRefreshApp: {}
+
+  # -- Configuration for the GitHub CI app integration.
+  # @default -- disabled.`
+  githubCiApp: {}
```
rra, line 35: It would be really nice to be able to get documentation for all of the available options into the Helm `README.md` file by documenting each field, which was one of the other things that pushed me towards a combined configuration file format. But avoiding that for now, maybe add a URL to the mobu documentation?

The `@default` string here seems to be invalid Markdown.

rra (approved):

PR #3511, opened 2024-07-11
Add cert-manager annotation to monitoring ingress

athornton:


afausti (approved):
Oh ok. So, every hostname needs the letsencrypt annotation. For the phalanx "environment" hostname the annotation is added to argo-cd. Having a separate hostname for monitoring breaks the symmetry a bit, but we need it because it is an InfluxDB2 UI limitation.

PR #3543, opened 2024-07-19
DM-40434: Prompt Processing continues processing after timeout or termination

kfindeisen:
This PR reworks Prompt Processing's timeout configuration to be worker-centric rather than Knative-centric. It also adds a control for each worker's grace period.

### DIFF
--- a/applications/prompt-proto-service-hsc-gpu/README.md
+++ b/applications/prompt-proto-service-hsc-gpu/README.md
@@ -39,10 +39,10 @@ Prompt Proto Service is an event driven service for processing camera images. Th
 | prompt-proto-service.knative.extraTimeout | int | `10` | To acommodate scheduling problems, Knative waits for a request for twice `worker.timeout`. This parameter adds extra time to that minimum (seconds). |
 | prompt-proto-service.knative.gpu | bool | `true` | GPUs enabled. |
 | prompt-proto-service.knative.gpuRequest | int | `1` | The number of GPUs to request. |
-| prompt-proto-service.knative.idleTimeout | int | `900` | Maximum time that a container can send nothing to the fanout service (seconds). |
+| prompt-proto-service.knative.idleTimeout | int | `0` | Maximum time that a container can send nothing to Knative (seconds). This is only useful if the container runs async workers. If 0, idle timeout is ignored. |
 | prompt-proto-service.knative.memoryLimit | string | `"8Gi"` | The maximum memory limit. |
 | prompt-proto-service.knative.memoryRequest | string | `"2Gi"` | The minimum memory to request. |
-| prompt-proto-service.knative.responseStartTimeout | int | `900` | Maximum time that a container can send nothing to the fanout service after initial submission (seconds). |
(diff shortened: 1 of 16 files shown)


applications/prompt-proto-service-hsc-gpu/README.md
```diff
@@ -36,13 +36,13 @@ Prompt Proto Service is an event driven service for processing camera images. Th
 | prompt-proto-service.knative.cpuRequest | int | `1` | The cpu cores requested. |
 | prompt-proto-service.knative.ephemeralStorageLimit | string | `"5Gi"` | The maximum storage space allowed for each container (mostly local Butler). |
 | prompt-proto-service.knative.ephemeralStorageRequest | string | `"5Gi"` | The storage space reserved for each container (mostly local Butler). |
+| prompt-proto-service.knative.extraTimeout | int | `10` | To acommodate scheduling problems, Knative waits for a request for twice `worker.timeout`. This parameter adds extra time to that minimum (seconds). |
 | prompt-proto-service.knative.gpu | bool | `true` | GPUs enabled. |
 | prompt-proto-service.knative.gpuRequest | int | `1` | The number of GPUs to request. |
-| prompt-proto-service.knative.idleTimeout | int | `900` | Maximum time that a container can send nothing to the fanout service (seconds). |
+| prompt-proto-service.knative.idleTimeout | int | `0` | Maximum time that a container can send nothing to the fanout service (seconds). This is only useful for async workers. If 0, idle timeout is ignored. |
```
dspeck1, line 42: What do you mean by async workers?  Knative does not support async.

charts/prompt-proto-service/templates/prompt-proto-service.yaml
```diff
@@ -42,6 +42,12 @@ spec:
         env:
         - name: WORKER_RESTART_FREQ
           value: {{ .Values.worker.restart | toString | quote }}
+        - name: WORKER_TIMEOUT
+          value: {{ .Values.worker.timeout | toString | quote }}
+        - name: WORKER_GRACE_PERIOD
+          value: {{ .Values.worker.grace_period | toString | quote }}
+        {{- /* Knative doesn't allow timeouts longer than 1200 seconds */ -}}
```
dspeck1, line 49: Timeouts can be increased by changing the config-defaults config map value of  `max-revision-timeout-seconds: "1200"` to a higher value.

dspeck1 (approved):
LGTM.   Added two comments for consideration in updating comments.

applications/prompt-proto-service-hsc-gpu/README.md
kfindeisen: I was referring to [Gunicorn](https://docs.gunicorn.org/en/stable/design.html) support for intermediate responses and keepalives. I think in addition to switching worker type we'd have to use something other than Flask for the actual request processing?

dspeck1: Oh I see.  Thanks.  Correct. 

Something like this?  "Knative timeout values.  Note that gunicorn/flask timeout values are set"

PR #3545, opened 2024-07-19
Enable Times-Square for IDF environments

athornton:
Since we're now using Times Square for templating of Rubin-menu Notebook Aspect features, we need to enable it on "our" RSPs.

We should probably encourage T&S and USDF to do the same.


jonathansick:
We can't do this quite yet because I need to set up the GitHub App for each environment and put its secrets into 1Password. We'll talk on Monday about how to do the roll-out.

This also won't work on any environment that's behind a VPN, so we'll also need to talk about this on Monday.

athornton:
Fair, and I need SQL setup in each environment too, so converting back to draft.

jonathansick:
If you want to set up the GitHub Apps, here's the documentation page on their configuration. There's a pre-configured URL you can use to start the setup. https://times-square.lsst.io/user-guide/github-app-config.html

For data.lsst.cloud, I'd recommend increasing the Helm value for the Redis storage. This is something I want to do in the existing environments too, but it'd be good to get ahead of this. This Redis contains the HTML cache of notebook renders so it'd be good to have the redis storage in the scale of hundreds of GB if that's reasonable.

athornton:
Tabling the redis storage for now.

You're right that we need more, but I think I need to open a co-work discussion for redis and firefly (in particular) but generally the "ephemeral storage needs fixin' right now" story, because I think it's a pretty big overall storage increase and we need to figure out how much we are willing to spend right now.

I'll open the ticket, anyway.  DM-45405

rra:
Maybe I'm saying something everyone already knows, but just in case: Redis doesn't use ephemeral storage. It has a PVC.

athornton:
Yes; that ticket has details, but we're talking about a significant expansion of our PVC footprint, as well as some tricks related to how we're going to use (or not) ephemeral storage.

PR #3554, opened 2024-07-24
DM-43599: Add progress logs to slow AP tasks

kfindeisen:
This PR adds verbose logging to two tasks known to run slowly in Prompt Processing. This activates some logs that were added on lsst/meas_base#280, lsst/meas_transiNet#27, and lsst/ap_association#229. The tasks do not produce verbose logs otherwise.


bsmartradio (approved):
Looks good to me. Does the logLevel readme need to be updated to mention that diaPipe and rbClassify are verbose? I wouldn't think its necessary but it might be worth it to mention it so people don't except otherwise.

kfindeisen:
I hope the extra logging will be temporary, much like when we turn on debug for just one module.

PR #3559, opened 2024-07-26
Switch to using FDQN for Qserv - no op

frossie:
Still pointing to USDF Qserv prod but now using DNS


stvoutsin:
Looks good to me although I haven't tested it

stvoutsin (approved):
Looks good to me although I haven't tested it

PR #3562, opened 2024-07-26
tickets/DM-45405: Make portal private workarea able to come from allocated storage

athornton:


applications/portal/templates/statefulset.yaml
```diff
@@ -181,3 +211,15 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+  {{- if .Values.config.volumes.privateWorkarea.pvc }}
+  volumeClaimTemplates:
+    - metadata:
+        name: "firefly-private-workarea"
```
rra, line 217: The `StatefulSet` pod name will be appended to this name, so you want to set it to something way shorter, like `workarea`. It will be obvious that it's private since the pod name will be part of the name.

applications/portal/values.yaml
```diff
@@ -72,26 +72,39 @@ config:
...
+  # "path", which indicates the hostPath to mount there.
+  # For the key "nfs", the value is an object with "path" and "server"
+  # keys, both strings.
+  # For the key "pvc", the value is an object with "size" and
+  # "storageClass" keys.  If "pvc" is chosen, the storage class must
+  # support mode "ReadWriteOnce" for the `privateWorkarea`,
+  # "ReadWriteMany"
+  # for `sharedWorkarea`, and "ReadOnlyMany" for `config`.
+  # If all objects have null values, an `emptyDir` is used for this volume.
+  # @default -- See `values.yaml`
```
rra, line 88: This turns into a really long documentation string and is going to cause problems for the page rendering. I think the best way to document something like this may be to add a documentation page to the Phalanx documentation under applications/portal about how to set up the storage and just link to it.

The constraints on mode should be moved to the documentation for each type of volume, since they're only relevant there and change for each type of volume.

rra (approved):
Basic change seems fine, just a couple of notes on the documentation.

applications/portal/values.yaml
athornton: I've added the page and put both what I've done, and why, in there.  That will also be helpful, once merged, for my upcoming discussion with @gpdf about this.

PR #3572, opened 2024-07-29
DM-45466: Optional QServ pw secret & idfint to point to new pw-protected Qserv

stvoutsin:
This PR modifies the cadc-tap chart to accept and pass in an optional QServ password. The password is expected and passed in when `config.qserv.passwordEnabled` is set to true otherwise the qserv password env value is empty, which is what it was previously to this change.

The other change is to move idfint to use the new QServ host: "qserv-int.slac.stanford.edu:4040" with passwordEnabled: true.
 Additional changes include adding a qserv-password for idfint (secrets-idfint.yaml).

**JIRA issue:**
https://rubinobs.atlassian.net/browse/DM-45466

**Tests:**
The change was tested on dev, by applying the equivalent changes here, and testing with the new Qserv and passwordEnabled, as well as with the previous and without passwordEnabled

### DIFF
--- a/applications/tap/secrets-idfint.yaml
+++ b/applications/tap/secrets-idfint.yaml
@@ -1,3 +1,7 @@
 uws-db-password:
   description: >-
     Password for external UWS PostgreSQL server
+qserv-password:
+  description: >-
+    Password for the QServ database server
+  if: cadc-tap.config.qserv.passwordEnabled
--- a/charts/cadc-tap/templates/tap-deployment.yaml
+++ b/charts/cadc-tap/templates/tap-deployment.yaml
@@ -86,7 +86,7 @@ spec:
                 {{- end }}
                 {{- if eq .Values.config.backend "qserv" }}
                 -Dqservuser.username=qsmaster
-                -Dqservuser.password=
+                -Dqservuser.password=$QSERV_DB_PASSWORD
                 -Dqservuser.driverClassName=com.mysql.cj.jdbc.Driver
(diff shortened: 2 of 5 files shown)


charts/cadc-tap/templates/tap-deployment.yaml
```diff
@@ -100,6 +100,13 @@ spec:
                 -Dbase_url={{ .Values.global.baseUrl }}
                 -Dca.nrc.cadc.util.PropertiesReader.dir=/etc/creds/
                 -Xmx{{ .Values.config.jvmMaxHeapSize }}
+            {{- if and (eq .Values.config.backend "qserv") (eq .Values.config.qserv.passwordEnabled true) }}
```
rra, line 103: Nit: rather than `(eq ... true)` you can just use the value itself for its truth value.
```suggestion
            {{- if (and (eq .Values.config.backend "qserv") .Values.config.qserv.passwordEnabled) }}
```
Pretty sure that works.

applications/tap/secrets-idfint.yaml
```diff
@@ -1,3 +1,6 @@
 uws-db-password:
   description: >-
     Password for external UWS PostgreSQL server
+qserv-password:
+  description: >-
+    Password for the QServ database server
```
rra, line 6: You need the conditional here so that the secret isn't required everywhere.
```suggestion
qserv-password:
  description: >-
    Password for the QServ database server
  if: cadc-tap.config.qserv.passwordEnabled
```

rra (approved):
Looks good to me. The secrets management system requires conditionality be keyed off of some `values.yaml` setting, so there's no way to avoid having to add a new setting, I'm afraid.

PR #3573, opened 2024-07-30
DM-45494: Add fastapi-safir-uws starter

rra:
Add a starter for vo-cutouts-style FastAPI Safir apps that use UWS. Update the code underlying phalanx application create so that it will substitute `<CHARTENVPREFIX>` in every template, not just `configmap.yaml`.

### DIFF
--- a/starters/fastapi-safir-uws/values.yaml
+++ b/starters/fastapi-safir-uws/values.yaml
@@ -1,0 +1,232 @@
+# Default values for <CHARTNAME>.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+config:
+  # -- Choose from the text form of Python logging levels
+  loglevel: "INFO"
+
+  # -- Logging profile (`production` for JSON, `development` for
+  # human-friendly)
+  logProfile: "production"
+
+  # -- URL path prefix for the API
+  pathPrefix: "/api/<CHARTNAME>"
+
+  # -- URL for the PostgreSQL database if Cloud SQL is not in use
+  # @default -- None, must be set if `cloudsql.enabled` is false
+  databaseUrl: null
+
+  # -- Grace period in seconds to wait for worker jobs to finish
+  gracePeriod: 60
+
+  # -- Lifetime of job results in Safir `parse_timedelta` format
+  lifetime: "30d"
+
+  # -- Google service account with an IAM binding to the Kubernetes service
+  # account for the application. Must have the `cloudsql.client` role, access
+  # to write to the GCS bucket, and ability to sign URLs as itself
+  # @default -- None, must be set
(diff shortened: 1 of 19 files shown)


rra:
This depends on https://github.com/lsst-sqre/safir/pull/281 and https://github.com/lsst/templates/pull/260.

starters/fastapi-safir-uws/values.yaml
```diff
@@ -0,0 +1,232 @@
...
+  # @default -- See `values.yaml`
+  resources:
+    limits:
+      cpu: "100m"
+      memory: "20Mi"
+    requests:
+      cpu: "5m"
+      memory: "7Mi"
+
+cutoutWorker:
```
jonathansick, line 108: Seems like this should be generic, like "worker" or "pipelinesWorker" ?

starters/fastapi-safir-uws/values.yaml
```diff
@@ -0,0 +1,232 @@
...
+    secretKey: "redis-password"
+
+  persistence:
+    # -- Whether to persist Redis storage and thus tokens. Setting this to
+    # false will use `emptyDir` and reset all tokens on every restart. Only
+    # use this for a test deployment.
+    enabled: true
+
+    # -- Amount of persistent storage to request
+    size: "100Mi"
```
jonathansick, line 181: I wonder if it's better to error on the side of being a bit larger since resizing PVCs is a pain. Up to you, though.

jonathansick (approved):
Looks good 👍

starters/fastapi-safir-uws/values.yaml
rra: Thank you, good catch. I grepped for `cutouts` and not `cutout`. Found and fixed all of these, I think.

starters/fastapi-safir-uws/values.yaml
rra: I was going to say that the amount of data stored should be quite small since the largest thing stored here in this pattern is the arguments to the backend, but the vo-cutouts data is about 50MiB, so maybe it will be larger than I expect? Although that's also very surprising to me; that's way more data than I was expecting. I'll dig into this a bit.

jonathansick: Aren't task outputs also stored here, including exception traces?

rra: Outputs are not, only a URL. Exception traces are, but only for a day.

It looks like Redis is not periodically cleaning up the appendonly database for some reason. I may be missing some command-line flags in our Redis chart. When I manually ran `BGREWRITEAOF`, the size of the vo-cutouts Redis database went down to about 1KB.

PR #3581, opened 2024-08-02
[DM-45537] Deploy sasquatch-backpack at the summit

afausti:
Configure Sasquatch REST Proy URL, enable sasquatch-backpack at the summit and replication to USDF.


Fireye04 (approved):
👍

PR #3583, opened 2024-08-02
DM-45572: Move Butler server configuration files into Phalanx

dhirving:
Butler configuration files on IDF were previously hosted in an S3 bucket, which is difficult to manage and not version controlled.  They are now injected from Phalanx as a ConfigMap instead.


rra (approved):
Looks good to me!

PR #3595, opened 2024-08-05
DM-45616: Serve DirectButler configuration files from Butler server

dhirving:
Set up the configuration files used for DirectButler to be generated inside Phalanx, and serve them using a new static files feature in Butler server.

This will allow these files to be version controlled and deployed as part of the normal IDF deployment process.


applications/butler/Chart.yaml
```diff
@@ -4,4 +4,4 @@ version: 1.0.0
 description: Server for Butler data abstraction service
 sources:
   - https://github.com/lsst/daf_butler
-appVersion: w.2024.27
+appVersion: tickets-dm-45616
```
dhirving, line 7: I will make a proper release of the server prior to merging this

rra (approved):
Looks good!

PR #3596, opened 2024-08-05
DM-45620: Removed unused butlerRepositoryIndex config

dhirving:
The environment-global configuration item`butlerRepositoryIndex` appears to be unused now that datalinker and vo-cutouts have dropped support for DirectButler, so it can be removed.  (nublado always separately hardcoded this path in its own configuration, and nothing else is using it.)


dhirving:
I can't fully test this because I no longer have permissions to override the `science-platform` root project, which is needed to sync the change the environment-global configuration.  But `vo-cutouts` and `datalinker` look OK.

rra (approved):
I was going to ask if that was still necessary. Thanks for doing this cleanup -- much appreciated!

PR #3603, opened 2024-08-08
DM-45675: Fix error with minikube test deploy (Re-sync infrastructure apps after fail)

stvoutsin:
**Description:**
Upgrading to argo-helm version 7.4.1 causes minikube deploy to fail with an error. This PR addresses the issue, by re-syncing the infrastructure apps after the first fail. For context the actual error is:
`InvalidSpecError: Application referencing project infrastructure which does not exist
`
A similar fix is used when syncing the argocd app.

**Additional change**
Because the fix causes the install method to be too complex for ruff's linting checks, for now I've added a noqa.
I think it may be worth breaking out the functionality into smaller private methods, which I can do in a separate PR if that makes sense.

**Jira Issue:**
https://rubinobs.atlassian.net/browse/DM-45675

**Tested:**
This fix was tested by a PR to trigger the deploy which can be seen here:
https://github.com/lsst-sqre/phalanx/pull/3602/files
https://github.com/lsst-sqre/phalanx/actions/runs/10306433160

**Note:**
I'm not upgrading argo-helm here to try and keep the fixes/PR's separate.

### DIFF
--- a/src/phalanx/services/environment.py
+++ b/src/phalanx/services/environment.py
@@ -4,15 +4,18 @@ from __future__ import annotations
 
 from datetime import timedelta
 
+from pydantic import SecretStr
+
 from ..exceptions import CommandFailedError, VaultNotFoundError
 from ..github import action_group, add_mask
 from ..models.applications import Project
+from ..models.environments import Environment
 from ..models.vault import VaultCredentials
 from ..storage.argocd import ArgoCDStorage
 from ..storage.config import ConfigStorage
 from ..storage.helm import HelmStorage
 from ..storage.kubernetes import KubernetesStorage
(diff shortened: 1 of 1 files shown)


rra (approved):
Could you add a comment explaining why we're doing this?

Also, as mentioned in Slack, I agree with refactoring this function to break it up a little, either in this PR or another one.

stvoutsin:
Okay yep I'll add a comment and break it up in this PR

stvoutsin:
I've broken up the install method a bit & added a comment on the re-sync step for the apps. 
Open to suggestions if the change is missing anything or isn't as clear or extensible as we want.

rra:
Looks good to me. Go ahead and merge!

PR #3620, opened 2024-08-15
DM-45807: Update USDF TAP authentication to qserv.

ctslater:


ctslater:
This is deployed on a branch on USDF and working, but I have one question for @pav511: I coped the `secrets-usdf*` files from the IDF versions, including the field for `uws-db-password`. I'm not sure if that secret is actually used at USDF, or if it matters to have it in here at all? It doesn't seem to be hurting anything, but I can clean it up if it's unnecessary.

pav511:
I don't think these secrets are needed for usdf, the qserv password is already present in a different secret in usdf vault. Which env was this tested in?  It looks like all environments were synced to main. I moved usdf-int tap to point to this ticket branch though, It looks like changes are ok.

ctslater:
I got super confused by testing this with an invalid RA, Dec (turns out there's no data at 62, **+37**), so the only net change through these pushes was to remove the unused secret field.

PR #3628, opened 2024-08-19
DM-42606 Update schema registry sync to load all schemas

bsmartradio:
Remakes schema registry every time the broker is fully synced. Keeps versioning consistent in the registry.

### DIFF
--- a/applications/alert-stream-broker/charts/alert-stream-schema-registry/templates/sync-schema-job.yaml
+++ b/applications/alert-stream-broker/charts/alert-stream-schema-registry/templates/sync-schema-job.yaml
@@ -18,12 +18,15 @@ spec:
         app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
         app.kubernetes.io/instance: {{ .Release.Name | quote }}
         helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+      annotations:
+        description: "Load schemas into registry."
     spec:
       restartPolicy: Never
       containers:
       - name: sync-schema-job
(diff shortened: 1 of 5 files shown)


applications/alert-stream-broker/charts/alert-stream-schema-registry/templates/sync-schema-job.yaml
```diff
@@ -18,12 +18,16 @@ spec:
...
         app.kubernetes.io/instance: {{ .Release.Name | quote }}
         helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+      annotations:
+        description: "Load schemas into registry."
     spec:
       restartPolicy: Never
       containers:
       - name: sync-schema-job
-        image: "{{ .Values.schemaSync.image.repository }}:{{ .Values.schemaSync.image.tag | default .Chart.AppVersion }}"
+        #image: "{{ .Values.schemaSync.image.repository }}:{{ .Values.schemaSync.image.tag | default .Chart.AppVersion }}"
```
ebellm, line 27: do we still need the commented line?

ebellm (approved):

applications/alert-stream-broker/charts/alert-stream-schema-registry/templates/sync-schema-job.yaml
bsmartradio: I can probably remove it. I was switching back and forth initially, but using the digest is more fine grained and makes sure that the correct docker image gets used, so its probably better to stick with that.

ebellm (approved):

PR #3650, opened 2024-08-27
DM-45907: Increase Butler memory limit

dhirving:
Butler server's memory limit was set too low previously -- it could easily result in an out-of-memory kill with just a couple in-flight requests.


rra (approved):
Looks good to me.

PR #3654, opened 2024-08-30
[DM-40858] Deploy the Telegraf-based connectors on the base cluster

afausti:
Remove kafka-connect-manager configuration and switch over to Telegraf.

### DIFF
--- a/applications/sasquatch/values-base.yaml
+++ b/applications/sasquatch/values-base.yaml
@@ -132,146 +132,78 @@ influxdb:
     hostname: base-lsp.lsst.codes
 
 telegraf-kafka-consumer:
-  enabled: false
+  enabled: true
   kafkaConsumers:
     auxtel:
       enabled: true
-      replicaCount: 1
-      interval: "1s"
-      flush_interval: "1s"
-      union_mode: "nullable"
-      timestamp_format: "unix"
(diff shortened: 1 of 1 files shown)


applications/sasquatch/values-base.yaml
```diff
@@ -131,199 +131,95 @@ influxdb:
...
-        topicsRegex: "lsst.sal.LaserTracker"
-      genericcamera:
-        enabled: true
-        topicsRegex: "lsst.sal.GCHeaderService|lsst.sal.GenericCamera"
-      lsstcam:
-        enabled: true
-        topicsRegex: "lsst.sal.MTCamera|lsst.sal.MTHeaderService|lsst.sal.MTOODS"
-
 telegraf-kafka-consumer:
   enabled: false
```
mareuter, line 135: This seems odd to be `false`. Does this do anything?

mareuter (approved):
One minor thing otherwise this looks good to me. Not sure if we need the `newest` setting since I plan on burning all the topic down before starting up again. The topics won't have a lot in them when the new sinks are turned on.

applications/sasquatch/values-base.yaml
afausti: Thanks this needs to be `true`.

PR #3662, opened 2024-09-04
[DM-46096] love: refactor configurations to split managers producers by categories on BTS

sebastian-aranda:


mareuter (approved):
Looks good to me.

PR #3664, opened 2024-09-06
tickets/DM-46156: Add LSSTCam butler directories to nublado on the summit

mareuter:
Summit: Add LSSTCam butler directories to nublado.


athornton (approved):
LGTM

PR #3668, opened 2024-09-06
[DM-46176] Add an extra tag for the Telegraf internal metrics 

afausti:
We need an extra tag to identify the Telegraf instance from which the internal metrics are collected when running multiple instances of Telegraf (e.g. one Telegraf instance per connector).
As a result route the Telegraf internal metrics to its own database to separate them and configure a different retention period for Telegraf database.

Example:

image


mareuter (approved):
Looks good to me.

PR #3673, opened 2024-09-10
[DM-46252] Review Telegraf connector configuration for high throughput 

afausti:
The Kafka Consumer input plugin has three configuration parameters that impact throughput, `metric_batch_size`,  `max_undelivered_messages` and `compression_codec`.
When increasing `metric_batch_size` we also need to increase `max_undelivered_messages` to flush the messages from the brokers. 
Based on tests at the Summit tuning these parameters seems enough to increase the connector throughput to handle M1M3 and MTMount throughput and is a better strategy than running multiple replicas of Telegraf to keep up with the telemetry stream.

```
  ## Max undelivered messages
  ## This plugin uses tracking metrics, which ensure messages are read to
  ## outputs before acknowledging them to the original broker to ensure data
  ## is not lost. This option sets the maximum messages to read from the
  ## broker that have not been written by an output.
  ##
  ## This value needs to be picked with awareness of the agent's
  ## metric_batch_size value as well. Setting max undelivered messages too high
  ## can result in a constant stream of data batches to the output. While
  ## setting it too low may never flush the broker's messages.
  # max_undelivered_messages = 1000
```
The other configuration parameter that has an impact is compression.
```
  ## Compression codec represents the various compression codecs recognized by
  ## Kafka in messages.
  ##  0 : None
  ##  1 : Gzip
  ##  2 : Snappy
  ##  3 : LZ4
  ##  4 : ZSTD
  # compression_codec = 0
```
This configuration is handling well the high throughput CSCs with only one Telgraf replica:

image


mareuter (approved):
Looks good to me. Need to get this on BTS to see what difference it makes.

PR #3701, opened 2024-09-20
[DM-46250] Review configuration for the Telegraf connectors at BTS

afausti:
We optimized the connector configuration for throughput, we can run m1m3 and mtmount connectors with one replica now. Also use a Telegraf version that fixes a race condition bug found in the Telegraf Avro parser.


mareuter (approved):
Looks good to me.

PR #3702, opened 2024-09-20
DM-46286: Make Butler client/server default on IDF int/dev dp02

dhirving:
To facilitate testing by the Community Science Team and start getting stability testing from Mobu, switch the default dp02 alias on `idfdev` and `idfint` to point to client/server Butler instead of DirectButler.

`idfprod` will remain on DirectButler until testing has been completed and information about the transition has been communicated to users.


rra (approved):
Thanks, looks good to me.

PR #3715, opened 2024-09-24
Sasquatch USDF-DEV additional next visit test topic

dspeck1:
Add additional test topic for testing job based processing with keda.


afausti (approved):
LGTM

PR #3722, opened 2024-09-25
DM-46257: Add initial version of ppdb-replication application

JeremyMcCormick:
Add initial version of an application which replicates data between the APDB and PPDB using utilities from [dax_ppdb](https://github.com/lsst/dax_ppdb).

This application is only enabled on `usdf-rsp-dev` for now.

### DIFF
--- a/applications/ppdb-replication/templates/ingress.yaml
+++ b/applications/ppdb-replication/templates/ingress.yaml
@@ -1,0 +1,30 @@
+apiVersion: gafaelfawr.lsst.io/v1alpha1
+kind: GafaelfawrIngress
+metadata:
+  name: "ppdb-replication"
+  labels:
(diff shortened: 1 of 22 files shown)


applications/ppdb-replication/templates/ingress.yaml
andy-slac, line 1: I do not know if need ingress at all, but probably OK to keep it for now.

andy-slac (approved):
Looks reasonable to me, though I know nothing about phalanx/k8s. 🙂

applications/ppdb-replication/templates/ingress.yaml
JeremyMcCormick: This template is automatically copied into the application when it is create by phalanx, and I just left it.

PR #3725, opened 2024-09-25
Add lsst.obsenv namespace for telescope environments

afausti:
The `lsst.obsenv` namespace was requested in Sasquatch for additional information about the observing environment. 
This PR registers the new namespace with the REST Proxy API, add the connector configuration for the telescope environments and enables replication to USDF.


tribeiro (approved):
Looks good! Thanks for the update!

PR #3730, opened 2024-09-26
DM-46399: Convert manual ingresses to GafaelfawrIngress

rra:
The next release of Gafaelfawr will drop support for manually configuring Kubernetes ingresses to use Gafaelfawr and will require the use of `GafaelfawrIngress` resources. Convert the three remaining relevant usages in Phalanx.

These ingresses previously didn't require any specific scopes for access. I added `exec:internal-tools` for this purpose about a year ago so everyone's tokens should now have that scope. Add that scope restriction as part of this conversion if authentication is enabled.

Currently, the authentication portion was disabled. Preserve that by configuring the ingreses to be anonymous by default, but add a configuration option that can be used to enable authentication when that is ready.


rra:
FYI, I'm about to go on vacation for a week, so I won't merge this before October 7th (and therefore won't merge this before teh next summit patch window) to ensure I'll be around if there are any problems.

sebastian-aranda (approved):
Looks good to me :+1:.

PR #3731, opened 2024-09-26
DM-46399: Convert manual ingress to GafaelfawrIngress

rra:
The next release of Gafaelfawr will drop support for manually configuring Kubernetes ingresses to use Gafaelfawr and will require the use of `GafaelfawrIngress` resources. Convert the ingress for the alert-database subchart of alert-stream-broker to use `GafaelfawrIngress`.


bsmartradio (approved):
This looks good to me. Is there any documentation I should read about this new implementation?

rra:
Everything should be documented here: https://gafaelfawr.lsst.io/user-guide/gafaelfawringress.html

For your use case, it should be fairly straightforward, since you're not using any of the fancy Gafaelfawr features. Unfortunately there isn't a great way to test and I'm about to go on vacation for a week, so I'll hold off on merging this until I'm back and can be on the patching call with Peter when he updates usdf-dev in case there are any problems. I think this ingress is only deployed there, not in the alert-stream-broker cluster.

PR #3734, opened 2024-09-27
[DM-46540] nightreport: fix summit site_id

sebastian-aranda:


pothiers (commented):
trivial but important change

PR #3737, opened 2024-09-27
DM-45522: App metrics events in Sasquatch

fajpunk:
Sasquatch subchart that creates and configures:

* A new telegraf consumer for all app metrics
* Per-app Avro tags config for that consumer
* KafkaUser and KafkaTopic resources per-app

See https://github.com/lsst-sqre/sasquatch/pull/41 for how to use this
new subchart.


jonathansick:
It looks like you may need to enable pre-commit locally to get helm docs to run? I think that's the linting error.

fajpunk:
Gah, sorry! Lint should be fixed now.

applications/sasquatch/charts/app-metrics/values.yaml
```diff
@@ -0,0 +1,114 @@
...
+#     influxTags:
+#       - some
+#       - tags
+#   some-other-app:
+#     influxTags:
+#       - some
+#       - tags
+#       - and
+#       - others
+globalAppConfig: {}
```
jonathansick, line 18: These big helm-docs comments kind of break the Markdown table, which is oriented around single paragraph descriptions. I suppose we could [look into HTML tables](https://github.com/norwoodj/helm-docs#advanced-table-rendering), but in the meantime I wonder if we could link out to documentation for this structure elsewhere? Like in the Sasquatch docs?

applications/sasquatch/charts/app-metrics/values.yaml
```diff
@@ -0,0 +1,114 @@
...
+
+# -- A list of applications that will publish metrics events, and the keys that should be ingested into InfluxDB as tags.
+# The names should be the same as the app names in Phalanx.
+apps: []
+
+# -- Keys in an every event sent by any app that should be recorded in InfluxDB
+# as "tags" (vs. "fields").
+#
+# -- These will be concatenated with the `influxTags` from `globalAppConfig`
```
jonathansick, line 27: It looks like helm-docs ignores the first paragraph and only includes the second. So Perhaps collapse into a single line?

applications/sasquatch/values.yaml
```diff
@@ -283,3 +283,10 @@ global:
   # -- Base path for Vault secrets
   # @default -- Set by Argo CD
   vaultSecretsPath: ""
+
+app-metrics:
```
jonathansick, line 287: I think the Helm style uses camelCase for variables, so this would be `appMetrics`?

jonathansick (approved):
Looks like a nice solution for configuring the metrics submission. I just have some style notes.

applications/sasquatch/values.yaml
jonathansick: Oh right. Good call. I guess using the sub chart name is necessary to map to its own values.

applications/sasquatch/charts/app-metrics/templates/telegraf-configmap.yaml
```diff
@@ -0,0 +1,65 @@
...
+      consumer_group = "telegraf-kafka-consumer-app-metrics"
+      sasl_mechanism = "SCRAM-SHA-512"
+      sasl_password = "$TELEGRAF_PASSWORD"
+      sasl_username = "telegraf"
+      data_format = "avro"
+      avro_schema_registry = "http://sasquatch-schema-registry.sasquatch:8081"
+      avro_timestamp = "timestamp_ns"
+      avro_timestamp_format = "unix_ns"
+      avro_union_mode = "nullable"
+      avro_tags = {{ include "helpers.toTomlArray" (index $.Values.globalAppConfig $app "influxTags")  }}
```
afausti, line 52: You'll probably always have at least one tag configured for the app, but I would be a little more defensive here testing if `influxTags` is not empty first.

applications/sasquatch/charts/app-metrics/values.yaml
```diff
@@ -0,0 +1,114 @@
...
+# @default -- false
+debug: false
+
+influxdb:
+  # -- URL of the InfluxDB v1 instance to write to
+  url: "http://sasquatch-influxdb.sasquatch:8086"
+
+# -- Number of Telegraf  replicas. Increase this value to increase the consumer
+# throughput.
+replicaCount: 1
```
afausti, line 83: Just notice that this will increase the consumer throughput only if you create the kafka topics with multiple partitions, where the number of partitions is a multiple of the number of Telegraf replicas.
Even for the EFD use case where we deal with high volumes of data I don't need this.
I think what would be more interesting for the app metrics case is to run 3 replicas for high availability.

afausti (approved):
This looks great, I've made a few minor comments.

PR #3738, opened 2024-09-27
[DM-46550] Add a named template for generating the Telegraf configmap

afausti:
Use a named template for generating the Telegraf ConfigMap and annotate the Deployment with the ConfigMap checksum. This will make the Deployment out-of-sync when the ConfigMap changes, and when the Deployment is synced it will cause the Pod to restart.

The only caveat is that we need to sync the ConfigMap first and then sync the Deployment.

Tested with the Telegraf instances running on data-int and is working fine.


mareuter (approved):
Looks good to me.

PR #3742, opened 2024-09-30
DM-45522 mobu: App metrics events config

fajpunk:
Provision initial app metrics Sasquatch config for mobu, following these instructions:
https://sasquatch.lsst.io/user-guide/app-metrics.html#configuration

### DIFF
--- a/applications/sasquatch/charts/app-metrics/values.yaml
+++ b/applications/sasquatch/charts/app-metrics/values.yaml
@@ -5,7 +5,12 @@
 # overridden.
 # See [here](https://sasquatch.lsst.io/user-guide/app-metrics.html#configuration)
 # for the structure of this value.
-globalAppConfig: {}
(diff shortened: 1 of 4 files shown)


applications/sasquatch/charts/app-metrics/values.yaml
```diff
@@ -5,7 +5,10 @@
 # overridden.
 # See [here](https://sasquatch.lsst.io/user-guide/app-metrics.html#configuration)
 # for the structure of this value.
-globalAppConfig: {}
+globalAppConfig:
```
rra, line 8: helm-docs doesn't handle complex default values very well so we punt as a rule.
```suggestion
# @default -- See `values.yaml`
globalAppConfig:
```

rra (approved):
Just one documentation nit.

PR #3743, opened 2024-10-01
DM-46493: Remove SIAv2 application from phalanx

stvoutsin:
**Summary**
This PR deletes the old SIAv2 application from phalanx and removes all instances were it appears in the docs.
There will be a new SIA application (currently in development) which will replace this one, which is no longer supported.


stvoutsin:
SIAv2 was enabled on USDF. Was it working and/or used there? We should confirm before merging this in

PR #3756, opened 2024-10-03
[DM-46593] Summit maintenance window Oct 04

afausti:
These changes adjust resources requests and limits and connector configuration on TTS.
For the Telegraf connector configuration reducing max_batch_size and increasing flush_interval fixed the issues we were seeing.
I also added the Connectors dashboard there.


athornton (approved):
LGTM

PR #3771, opened 2024-10-09
Update consdb secrets in TTS

Vebop:
Add consdb to Sasquatch's kafka users. Reorganize secrets for consdb. This first step was meant to create consdb on TTS, with the same configuration as on Summit as far as we can. The next steps will be to upgrade consdb, hinfo, pqserver to the most recent commits and test that they work as expected with the correct alembic migration already applied. (At TTS then at Summit)

### DIFF
--- a/applications/consdb/secrets.yaml
+++ b/applications/consdb/secrets.yaml
@@ -4,14 +4,24 @@ consdb-password:
   copy:
     application: sasquatch
     key: consdb-password
+exposurelog-password:
+  description: >-
+    PostgreSQL password for the exposurelog user exposurelog database.
+  copy:
+    application: exposurelog
+    key: exposurelog_password
 oods-password:
   description: >-
     PostgreSQL password for the OODS user Butler database.
 lfa-password:
   description: >-
     LFA password
-exposurelog-password:
-  description: "Password for the TTS where we use exposurelog database."
   copy:
-    application: exposure-log
-    key: exposurelog_password
+    application: auxtel
+    key: aws-secret-access-key
+lfa-key:
+  description: >-
+    LFA key
+  copy:
+    application: auxtel
+    key: aws-access-key-id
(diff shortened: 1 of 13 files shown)


mareuter (commented):
For the `db` entry, why do all of the other sites use the `oods` user except TTS which is using `exposurelog`? Are the other sites going to convert eventually?

Vebop:
@ktlim Might have to help me answer Michael's question, but I believe it was because we didn't have access to the oods password and user initially when adding consdb to the existing postgres on TTS. I think also the existing postgres schema were exposurelog.

ktlim:
Actually, we didn't have access to the `postgres` superuser, which I think would have been necessary to grant access to `oods` to the `cdb` schemas (although I must admit I didn't try doing a grant to `oods` using the `exposurelog` role).

applications/consdb/secrets.yaml
```diff
@@ -4,14 +4,24 @@ consdb-password:
...
 lfa-password:
   description: >-
     LFA password
-exposurelog-password:
-  description: "Password for the TTS where we use exposurelog database."
   copy:
-    application: exposure-log
-    key: exposurelog_password
+    application: auxtel
+    key: aws-secret-access-key
```
ktlim, line 21: We're sure this is the read-only key, not the read-write one, correct?

ktlim (approved):

applications/consdb/secrets.yaml
Vebop: I am more sure that this is not the read only key, in auxtel/secrets.yaml it mentions that these are for writing to the lfa.

Vebop: So I should switch it back to lfa-password with no copy directive, and get the password from 1password to give to the admins to put in the vault?

mareuter: Yes, that is the read/write key. Do we have a read only key for the TTS LFA?

Vebop: In 1Pass I only find what looks to be the same as what auxtel has, based on the names 'aws-secret-access-key(-id)'

PR #3787, opened 2024-10-17
Tickets/dm 46557 upgrade cons db

Vebop:
Add temp, and astropy directory, update consdb to 24.10.1 tag. This include multicolumn primary keys. Tested on TTS and Summit with LATISS.

### DIFF
--- a/applications/consdb/values-tucson-teststand.yaml
+++ b/applications/consdb/values-tucson-teststand.yaml
@@ -25,7 +25,7 @@ pq:
 resources:
   requests:
     cpu: 200m
-    memory: 80Gi
+    memory: 20Gi
   limits:
     cpu: 500m
-    memory: 80Gi
+    memory: 20Gi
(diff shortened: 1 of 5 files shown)


applications/consdb/values-tucson-teststand.yaml
```diff
@@ -8,15 +8,24 @@ lfa:
...
     enable: false
-    tag: "tickets-DM-44551"
+    tag: "24.10.1"
+
 pq:
   image:
-    tag: "main"
+    tag: "24.10.1"
+
+resources:
```
bbrondel, line 25: Are these good choices for resource allocation?

bbrondel: Certainly the memory seems overkill for pqserver. And the request memory seems excessive for both. What's the size of the containers?

Vebop: I would almost just remove these as we have not done a thorough review of the resources we need. I remember we added these when debugging because of seeing no space left on device while on TTS.

Vebop:
image

PR #3790, opened 2024-10-17
DM-46934: Set the TTL after finishing to 0.

ktlim:


mareuter (approved):
Thank you for the changes. Looks good to me.

PR #3794, opened 2024-10-18
DM-45906: Enable SIA over Butler application in phalanx

stvoutsin:
**Summary**
This PR adds a new application to phalanx called "sia". This is a separate app to the existing CADC "siav2" application, which implements the same IVOA Simple Image Access protocol but using the butler and dax_obscore packages:

The SIAv2 app is another Safir-FastAPI application, and can be deployed on top of a number of Butler Repositories, either Direct or Remote. There is some additional complexity added to enable setting up the pg secrets in case of use with a Postgres database as the Butler repo db server.

The SIA application is developed here:
https://github.com/lsst-sqre/sia

And the relevant butler & obscore packages:
- https://github.com/lsst/daf_butler
- https://github.com/lsst-dm/dax_obscore

**How was this tested?**
- SIA Validators
The app was run through two separate external validation services (using a sample repo without data):
      - Paris VO: https://voparis-validator.obspm.fr/index.php?
      - EuroVO: https://registry.euro-vo.org/evor/

- Example Notebook
https://github.com/lsst-sqre/sia/blob/u/stvoutsin/example-notebook/examples/SIAv2_Example.ipynb

- Manual Queries via curl
Example: https://data-dev.lsst.cloud/api/sia/dp02/query?POS=CIRCLE+55.7467+-32.2862+0.05&time=60550.31803461111+60550.31838182871

### DIFF
--- a/applications/sia/secrets.yaml
+++ b/applications/sia/secrets.yaml
@@ -1,0 +1,20 @@
+"aws-credentials.ini":
+  copy:
+    application: nublado
+    key: "aws-credentials.ini"
+"butler-gcs-idf-creds.json":
+  copy:
+    application: nublado
+    key: "butler-gcs-idf-creds.json"
+"postgres-credentials.txt":
+  copy:
+    application: nublado
+    key: "postgres-credentials.txt"
+slack-webhook:
+  description: >-
+    Slack web hook used to report internal errors to Slack. This secret may be
+    changed at any time.
+  if: config.slackAlerts
+  copy:
+    application: mobu
+    key: app-alert-webhook
--- a/applications/sia/secrets.yaml
+++ b/applications/sia/secrets.yaml
@@ -1,12 +1,15 @@
 "aws-credentials.ini":
+  if: config.directButlerEnabled
   copy:
     application: nublado
     key: "aws-credentials.ini"
 "butler-gcs-idf-creds.json":
+  if: config.directButlerEnabled
   copy:
     application: nublado
     key: "butler-gcs-idf-creds.json"
 "postgres-credentials.txt":
+  if: config.directButlerEnabled
   copy:
     application: nublado
     key: "postgres-credentials.txt"
--- a/applications/sia/templates/deployment.yaml
+++ b/applications/sia/templates/deployment.yaml
@@ -1,0 +1,109 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: "sia"
+  labels:
+    {{- include "sia.labels" . | nindent 4 }}
+spec:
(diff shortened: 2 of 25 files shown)


applications/sia/secrets.yaml
```diff
@@ -0,0 +1,20 @@
...
+    application: nublado
+    key: "aws-credentials.ini"
+"butler-gcs-idf-creds.json":
+  copy:
+    application: nublado
+    key: "butler-gcs-idf-creds.json"
+"postgres-credentials.txt":
+  copy:
+    application: nublado
+    key: "postgres-credentials.txt"
```
rra, line 12: It would be better if we could make these all conditional on enabling support for local Butler. I think that can't be done right now given the layout of `values.yaml` because this is a property of each individual repository, so maybe we could disable local Butler support by default and add a switch to enable it just if we wanted to test it?

I want to make sure we don't give the SIA service those secrets during normal operation, since we want to make sure we're only using remote Butler.

applications/sia/templates/deployment.yaml
```diff
@@ -0,0 +1,109 @@
...
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - "all"
+          volumeMounts:
+            - name: "secrets"
+              mountPath: "/etc/butler/secrets"
+            - name: "secrets-raw"
+              mountPath: "/tmp/secrets-raw"
```
rra, line 46: Likewise here and a few other places in the service definition, all of this complexity should be conditional on whether we are supporting local Butler and disabled by default.

applications/sia/templates/ingress-anonymous.yaml
```diff
@@ -0,0 +1,43 @@
...
+                  name: {{ template "sia.fullname" . }}
+                  port:
+                    number: 8080
+            - path: "{{ .Values.ingress.path }}/.+/availability"
+              pathType: "Exact"
+              backend:
+                service:
+                  name: {{ template "sia.fullname" . }}
+                  port:
+                    number: 8080
```
rra, line 43: Technically you can't use `Exact` with regexes. NGINX figures out what you mean and does the right thing anyway, but these should have a path type of `ImplementationDefined`. Also I think you're technically supposed to add an ingress annotation saying to enable regex processing.

rra (approved):
Let's make the local Butler support disabled by default, and I noticed one other nit. Otherwise, looks good to me!

stvoutsin:
I've added an **_directButlerEnabled_**  boolean to the config, and all the direct butler specific features are hopefully now dependent on that being true.

PR #3802, opened 2024-10-21
DM-46984: Update Butler server for new embargo repo

dhirving:
Update the Butler server for the recent change in Butler repositories at USDF.  The new embargo repo is using a different S3 profile and configuration file than the old one.  SQRE is setting up an SIAv2 service that needs to access data from the new repo.

Also removed `ir2` from IDF -- this repository is no longer used.


applications/butler/templates/deployment.yaml
```diff
@@ -52,6 +52,10 @@ spec:
               value: "/opt/lsst/butler/secrets/butler-gcs-creds.json"
             - name: S3_ENDPOINT_URL
               value: {{ .Values.config.s3EndpointUrl | quote }}
+            {{- range $name, $url := .Values.config.additionalS3EndpointUrls }}
+            - name: LSST_RESOURCES_S3_PROFILE_{{ $name }}
```
stvoutsin, line 56: Would it make sense to change `$name` to  `$name | upper ` or if this is parsed elsewhere feel free to ignore

stvoutsin (approved):
Looks good to me!

applications/butler/templates/deployment.yaml
dhirving: Yeah it's kind of a goofy setup but it's the case-sensitive name of the profile so it needs to be in lowercase.

PR #3808, opened 2024-10-22
Sync from ccin2p3

gabrimaine:


applications/gafaelfawr/values-ccin2p3.yaml
```diff
@@ -14,6 +14,9 @@ config:
 
   # github:
   #   clientId: ae314e45a6af43ea910a
+  metrics:
+    application: gafaelfawr
+    enabled: false
```
rra, line 19: This should be the default with 12.0.1 so you can probably drop this again.

rra (approved):

rra:
Oh, and to be clear, feel free to click the merge button whenever you're ready. If you can't self-merge let me know; you should be able to I think.

PR #3823, opened 2024-10-28
DM-47199: `idfprod` Sasquatch

fajpunk:


applications/sasquatch/values-idfprod.yaml
```diff
@@ -0,0 +1,177 @@
...
+  #   source:
+  #     bootstrapServer: sasquatch-dev-kafka-bootstrap.lsst.cloud:9094
+  #     topicsPattern: "registry-schemas, lsst.sal.*, lsst.dm.*"
+  #   resources:
+  #     requests:
+  #       cpu: 2
+  #       memory: 4Gi
+  #     limits:
+  #       cpu: 4
+  #       memory: 8Gi
```
fajpunk, line 31: Not needed since we're not replicating

fajpunk, line 31: Not needed since we're not replicating

applications/sasquatch/values-idfprod.yaml
```diff
@@ -0,0 +1,177 @@
...
+  kraft:
+    enabled: true
+  kafkaController:
+    enabled: true
+    resources:
+      requests:
+        # memory: 16Gi
+        memory: 8Gi
+        # cpu: "2"
+        cpu: "1"
```
fajpunk, line 54: For this and other resource limits and requests, I don't think this needs to be as big as the other installations given that it's only going to handle app metrics.

applications/sasquatch/values-idfprod.yaml
```diff
@@ -0,0 +1,177 @@
...
+    GENERIC_NAME: "OIDC"
+    GENERIC_AUTH_URL: https://data.lsst.cloud/auth/openid/login
+    GENERIC_TOKEN_URL: https://data.lsst.cloud/auth/openid/token
+    USE_ID_TOKEN: 1
+    JWKS_URL: https://data.lsst.cloud/.well-known/jwks.json
+    GENERIC_API_URL: https://data.lsst.cloud/auth/openid/userinfo
+    GENERIC_SCOPES: openid
+    GENERIC_API_KEY: sub
+    PUBLIC_URL: https://data.lsst.cloud/
+    # STATUS_FEED_URL: https://raw.githubusercontent.com/lsst-sqre/rsp_broadcast/main/jsonfeeds/idfint.json
```
fajpunk, line 171: Should I make an `idfprod.json` file in `rsp_broadcast`?

fajpunk:
@afausti This is successfully synced, thanks for your help! I've left comments in for what I've changed from the `idfint` config. If this looks good to you, I'll remove the commented-out config before merging.

applications/sasquatch/values-idfprod.yaml
afausti: Sounds good, we can start with that.

applications/sasquatch/values-idfprod.yaml
afausti: Yes, that's the mechanism we use to populate the Chronograf news feed section, in the first page after you log into Chronograf. Can be used for notifications about this deployment.

afausti (approved):
Looks good to me!

PR #3829, opened 2024-10-29
DM-47220: Update exposurelog to point to new embargo rack

dhirving:
For the `usdfdev` environment, the S3 endpoint was configured incorrectly.  It was pulling in the config from the old embargo Butler repo instead of the new one.


dhirving:
FYI @sebastian-aranda -- I went ahead and deployed this to `usdfdev`.

PR #3833, opened 2024-10-30
Update sync-secrets.rst

ac6y:
This *appears* to be a typo.  Assuming that by default, obsolete secrets are *not* deleted, and to delete them the user must explicitly add the `--delete` flag to the `phalanx sync secrets`, then the word "not" should be added here.

The code indeed appears to require the `--delete` flag in order to trigger the `self._clean_vault_secrets()` method call:

https://github.com/lsst-sqre/phalanx/blob/main/src/phalanx/services/secrets.py#L243-L297

### DIFF
--- a/docs/admin/sync-secrets.rst
+++ b/docs/admin/sync-secrets.rst
@@ -23,7 +23,7 @@ It can then be run again whenever the secrets for that environment change.
 Deleting secrets
 ================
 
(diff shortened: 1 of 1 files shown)


rra (approved):
Thanks! That was an unfortunate missing word.

PR #3878, opened 2024-11-12
DM-47529: Migrate Prompt Processing PipelinesConfig to YAML

kfindeisen:
This PR replaces the bespoke pipelines config format with an extensible YAML-based format. The YAML-based config is still loaded into an environment variable; while a ConfigMap would be more convenient, we need the config to be fixed for a particular revision and that's hard to guarantee with a mapped file.

This PR must be merged together with lsst-dm/prompt_processing#222, which makes the corresponding changes to the parser.


erinleighh (approved):
Everything looks good! And just in case you need the reminder, don't forget to update the tag. Thank you!

kfindeisen:
Sorry, what tag are you referring to?

kfindeisen:
Clarified on Slack; it was a reminder not to commit the branch image tag.

kfindeisen:
Rebased onto #3894; I've double-checked that the new config is equivalent to that one.

PR #3894, opened 2024-11-19
Update the survey to BLOCK-320 for ComCam Prompt Processing

hsinfang:
`BLOCK-T248` and `PP-SURVEY` is no longer expected.   `BLOCK-320` 

< performs the observations for the individual visits that are queued up by the Feature Based Scheduler. BLOCK-320 is expected to be the standard json block for taking science images with ComCam going forward

### DIFF
--- a/applications/prompt-proto-service-lsstcomcam/values-usdfprod-prompt-processing.yaml
+++ b/applications/prompt-proto-service-lsstcomcam/values-usdfprod-prompt-processing.yaml
@@ -22,25 +22,37 @@ prompt-proto-service:
       # BLOCK-T60 is optics alignment
       # BLOCK-T75 is giant donuts
       # BLOCK-T88 is optics alignment
+      # BLOCK-T215 is evening twilight flats
+      # BLOCK-T216 is morning twilight flats
+      # BLOCK-T219 is pretty picture
       # BLOCK-T246 is instrument checkout
       # BLOCK-T249 is AOS alignment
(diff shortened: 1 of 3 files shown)


applications/prompt-proto-service-lsstcomcam/values-usdfprod-prompt-processing.yaml
```diff
@@ -22,25 +22,37 @@ prompt-proto-service:
       # BLOCK-T60 is optics alignment
       # BLOCK-T75 is giant donuts
       # BLOCK-T88 is optics alignment
+      # BLOCK-T215 is evening twilight flats
+      # BLOCK-T216 is morning twilight flats
+      # BLOCK-T219 is first target test and should no longer be needed soon
```
kfindeisen, line 27: I don't know if I quite agree with this description -- the November 13 run seemed to be motivated by "we need more PR photos", so I think the block may have been co-opted as a generic "pretty pictures" script. But I can't find any documentation of that (or how/why it was run on the 13th).

kfindeisen (approved):
Looks good, just one comment about the comments.

applications/prompt-proto-service-lsstcomcam/values-usdfprod-prompt-processing.yaml
hsinfang: I was quite confused by some older slack messages that said they were going to remove the block in Oct but still used it last week. 

I modified the comment to be "pretty picture".

PR #3895, opened 2024-11-19
Update schema registry version to 7.7.1

afausti:
We are using the default version of the Schema Registry  deployed by the schema-registry-operator, which is now quite old. 
This was an oversight of my part and we should set the Schema Registry version as part of the Sasquatch configuration.


mareuter (approved):
Looks good to me.

PR #3901, opened 2024-11-19
tickets/DM-47661: Create S3 topics for Kafka

mareuter:


applications/sasquatch/charts/strimzi-kafka/values.yaml
```diff
@@ -263,6 +263,9 @@ users:
     # -- Enable user ts-salkafka, used at the telescope environments
     enabled: false
 
+    # -- Create topics for the ts-salkafka user not controlled by CSCs
```
afausti, line 266: Note that this is not generic enough to create an arbitrary topic, you still need to configure ts-salkafka access to the topic.

One option is to say in the comment to use topic prefix `lsst.s3` or make it more generic and configure the `topicPrefixes` here as well, like we do in the [rest-proxy](https://github.com/lsst-sqre/phalanx/blob/main/applications/sasquatch/charts/rest-proxy/values.yaml#L64) sub chart.

afausti (changes requested):
Looks good.  Depending on how you see this growing we should either make this more generic or indicate in the documentation for this option only topics with a given prefix are supported.

applications/sasquatch/charts/strimzi-kafka/values.yaml
mareuter: OK. I made the doc string a bit more specific.

afausti (approved):
Perfect thank you!

PR #3907, opened 2024-11-21
DM-47581: Add Sentry support for Gafaelfawr

rra:
Add optional Sentry support for Gafaelfawr, which will let us send traces, errors, and other telemetry to Sentry.


rra:
Created this with the wrong ticket number.

PR #3929, opened 2024-11-25
DM-47794: Update rubintv app version to v2.5.4 for summit and usdf production deployments

sebastian-aranda:


aibsen (approved):
change in rubintv image tag for production environments, looks ok.

PR #3947, opened 2024-12-03
DM-43082: Knative

dspeck1:
Adds knative platform and image notification kafka consumer offset.


kfindeisen (approved):
Looks good, assuming you left out the many Keda variables on purpose.

PR #3948, opened 2024-12-03
Update Prompt Processing to 4.10.0 and pull in configs from Argo

kfindeisen:
This PR updates our Prompt Processing version, and makes permanent some survey and resource settings that have already been deployed through Argo.

### DIFF
--- a/applications/prompt-proto-service-lsstcomcam/values-usdfprod-prompt-processing.yaml
+++ b/applications/prompt-proto-service-lsstcomcam/values-usdfprod-prompt-processing.yaml
@@ -25,8 +25,6 @@ prompt-proto-service:
           - ${PROMPT_PROCESSING_DIR}/pipelines/LSSTComCam/ApPipe.yaml
           - ${PROMPT_PROCESSING_DIR}/pipelines/LSSTComCam/SingleFrame.yaml
           - ${PROMPT_PROCESSING_DIR}/pipelines/LSSTComCam/Isr.yaml
-        - survey: BLOCK-T60  # optics alignment
-          pipelines: []
         - survey: BLOCK-T75  # giant donuts
           pipelines: []
         - survey: BLOCK-T88  # optics alignment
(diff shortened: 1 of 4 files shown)


applications/prompt-proto-service-lsstcomcam/values-usdfprod-prompt-processing.yaml
```diff
@@ -35,27 +33,37 @@ prompt-proto-service:
...
           pipelines: []
         - survey: BLOCK-T250  # TMA daytime checkout
           pipelines: []
+        - {survey: BLOCK-T306, pipelines: []}  # Manual observations during SV block
+        - {survey: BLOCK-T307, pipelines: []}  # Manual observations during SV block
+        - {survey: BLOCK-T308, pipelines: []}  # Manual observations during SV block
+        - {survey: BLOCK-T309, pipelines: []}  # Manual observations during SV block
+        - {survey: BLOCK-T310, pipelines: []}  # Manual observations during SV block
+        - {survey: BLOCK-T311, pipelines: []}  # Manual observations during SV block
+        - {survey: BLOCK-T312, pipelines: []}  # Manual observations during SV block
```
hsinfang, line 48: Do we want to add `BLOCK-T318`, just because it might stay around for a bit longer?

hsinfang (approved):
Do we want to formally change the pod count, preload padding, worker restart frequency here too?

applications/prompt-proto-service-lsstcomcam/values-usdfprod-prompt-processing.yaml
kfindeisen: Do you think it might? It looked like one of the usual "run it a few times until we're satisfied" type tests.

kfindeisen:
> Do we want to formally change the pod count, preload padding, worker restart frequency here too?

I hope I can _revert_ the pod count and worker restart patches once 4.10 is deployed!

Preload padding I'm less sure about, but I think the 120" was too cautious. Last night the largest slew error was 8", and on Dec 1 it was 2.4".

applications/prompt-proto-service-lsstcomcam/values-usdfprod-prompt-processing.yaml
hsinfang: I don't know, and indeed it looks like that type you described.  Either including it here or not is fine to me.

kfindeisen: Then I'll leave it out for now (though I guess we're running out of "for now"!).

PR #3953, opened 2024-12-04
Add a format warning to Prompt Processing values.

kfindeisen:
This PR clarifies for config editors under which circumstances a pipeline config will get corrupted. This prevents the issue introduced in #3948 and fixed in #3949.


erinleighh (approved):
The format warning is clear and easy to follow. Thank you!

PR #3958, opened 2024-12-06
DM-48004: Add resource limits for sia deployment

stvoutsin:
Added resources for SIAv2 application. This may need to be revisited after we get a better idea of what these need to be exactly, along with what row limits we will have and pod replicas etc. 
Currently there is a bug or feature which may be redesigned in the client/server butler which currently fetches and caches a lot more data then needed. So the current usage may not represent the actual expected usage in the future. In any case with things as they stand data-int is using 24 Gb of mem,

For this reason to start with I'm with 30 Gb of RAM same as the resource limits for TAP.

### DIFF
--- a/applications/sia/values.yaml
+++ b/applications/sia/values.yaml
@@ -66,7 +66,13 @@ podAnnotations: {}
 
 # -- Resource limits and requests for the sia deployment pod
 # @default -- See `values.yaml`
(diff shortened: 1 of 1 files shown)


athornton (approved):
LGTM at least pending further usage data.

PR #3979, opened 2024-12-12
DM-48116: Fix ingress/tls indentation issue with argo-cd values files

stvoutsin:
**Summary**
There was an indentation issue which showed up recently at the roe environment when they tried to synchronize. The symptoms were that argo-cd failed to synchronize, with error referring to ` letsencypt-dns`. Upon further investigation it seems that the `tls: false` was missplaced, and should be under `ingress` rather than `server`.

This PR addresses the issue by fixing the indentation to match the expected nesting of the ingress setting.

**How was this tested**
Tested on the ROE environments (test & prod) and it works, they can now synchronize argo-cd successfully

### DIFF
--- a/applications/argocd/values-ccin2p3.yaml
+++ b/applications/argocd/values-ccin2p3.yaml
@@ -21,5 +21,5 @@ argo-cd:
 
   server:
     ingress:
-      hostnmae: "data-dev.lsst.eu"
+      hostname: "data-dev.lsst.eu"
       tls: false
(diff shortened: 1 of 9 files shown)


rra (approved):

stvoutsin:
One more changed made, fixed typo in ccin2p3 argocd values file.

rra (approved):
Good catch, thanks!

PR #3989, opened 2024-12-13
DM-47389: mobu app metrics events

fajpunk:
Mobu and Sasquatch config to publish events created in [this mobu PR](https://github.com/lsst-sqre/mobu/pull/387)


applications/sasquatch/charts/app-metrics/values.yaml
```diff
@@ -13,7 +13,14 @@ globalAppConfig:
...
     influxTags:
-      - "type"
+      - "sync"
+      - "success"
+      - "flock"
+      - "business"
+      - "username"
+      - "notebook"
+      - "repo"
+      - "repo_ref"
```
rra, line 23: I would leave this off, particularly if we go with adding the hash of the repo, but even if not. I don't think we're going to routinely run checks from multiple refs from the same repo in the same environment in circumstances where we care about metrics. I'm betting the only values will be whatever mobu runs by default and then a totally random variety of development branches from the GitHub verification.

rra (approved):

applications/sasquatch/charts/app-metrics/values.yaml
fajpunk: I ditched `repo_ref`, and added `repo_hash`, but [you convinced me that it doesn't need to be a tag](https://github.com/lsst-sqre/mobu/pull/387#discussion_r1887672199).

PR #3997, opened 2024-12-16
DM-48144: Add config files to BTS for EAS and sensor CSCs

aibsen:


aibsen:
Add config files to BTS for EAS and sensor CSCs

PR #4010, opened 2024-12-17
DM-48184: Bump Wobbly to 0.2.0, add maintenance CronJob

rra:
Upgrade Wobbly to 0.2.0 and add a maintenance `CronJob` that runs `wobbly expire` to delete all jobs that have passed their expiration times. Use the same requests and limits as the main Wobbly pod for simplicity.

Delete a stray environment variable from the main Wobbly deployment that was left over from cut and paste.

### DIFF
--- a/applications/wobbly/templates/cronjob-maintenance.yaml
+++ b/applications/wobbly/templates/cronjob-maintenance.yaml
@@ -1,0 +1,164 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: "wobbly-maintenance"
+  labels:
+    {{- include "wobbly.labels" . | nindent 4 }}
+spec:
+  schedule: {{ .Values.maintenance.schedule | quote }}
+  concurrencyPolicy: "Forbid"
+  jobTemplate:
+    spec:
+      activeDeadlineSeconds: {{ .Values.maintenance.deadlineSeconds }}
+      ttlSecondsAfterFinished: {{ .Values.maintenance.cleanupSeconds }}
+      template:
+        metadata:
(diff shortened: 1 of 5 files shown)


applications/wobbly/templates/cronjob-maintenance.yaml
```diff
@@ -0,0 +1,164 @@
...
+            {{- if .Values.cloudsql.enabled }}
+            - name: "cloud-sql-proxy"
+              # Running the sidecar as normal causes it to keep running and
+              # thus the Pod never exits, the Job never finishes, and the hook
+              # blocks the sync. Have the main pod signal the sidecar by
+              # writing to a file on a shared emptyDir file system, and use a
+              # simple watcher loop in shell in the sidecar container to
+              # terminate the proxy when the main container finishes.
+              #
+              # Based on https://stackoverflow.com/questions/41679364/
```
fajpunk, line 43: Could we use a [sidecar container :tm: ](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/#jobs-with-sidecar-containers) for this? I think they're available  in k8s >= 1.29, which .....

Looks like usdf prod at least is 1.27 :cry: so nevermind.

fajpunk (approved):

applications/wobbly/templates/cronjob-maintenance.yaml
rra: Ooo, I didn't know about those! Those will simplify a lot of patterns with Cloud SQL. As soon as we have 1.29 everywhere, I will definitely switch.

PR #4023, opened 2024-12-19
DM-48143: Enable new SIAv2 over Butler application for usdf

stvoutsin:
Changes include adding configurations for the usdf environments for the new SIAv2 over Butler application which will replace the previous CADC-based SIAv2 application & change to the usdfdev environment to enable it there.

### DIFF
--- a/applications/sia/values-usdfdev.yaml
+++ b/applications/sia/values-usdfdev.yaml
@@ -1,0 +1,10 @@
+config:
+
+  # Data (Butler) Collections
+  butlerDataCollections:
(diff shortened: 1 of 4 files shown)


dhirving (approved):

stvoutsin:
The SIAv2 chart has been updated to 0.1.5 which includes the bugfixes to dax_obscore and daf_butler. We are testing on dev & int after which I assume we are safe to merge this in and deploy on usdfdev whenever?

dhirving:
Yep, deploying to `usdfdev` should be fine.

PR #4054, opened 2025-01-08
DM-48364: mobu - update docs

fajpunk:


applications/mobu/tests/github_ci_app_enabled_test.yaml
```diff
@@ -52,6 +52,7 @@ tests:
           path: "data['config.yaml']"
           value: |
             autostart: []
+            available_services: []
```
rra, line 55: Here and elsewhere, this should be `availableServices` to match the Helm case convention (as with `githubCiApp`).

applications/mobu/values.yaml
```diff
@@ -22,6 +22,11 @@ ingress:
   annotations: {}
 
 config:
+  # -- Which applications (tap, butler, etc.) are available in this environment.
+  # Notebooks can specify a `mobu.required_services` list in their metadata, and mobu will only run them if all services in that list are in this `available_services` list.
+  # See https://mobu.lsst.io/user_guide/in_repo_config.html#service-specific-notebooks
```
rra, line 27: Should wrap at 80 columns. The doc string uses Markdown, so the link should be a Markdown link (otherwise the whole URL will be added to the table and will probably break the formatting).

rra (commented):
I think we talked about this, but I don't remember the result: I think we decided that we were going to inject this value via Argo CD based on the environment configuration so that, eventually, no one needs to set this manually, right?

fajpunk:
> I think we talked about this, but I don't remember the result: I think we decided that we were going to inject this value via Argo CD based on the environment configuration so that, eventually, no one needs to set this manually, right?

Yep, ideally no one has to set this manually, but I didn't think we decided exactly how we were going to do it. Is it possible to do it from the ArgoCD environment now? Other than doing something like parsing the `environments/` values yaml files somehow?

rra (approved):

applications/mobu/values.yaml
```diff
@@ -22,18 +22,24 @@ ingress:
...
+  # [here](https://mobu.lsst.io/user_guide/in_repo_config.html#service-specific-notebooks)
+  availableServices: []
+
   # -- Autostart specification. Must be a list of mobu flock specifications.
   # Each flock listed will be automatically started when mobu is started.
   autostart: []
 
   # -- Configuration for the GitHub refresh app integration.
-  # See https://mobu.lsst.io/operations/github_refresh_app.html#add-phalanx-configuration
+  # See [here](https://mobu.lsst.io/operations/github_refresh_app.html#add-phalanx-configuration)
```
jonathansick, line 37:
```suggestion
  # See [the Mobu documentation](https://mobu.lsst.io/operations/github_refresh_app.html#add-phalanx-configuration)
```

jonathansick (approved):
It's besides this review but I noticed that Mobu has two GitHub Apps now. Did you consider having one app and putting either a configuration file in the target repos or configuration in Phalanx saying what operations are enabled on that repo?

fajpunk:
> Did you consider having one app and putting either a configuration file in the target repos or configuration in Phalanx saying what operations are enabled on that repo?

A few reasons, not all of them particularly convincing :)
* We never want the CI functionality enabled in production environments. This could be configured in Phalanx though.
* Different repos may want to do their CI in different environments.
  * The CI environments could be configured in an in-repo config file, but then Mobu would have to clone the repo and check the config file for every `check_run` and `check_suite` event. That would probably be fine
  * The repos that an env handles CI for could be configured in Phalanx. This could be a slight pain for new repos, but maybe is more in the spirit of using Phalanx as the source of truth for platform configuration.

frossie:
> I think we talked about this, but I don't remember the result: I think we decided that we were going to inject this value via Argo CD based on the environment configuration so that, eventually, no one needs to set this manually, right?

I think this is a placeholder pending service discovery?

fajpunk:
>> I think we talked about this, but I don't remember the result: I think we decided that we were going to inject this value via Argo CD based on the environment configuration so that, eventually, no one needs to set this manually, right?

> I think this is a placeholder pending service discovery?

Yep, and the Argo CD injection was done in [this PR](https://github.com/lsst-sqre/phalanx/pull/4057).

PR #4057, opened 2025-01-08
DM-48365: mobu - Inject availableServices in the ArgoCD Application

fajpunk:
In `idfdev`:

image

### DIFF
--- a/environments/templates/_helpers.tpl
+++ b/environments/templates/_helpers.tpl
@@ -4,3 +4,11 @@
     {{- if $enabled }}@{{ $app }}{{ end }}
   {{- end }}
 {{- end }}
+
+{{- define "enabledServicesYamlList" }}
+{{- range $app, $enabled := .Values.applications }}
+  {{- if $enabled }}
+- {{ $app | quote }}
+  {{- end }}
+{{- end }}
+{{- end -}}
--- a/environments/templates/applications/infrastructure/mobu.yaml
+++ b/environments/templates/applications/infrastructure/mobu.yaml
@@ -31,4 +31,8 @@ spec:
       valueFiles:
         - "values.yaml"
         - "values-{{ .Values.name }}.yaml"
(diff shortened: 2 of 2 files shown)


environments/templates/_helpers.tpl
```diff
@@ -4,3 +4,13 @@
...
+
+{{- define "enabledServicesYamlList" -}}
+{{- $enabledServicesList := list -}}
+{{- range $app, $enabled := .Values.applications -}}
+  {{- if $enabled }}
+  {{- $enabledServicesList = append $enabledServicesList (quote $app) -}}
+  {{- end -}}
+{{- end -}}
+[ {{ join ", " $enabledServicesList }} ]
+{{- end -}}
```
rra, line 16:
```suggestion
{{- define "enabledServicesYamlList" }}
{{- range $app, $enabled := .Values.applications }}
  {{- if $enabled }}
- {{ $app | quote }}
  {{- end }}
{{- end }}
{{- end -}}
```

environments/templates/applications/infrastructure/mobu.yaml
```diff
@@ -31,4 +31,7 @@ spec:
       valueFiles:
         - "values.yaml"
         - "values-{{ .Values.name }}.yaml"
+      valuesObject:
+        config:
+          availableServices: {{ include "enabledServicesYamlList" . }}
```
rra, line 36:
```suggestion
          availableServices:
            {{- include "enabledServicesYamlList . | nindent 12 }}
```

rra (approved):
Looks good, but I think the approach shown below might be a bit more natural.

environments/templates/_helpers.tpl
fajpunk: I usually try to avoid having to explicitly specify indent levels if I can, but I don't feel too strongly about it.

PR #4063, opened 2025-01-09
DM-48359: Enable authentication for exposurelog, narrativelog and nightreport on USDF deployments.

sebastian-aranda:


aibsen (approved):
Looks good to me

PR #4071, opened 2025-01-10
DM-46898: Reconfigure RA for ComCam on sky

mfisherlevine:


applications/rapid-analysis/templates/comcam-gather2aAos-set.yaml
```diff
@@ -0,0 +1,249 @@
...
+              subPath: {{ $values.subPath }}
+              {{- end }}
+          {{- end }}
+          {{- end }}
+          {{- end }}
+          {{- if $.Values.butlerSecret }}
+            - name: {{ $.Release.Name }}-butler-secret
+              mountPath: {{ $.Values.butlerSecret.containerPath }}
+          {{- end }}
+          {{- if or $.Values.resources $script.resources }}
```
ctslater, line 139: There are a bunch of these ".values or script" blocks (nodeSelector, affinity, tolerations, etc), which I think can each be replaced by something like:
```
{{- if or $.Values.resources $script.resources }}
{{- $resources := $script.resources | default $.Values.resources }}
resources:
  {{- toYaml $resources | nindent 12 }}
{{- end }}
```
I've tried this out and it seems to behave correctly for all the combinations I could come up with. Or you could even inline it as `{{- $script.resources | default $.Values.resources | toYaml | nindent 12}}`, but I haven't tested that one. Still requires the if statement as a guard around the resources block.

applications/rapid-analysis/templates/comcam-gather2aAos-set.yaml
```diff
@@ -0,0 +1,249 @@
...
+              cp -RL /secrets-raw/* /secrets
+              cat /secrets/aws-credentials.ini > new-aws-credentials.ini
+              printf "\n" >> new-aws-credentials.ini
+              cat /secrets-rubintv/aws_credentials.ini >> new-aws-credentials.ini
+              printf "\n" >> new-aws-credentials.ini
+              mv new-aws-credentials.ini /secrets/aws-credentials.ini
+              chown 73006:73006 /secrets/*
+              chmod 0600 /secrets/*
+          volumeMounts:
+            - name: {{ $.Release.Name }}-raw-butler-secret
```
ctslater, line 166: Why does this include `$.Release.Name`?

ctslater (approved):
One inline suggestion to tighten up some syntax, overall looks good.

PR #4084, opened 2025-01-14
Ignore other surveys in LATISS Prompt Processing

hsinfang:


applications/prompt-proto-service-latiss/values-usdfprod-prompt-processing.yaml
```diff
@@ -38,6 +38,8 @@ prompt-proto-service:
         # the daily calibration sequence as of May 27, 2024
         - {survey: BLOCK-295, pipelines: []}
         - {survey: "", pipelines: []}
+        # Ignore anything else
+        - {pipelines: []}
```
kfindeisen, line 42: Do you want to get rid of any of the explicit entries above?

kfindeisen (approved):
Thanks, I missed the `BLOCK-T40_HD34816` and `BLOCK-T40_HD74180` from last night. I was hoping AuxTel would be standardized enough that we wouldn't need this hack.

applications/prompt-proto-service-latiss/values-usdfprod-prompt-processing.yaml
eigerx: Hm, I feel like it might be safer to keep it, even though it's technically redundant. "Ignore anything without a pipeline" is a rule that I imagine could bite us later or need to become more specific. If we're confident on easily being able to find and add daily calibrations back in though then /shrug.

hsinfang: I'll delete `cwfs` and `cwfs-focus-sweep` that haven't happened since 2024-09-01 or longer.

PR #4090, opened 2025-01-15
DM-48438: Scale down Knative queue-proxy in Prompt Processing

kfindeisen:
This PR reduces the CPU and memory limits for Prompt Processing's `queue-proxy` container to their default levels. It doesn't change the ephemeral storage limits, which were already below their defaults.


dspeck1 (approved):
LGTM

PR #4096, opened 2025-01-16
Set Summit job lifetime to 2 hours.

ktlim:
This limits spam in the status channel from all the failed jobs.


mareuter (approved):
Looks good to me.

PR #4104, opened 2025-01-21
tickets/DM-48511: Add redis queue to sasquatch-backpack

Fireye04:


applications/sasquatch-backpack/templates/_helpers.tpl
```diff
@@ -24,3 +24,12 @@ Selector labels
...
 {{- end }}
+
+
+{{/*
+Common environment variables
+*/}}
+{{- define "sasquatch-backpack.envVars" -}}
+- name: "BACKPACK_REDIS_URL"
+  value: "redis://sasqautch-backpack-redis.{{ .Release.Namespace }}:6379/0"
+{{- end }}
```
afausti, line 35: Looks like this definition is not being used, so you can remove it.

applications/sasquatch-backpack/values-idfint.yaml
```diff
@@ -1,2 +1,44 @@
 config:
-  sasquatchRestProxyUrl: https://data-int.lsst.cloud/sasquatch-rest-proxy
+  sasquatchRestProxyUrl: "https://data-int.lsst.cloud/sasquatch-rest-proxy"
+
+redis:
+  persistence:
+    # -- Whether to persist Redis storage and thus tokens. Setting this to
+    # false will use `emptyDir` and reset all tokens on every restart. Only
+    # use this for a test deployment.
```
afausti, line 8: Make sure this comment reflects the usage of Redis for this application.

afausti (changes requested):
Looks good! Just a couple of fixes.

applications/sasquatch-backpack/values-idfint.yaml
Fireye04: The redis helm chart that's being used has that exact string in its documentation of the setting; Is that what it's doing under the hood?

https://github.com/lsst-sqre/charts/blob/3beda2d70e123ff9a9a2b6647b0af315828afa95/charts/redis/README.md?plain=1#L31

applications/sasquatch-backpack/templates/configmap.yaml
```diff
@@ -9,3 +9,4 @@ data:
   SASQUATCH_BACKPACK_PATH_PREFIX: {{ .Values.config.pathPrefix | quote }}
   SASQUATCH_BACKPACK_PROFILE: {{ .Values.config.logProfile | quote }}
   SASQUATCH_REST_PROXY_URL: {{ .Values.config.sasquatchRestProxyUrl | quote }}
+  BACKPACK_REDIS_URL: "redis://sasquatch-backpack-redis.sasquatch-backpack:6379/0"
```
Fireye04, line 12: Oh it appears I'm defining it here manually, should I be doing so in values.yaml then using a .Values reference to it instead?

applications/sasquatch-backpack/templates/_helpers.tpl
Fireye04: Oop no, I checked and It looks like I'm setting it elsewhere. https://github.com/lsst-sqre/phalanx/pull/4104#discussion_r1999560133

applications/sasquatch-backpack/templates/configmap.yaml
Fireye04: Did this :+1:

applications/sasquatch-backpack/values-idfint.yaml
afausti: It is just that we are not using Redis to store tokens in this context, so the Helm docs here could be:
``` Whether to persist Redis storage.  Setting this to false will use `emptyDir` which is not recommend in a production environment.```

File: applications/sasquatch-backpack/values-idfint.yaml [hunk 4]
```diff
@@ -1,2 +1,43 @@
 config:
-  sasquatchRestProxyUrl: https://data-int.lsst.cloud/sasquatch-rest-proxy
+    sasquatchRestProxyUrl: "https://data-int.lsst.cloud/sasquatch-rest-proxy"
```
afausti, line 2: I think you have an extra indentation here.

afausti (approved):
Looks good, just make sure you fix the extra indentation and then good to go!

PR #4115, opened 2025-01-22
[DM-48533] Make message key format configuration default

afausti:
The Control System Kafka topics now have a key that has DEAULT format.

The configuration `--message.keyFormat=DEFAULT` can be used to set this by default when displaying messages in the kafdrop UI. It doesn't take effect for Kafka topics that don't have a key.

This PR use the DEFAULT key format in all Sasquatch environments now.


mareuter (approved):
Looks good to me.

PR #4125, opened 2025-01-23
DM-48548: Update love-manager and love-frontend to revision 001 on summit.

sebastian-aranda:


aibsen (commented):
Update in revision tag for love in summit environment looks fine.

PR #4134, opened 2025-01-24
[DM-48545] REST proxy Kafka topics should now be created in Sasquatch

afausti:
We disabled topic auto-creation in Kafka, so we need to create the topics managed by the REST Proxy.
Clean up some old configuration we don't need anymore for the REST Proxy.


mareuter (approved):
Looks good to me.

mareuter:
One thing I didn't see was `lsst.cp` topics created. Is something else handling this?

afausti:
Yes, those are created by the client itself.

PR #4151, opened 2025-01-29
DM-46178: Use new query system in Prompt Processing

kfindeisen:
This PR removes a Prompt Processing debugging flag that was sometimes needed to get correct results with the old calib-handling code removed on lsst-dm/prompt_processing#249.


TallJimbo (approved):
I'm not really qualified to review Phalanx changes, but to the extent I can infer what this is doing, it looks fine.

PR #4157, opened 2025-01-30
Summit Sasquatch fixes

afausti:
After OS and k8s upgrades at the Summit we had problems restarting Kafka, we had to fix the nodeAffinity configuration for the kafka brokers and also pin the Kafka controller node IDs.

Also we had to rollback InfluxDB OSS after what looks like a bug on version 1.11.8


mareuter (approved):
Looks good to me.

PR #4172, opened 2025-02-06
DM-48733: times-square - New template render route

fajpunk:
A new endpoint was added to expose template rendering for a GitHub notebook template at a different path that includes the GitHub URL path for the notebook. This updates the template rendering `GafaelfawrIngress` to restrict rendering to only `lsst-sqre/nublado-seeds` repo notebooks.

Once this is merged, notebook-from-portal-query functionality won't work with any lab image that doesn't also have an `rsp-jupyter-extensions` package with [this change](https://github.com/lsst-sqre/rsp-jupyter-extensions/pull/42).

### DIFF
--- a/applications/times-square/templates/ingress-templates.yaml
+++ b/applications/times-square/templates/ingress-templates.yaml
@@ -16,6 +16,7 @@ template:
     name: "{{ template "times-square.fullname" . }}-templates"
     {{- with .Values.ingress.annotations }}
     annotations:
+      nginx.ingress.kubernetes.io/use-regex: "true"
       {{- toYaml . | nindent 6 }}
     {{- end }}
   spec:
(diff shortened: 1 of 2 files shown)


jonathansick:
So before we merge this we're going to drop commit [4f1030e](https://github.com/lsst-sqre/phalanx/pull/4172/commits/4f1030e4174fae2012d0e842efbadae8a62071f2), is that right?

I hadn't realized that this change is effectively going to break the templating feature in older scilat-lab images because we can't backport rsp-jupyter-extensions to them (right?). This is worth bringing up at a standup, but on the other hand, I can't see a way around this.

fajpunk:
Yep, the ingress changes are going to break any lab without the rsp extensions changes for non-admins, and I too can't see a way to do it without breaking old sciplat-lab images. This was just intended to test that the planned ingress changes would work, I'm not intending it to be merged yet. I'm going to make another PR to just bump Times Square.

rra:
Deleting this as stale.

PR #4187, opened 2025-02-10
Fix Sasquatch ingresses

afausti:
Nginx ingress 4.12 does not support regular expressions in the ingress path when pathType is set to Prefix. Use pathType ImplementationSpecific when possible and add a custom ingress for InfluxDB OSS.


mareuter (approved):
Looks good to me. One comment: Wouldn't it be better to just disable the original ingress? Once the upstream gets fixed we'll just have to put everything back or is the custom on meant to live on?

PR #4191, opened 2025-02-11
Enable Sentry for TAP

stvoutsin:
## Description

Enable Sentry error logging and tracing for TAP service. 
Enables for idfint & idfdev. If left disabled TAP services for other environments should continue to work without any noticeable change.

## Jira Issue

https://rubinobs.atlassian.net/browse/DM-48801

## Related PR in TAP service

https://github.com/lsst-sqre/lsst-tap-service/pull/127

## Tests

I've tested on an dev & int with SENTRY_DSN enabled, and then the same build with SENTRY_DSN disabled in phalanx to ensure that tracing doesn't get sent in the second case

## TODO Before merge

Set image of TAP service to new release before merging

### DIFF
--- a/applications/livetap/README.md
+++ b/applications/livetap/README.md
@@ -15,6 +15,7 @@ IVOA TAP service
 | cadc-tap.config.pg.database | string | `"lsstdb1"` | Postgres database to connect to |
 | cadc-tap.config.pg.host | string | `"mock-pg:5432"` (the mock pg) | Postgres hostname:port to connect to |
 | cadc-tap.config.pg.username | string | `"rubin"` | Postgres username to use to connect |
+| cadc-tap.config.sentryEnabled | bool | `false` | Whether Sentry is enabled in this environment |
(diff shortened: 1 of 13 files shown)


fajpunk:
Another useful env var to add is `SENTRY_RELEASE`, like [here in mobu](https://github.com/lsst-sqre/phalanx/pull/4210/files). This will group errors and traces into releases, so you can quickly see if errors are only occurring in a particular set of releases. The value can be whatever makes sense for a particular Deployment

stvoutsin:
> Another useful env var to add is `SENTRY_RELEASE`, like [here in mobu](https://github.com/lsst-sqre/phalanx/pull/4210/files). This will group errors and traces into releases, so you can quickly see if errors are only occurring in a particular set of releases. The value can be whatever makes sense for a particular Deployment

Oh nice I'll add that as well. I'm assuming the java sentry sdk should look for it in the env variables similar to the SENTRY_DSN.

PR #4194, opened 2025-02-12
DM-48589: Set up preliminary DP1 Butler in idfdev

dhirving:
Set up Butler server at `idfdev` to host a preliminary version of Data Preview 1, using Google Cloud SQL and Google Cloud Storage.

Unlike Data Preview 0.2, end users will not be given credentials to directly access the Butler database and artifacts.  Previously the credentials for Butler server had been shared with Nublado, but we now have separate secrets for the Butler server.  We will continue to share secrets with Nublado on the USDF environments indefinitely, so sharing is configurable per-environment.

Similarly the Data Preview 0.2 Butler configuration file was a public resource, but for Data Preview 1 it is internal to Butler server.  This required creating a separating configmap to separate the public files from private files.


applications/butler/templates/deployment.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     metadata:
       annotations:
         # Force the pod to restart when the config maps are updated.
-        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        checksum/config: {{ print (include (print $.Template.BasePath "/configmap-public.yaml") .) (include (print $.Template.BasePath "/configmap-private.yaml") .)| sha256sum }}
```
dhirving, line 18: I wasn't sure if there is a less verbose way to accomplish this concatenation of files in helm.

applications/butler/templates/ingress-anonymous.yaml
```diff
@@ -29,14 +29,16 @@ template:
             # configuration file is loaded.
             {{- range $repositoryLabel, $unused := .Values.config.repositories }}
             - path: "{{ $.Values.config.pathPrefix }}/repo/{{ $repositoryLabel }}/butler.yaml"
-              pathType: "Exact"
+              # These should be 'pathType: Exact', but newer versions of Kubernetes don't allow periods
```
rra, line 32: Nit: This bug is in ingress-nginx specifically, not Kubernetes in general.

rra (approved):
Looks great. Thank you!

PR #4195, opened 2025-02-13
DM-48733: Noteburst in `idfint` and `idfprod`

fajpunk:
* Enable Noteburst in `idfint` and `idfprod`
* Add the times-square URL to Squareone in `idfint` and `idfprod`

Note that the default Times Square scope is `exec:admin`, and the template api scope is `exec:notebook` in both envs already. This means that only admins will be able to execute notebooks. It does allow anyone with `exce:notebook` to render any templates though, until we [change the template render routes in the ingress](https://github.com/lsst-sqre/phalanx/pull/4172)

### DIFF
--- a/applications/squareone/values-idfprod.yaml
+++ b/applications/squareone/values-idfprod.yaml
@@ -7,6 +7,7 @@ config:
   siteName: "Rubin Science Platform"
   docsBaseUrl: "https://rsp.lsst.io"
   semaphoreUrl: "https://data.lsst.cloud/semaphore"
+  timesSquareUrl: "https://data.lsst.cloud/times-square/api"
   coManageRegistryUrl: "https://id.lsst.cloud"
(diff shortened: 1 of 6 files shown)


applications/squareone/values-idfprod.yaml
```diff
@@ -7,6 +7,7 @@ config:
   siteName: "Rubin Science Platform"
   docsBaseUrl: "https://rsp.lsst.io"
   semaphoreUrl: "https://data.lsst.cloud/semaphore"
+  timesSquareUrl: "https://data-int.lsst.cloud/times-square/api"
```
jonathansick, line 10:
```suggestion
  timesSquareUrl: "https://data.lsst.cloud/times-square/api"
```

jonathansick (approved):
Just one timesSquareUrl value to fix, but otherwise good to go.

PR #4204, opened 2025-02-18
Prompt Processing permission to test topic

dspeck1:
Add permission for prompt processing to test next visit job topic.  This is dev only and temporary until replication is setup from summit.


afausti (approved):
Looks good to me, this grants read permission for the prompt-processing kafka user to the test.next-visit topic.

PR #4240, opened 2025-02-24
DM-49125: Use USDF for serving DP02 Butler files

dhirving:
Configure the `idfdev` Butler server to use Weka S3 at USDF for serving Data Preview 0.2 files.


ktlim (approved):
The logic and usability (e.g. need to synchronize with endpoint URLs) of the ConfigMap template is rather complex, but it does seem to do what is wanted.  Consider a refactor in the future.

PR #4267, opened 2025-02-27
DM-49223: Prompt Processing Keda service can't load S3 credentials files

kfindeisen:
This PR fixes a Kubernetes bug in the Prompt Processing Keda service.

### DIFF
--- a/charts/prompt-keda/templates/scaled-job.yaml
+++ b/charts/prompt-keda/templates/scaled-job.yaml
@@ -181,7 +181,9 @@ spec:
           - name: s3-credentials-file
             secret:
               secretName: {{ template "prompt-keda.fullname" . }}-secret
-              key: s3-credentials-file
+              items:
+                - key: s3_credentials_file
+                  path: credentials
           {{- end }}
           {{- if .Values.registry.centralRepoFile }}
           - name: central-repo-file
--- a/charts/prompt-keda/templates/scaled-job.yaml
+++ b/charts/prompt-keda/templates/scaled-job.yaml
@@ -108,6 +108,10 @@ spec:
               - name: AWS_SHARED_CREDENTIALS_FILE
                 value: /app/s3/credentials
               {{- end }}
+              {{- if .Values.s3.aws_profile }}
+              - name: AWS_PROFILE
+                value: {{.Values.s3.aws_profile }}
+              {{- end }}
               - name: LSST_DB_AUTH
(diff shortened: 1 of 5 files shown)


hsinfang:
Notes: this PR is the Keda equivalent of https://github.com/lsst-sqre/phalanx/pull/4243

charts/prompt-keda/templates/scaled-job.yaml
```diff
@@ -108,6 +108,10 @@ spec:
               - name: AWS_SHARED_CREDENTIALS_FILE
                 value: /app/s3/credentials
               {{- end }}
+              {{- if .Values.s3.aws_profile }}
+              - name: AWS_PROFILE
+                value: {{.Values.s3.aws_profile }}
+              {{- end }}
```
kfindeisen, line 114: I don't understand the reason for this change or for #4243. Can't we just define the profile in the credentials file and the `s3:` URLs? Every extra config hook that needs to be synchronized with the others is a potential point of failure...

kfindeisen: That said, I've confirmed setting `AWS_PROFILE=embargo` fixes our credentials problems.

eigerx: I know this is independent, but just to confirm since or5 is coming up so soon: `embargo_or5` works too, correct?

eigerx: (although I suppose it is totally ok if it just looks at the credential file, so perhaps ignore this comment)

eigerx (approved):

charts/prompt-keda/templates/scaled-job.yaml
kfindeisen: There is no `embargo_or5` profile in our credentials file. I meant that it lets us connect to `s3://repo-test-users`.

hsinfang: I missed this yesterday.  
My thinking behind this was to allow an easier switch between using our `rubin-pp-dev*` buckets versus `embargo_or5`'s `repo-test*` buckets (which share the same credentials as `embargo` repo). 
We are using `-dev` deployment to run OR5.  If we had deployed `-prod` to run OR5, we wouldn't have needed this, because that could have separated the repo access need to `rubin-pp-dev*` buckets versus repo-test*` buckets.

PR #4270, opened 2025-02-27
DM-49148: Enable Noteburst and Times Square in usdfint

fajpunk:


fajpunk:
Converting to a draft for now because we may need to get bot users created in int, which took like 3 months last time

jonathansick:
I think you can use the same user set as for usdfdev and usdfprod; like a human user they can use any of the environments at the same time.

PR #4299, opened 2025-03-03
Add dropdown values for nublado dev/int

athornton:


applications/nublado/values-idfdev.yaml
```diff
@@ -2,6 +2,17 @@ controller:
...
+      release:
+        cutoffVersion: "27.0.0"
+      weekly:
+        number: 78
+      daily:
+        number: 25
+      releaseCandidate:
+        cutoffVersion: "28.0.0"
+      experimental:
+        number: 5
```
rra, line 15: I think I'd rather see us use an age cutoff for experimental since I don't think the count is very meaningful.

rra (approved):
Looks good to me with one preference note about experimental filtering.

applications/nublado/values-idfdev.yaml
athornton: Sure thing.  I'll do one month.

athornton: That's not a thing.  Three weeks.

PR #4301, opened 2025-03-04
DM-49306: Add an environment variable for newer boto.

ktlim:


hsinfang (approved):

hsinfang:
At Slack you mentioned `AWS_RESPONSE_CHECKSUM_VALIDATION` too.  Is it not necessary?

ktlim:
I don't think we've seen it be necessary; all of our other usage against embargo seems to be running fine without it.  It'll be easy to add if it turns out to be.

PR #4306, opened 2025-03-05
DM-49593: Move to NetApp Cloud volumes

rra:
Drop support for the `config.realm` configuration setting, since Gafaelfawr no longer supports configuring it separately. Update the CRD for `GafaelfawrIngress`. Add the new `config.allowSubdomains` setting.

Disable cookie authentication to giftless since it uses a separate hostname anyway and cookies would never have worked (and otherwise the ingress would now be rejected). Clean up its `values.yaml` file a bit and drop the `config.baseUrl` setting in `GafaelfawrIngress` that is now ignored.

Disable cookie authentication for the service endpoint to Wobbly, since services will never have cookies.


athornton:
I ended up taking this one over since it was what was in dev when I started doing the netapp thing and we stayed on the same branch.

athornton (approved):
LGTM but then it would (Russ said it was ok)

PR #4310, opened 2025-03-05
Remove all Kafka user quotas.

mareuter:


afausti (approved):
My pleasure to approve

PR #4319, opened 2025-03-06
Disable production-tools on idfint

rra:
This hasn't started successfully in a while and I believe the intent is to remove it from this environment.


ctslater (approved):
Looks good, thanks, and sorry I haven't cleaned this up sooner.

PR #4326, opened 2025-03-08
DM-49297: Enable metrics for SIAv2 application on idfint

stvoutsin:
Enables initial metrics for the SIAv2 application in Sentry

### DIFF
--- a/applications/sasquatch/charts/app-metrics/values.yaml
+++ b/applications/sasquatch/charts/app-metrics/values.yaml
@@ -26,6 +26,9 @@ globalAppConfig:
   nublado:
     influxTags:
       - "username"
+  sia:
+    influxTags:
+      - "username"
   wobbly:
     influxTags:
       - "service"
--- a/applications/sia/templates/deployment.yaml
+++ b/applications/sia/templates/deployment.yaml
@@ -98,12 +98,44 @@ spec:
             - name: "SIA_SENTRY_TRACES_SAMPLE_RATE"
               value: {{ .Values.config.sentryTracesSampleRate | quote }}
             {{- end }}
-          {{- if .Values.config.directButlerEnabled }}
+            {{- if .Values.config.metrics.enabled }}
+            - name: "KAFKA_BOOTSTRAP_SERVERS"
+              valueFrom:
+                secretKeyRef:
+                  name: "sia-kafka"
+                  key: "bootstrapServers"
+            - name: "KAFKA_SECURITY_PROTOCOL"
+              valueFrom:
+                secretKeyRef:
+                  name: "sia-kafka"
+                  key: "securityProtocol"
+            - name: "KAFKA_CLIENT_CERT_PATH"
+              value: "/etc/sia-kafka/user.crt"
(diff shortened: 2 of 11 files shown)


applications/sia/templates/deployment.yaml
```diff
@@ -98,6 +98,24 @@ spec:
...
+                secretKeyRef:
+                  name: "sia-kafka"
+                  key: "securityProtocol"
+            - name: "KAFKA_CLIENT_CERT_PATH"
+              value: "/etc/sia-kafka/user.crt"
+            - name: "KAFKA_CLIENT_KEY_PATH"
+              value: "/etc/sia-kafka/user.key"
+            - name: "KAFKA_CLUSTER_CA_PATH"
+              value: "/etc/sia-kafka/ca.crt"
+            {{- end }}
```
fajpunk, line 118: You're also going to need  [volumeMounts](https://github.com/lsst-sqre/phalanx/blob/f15cbcb4b6cfe2536d052b5bb70e5f75e1a2c002/applications/mobu/templates/deployment.yaml#L124-L137) and [volumes](https://github.com/lsst-sqre/phalanx/blob/f15cbcb4b6cfe2536d052b5bb70e5f75e1a2c002/applications/mobu/templates/deployment.yaml#L148-L152) entries for the Kafka keys and certs from the Kafka secret.

stvoutsin: Oh oops I never added those! Thanks, I've now added them

applications/sasquatch/charts/app-metrics/values.yaml
```diff
@@ -26,6 +26,9 @@ globalAppConfig:
   nublado:
     influxTags:
       - "username"
+  sia:
+    influxTags:
+      - "success"
```
fajpunk, line 31: You can get rid of this now that you split querying into separate success and failure events

fajpunk (approved):

applications/sasquatch/charts/app-metrics/values.yaml
stvoutsin: Actually I need to figure out what tags we want in this case, since as I understand it, adding tags here is required for enabling the sasquatch metrics.

 Since my events only record duration (and error in the case of failed queries) does it make sense to have a duration tag? Or perhaps should I add a username to the event and have that be the influxTag configured here?

fajpunk: I didn't think you needed to specify any tags if you don't want any... is it not applying correctly if you don't have any tags speficied?

fajpunk: Ahh I see, it's assuming there's an entry for ever app specied in `.Values.apps`. I can fix that...

stvoutsin: Ok, in the meantime I'm also adding a username to the event here:
https://github.com/lsst-sqre/sia/pull/34

So I may actually just add that as a tag for sia in sasquatch

PR #4330, opened 2025-03-10
DM-49411: Add the rebase_checker workflow

jonathansick:
This reusable workflow checks if the main branch has been merged into a PR's branch. This commonly happens when folks update their branch with the GitHub UI. The downside of this approach is that it makes Git history really hard to understand in a PR because commits from main are intermixed with new commits. This workflows encourages people to use the rebases, with a link to the Developer Guide.


rra (approved):
Looks good, thank you! I have no idea what failed last time. Maybe I was trying to make it mandatory and ran into problems with the merge queue?

jonathansick:
Hmm. I'll make it required and see if it merges ok.

jonathansick:
Yep, that's the issue: gsactions/commit-message-checker@v2 doesn't work in a merge group. I'll see if there's a way to skip the rule in a merge_group context.

jonathansick:
When I tried skipping the rebase checker from the `merge_group` trigger I found that the merge queue would just hang. So I'm stuck at the same place you got to, unfortunately.

rra:
Oh, right, that's what it was. I think we can make it optional but not mandatory. Making it optional would still at least send a warning to the person making the PR that it failed, and would warn reviewers, so that may be worth doing.

PR #4331, opened 2025-03-10
SP-1929: Adding Rubin rag

cyab05:
Create rubin-rag from gai-helpers and deploy on data-int.

### DIFF
--- a/applications/rubin-rag/.helmignore
+++ b/applications/rubin-rag/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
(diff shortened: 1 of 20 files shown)


JeremyMcCormick:
The application should really be called "rubin-rag" rather than "rubin_rag" because of conventions in API usage which discourage underscores. (Notice that all existing applications in phalanx use dashes rather than underscores for word separators in their names.)

cyab05:
Everything other than the GitHub repo itself should be named “rubin-rag”. Did I miss something?

JeremyMcCormick:
> Everything other than the GitHub repo itself should be named “rubin-rag”. Did I miss something?

Yes, I think so - api should be `/api/rubin-rag`, app should be `rubin-rag`, etc.

JeremyMcCormick (approved):
Looks good - this has been tested and is known to work on `usdf-rsp-dev` and `idf-dev`.

PR #4364, opened 2025-03-17
tickets/DM-48511: Fix redis values configuration

Fireye04:


applications/sasquatch-backpack/values.yaml
```diff
@@ -110,3 +110,10 @@ redis:
 
   # -- Affinity rules for the Redis pod
   affinity: {}
+
+cluster:
+  name: backpack
```
afausti, line 115: This is the name of the Kafka cluster that backpack uses, i.e "sasquatch"

applications/sasquatch-backpack/values.yaml
```diff
@@ -110,3 +110,10 @@ redis:
...
+
+strimzi:
+  enabled: true
```
afausti, line 118: I would just list the Kafka topics you want to create here, assuming that Strimzi is deployed.

applications/sasquatch-backpack/values.yaml
```diff
@@ -110,3 +110,10 @@ redis:
...
+  topics: [lsst.backpack.usgs-earthquake-data]
```
afausti, line 119: I suggest you  group all configuration related to Kafka for example:
```
kafka:
  cluster:
    name: sasquatch
  topics:
    - lsst.backpack.usgs-earthquake-data
 ```

File: applications/sasquatch-backpack/templates/topics.yaml [hunk 4]
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.strimzi.enabled }}
```
afausti, line 1: You can't really test this effectively here, so just assume Strimzi is deployed that that backpack depends on kafka being deployed.

applications/sasquatch-backpack/templates/topics.yaml
```diff
@@ -0,0 +1,15 @@
...
+{{- $cluster := .Values.cluster.name }}
```
afausti, line 2: With the suggested configuration below, this would become:
```
{{- $cluster := .Values.kafka.cluster.name }}
```

applications/sasquatch-backpack/templates/topics.yaml
```diff
@@ -0,0 +1,15 @@
...
+{{- range $topic := .Values.strimzi.topics }}
```
afausti, line 3: And this would become:
```
{{- range $topic := .Values.kafka.topics }}
```

applications/sasquatch-backpack/templates/topics.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: KafkaTopic
+metadata:
+  name: {{ $topic }}
+  labels:
+      strimzi.io/cluster: {{ $cluster }}
+spec:
+  replicas: 3
+  partitions: 1
+{{- end }}
+{{- end }}
```
afausti, line 15: You can remove the extra end too.

afausti (changes requested):
Creating a KafkaTopic resource for backpack is the right move.
I've made a few suggestions to improve the configuration.

applications/sasquatch-backpack/values.yaml
```diff
@@ -72,3 +72,48 @@ global:
...
+  # -- Tolerations for the Redis pod
+  tolerations: []
+
+  # -- Affinity rules for the Redis pod
+  affinity: {}
+
+kafka:
+  cluster:
+    name: sasquatch
+  enabled: true
```
afausti, line 117: I think this flag can be removed, I don't see where it is being used.

afausti (approved):
Is the `kafka.enabled` flag being used? Otherwise looks good!

afausti (changes requested):
I just realized that for this to work we would need to deploy the Strimzi topic operator in the sasquatch-backpack namespace.
I think I prefer specifying the KafkaTopic resource in Sasquatch itself, like we are doing with other applications that depend on Sasquatch.
Let's remove the KafkaTopic resource from backpack and I'll address this in DM-49745.
Sorry for the back and forth on this.

afausti (approved):
LGTM!

PR #4368, opened 2025-03-18
DM-49545: Change vault policies to read mount point from configuration

stvoutsin:
This PR modifies how we generate the vault policies, to use the mount point from the vault path in the environment configuration rather than a fixed "/secret" path.


rra (approved):
Great catch. Thank you!

PR #4375, opened 2025-03-18
DM-49361: mobu - multi-instance

fajpunk:
* The replica count from the helm chart values is templated into the `replicas` value of the Mobu workload, and the `MOBU_REPLICA_COUNT` environment variable.

* The Mobu workload is a `StatefulSet` instead of a `Deployment` so that the [pod index label](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-index-label) can be passed as the value to the `MOBU_REPLICA_INDEX` env var via the Kubernetes [Downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/). This ensures that monkeys can be distributed as evenly as possible among all of the replicas in the `StatefulSet`.

Goes with [this Mobu PR](https://github.com/lsst-sqre/mobu/pull/448).

### DIFF
--- a/applications/mobu/Chart.yaml
+++ b/applications/mobu/Chart.yaml
@@ -5,4 +5,4 @@ description: "Continuous integration testing"
 home: https://mobu.lsst.io/
 sources:
   - "https://github.com/lsst-sqre/mobu"
-appVersion: 15.0.0
+appVersion: 15.1.0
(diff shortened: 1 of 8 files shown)


applications/mobu/values-idfdev.yaml
```diff
@@ -23,66 +27,82 @@ config:
...
-      business:
-        type: "SIAQuerySetRunner"
+        type: "NotebookRunnerInfinite"
         options:
-          query_set: "dp02"
+          repo_url: "https://github.com/lsst-sqre/scaletest.git"
+          repo_ref: "no-code"
+          delete_lab: false
         restart: true
+    # - name: "dfuchs-test"
```
rra, line 46: Reminder to remove the commented-out bits before merging.

rra (approved):
I didn't know you could do that trick with `fieldRef`. Neat!

PR #4386, opened 2025-03-20
DM-49555: Expand diagnostic logs for Prompt Processing

kfindeisen:
This PR standardizes/normalizes the PP log settings, and adds some verbose-level logs that are useful for debugging Middleware problems.


hsinfang (approved):
Thanks for cleaning up and unifying it for all pp service instances!

PR #4390, opened 2025-03-21
DM-49610: Get back on upstream telegraf

fajpunk:
We are using a specific nightly build of telegraf for app metrics collection because it has our added support for Avro unions.

[1.34](https://github.com/influxdata/telegraf/releases/tag/v1.34.0) finally has [our Avro union support](https://github.com/influxdata/telegraf/pull/16272).


afausti (approved):
That's great, glad they released as planned.

PR #4396, opened 2025-03-21
DM-49624: Update to Nublado 8.8.0

rra:
Add configuration for where to redirect the user after a JupyterHub logout so that we send the user to the Gafaelfawr logout handler instead of creating a redirect loop inside JupyterHub.


athornton (approved):
LGTM

PR #4405, opened 2025-03-24
DM-49665: Adjust resource limits for consdb tapSchema container

JeremyMcCormick:
The consdb container wants to allocate 600 MB initially, reflected in the new request limit, and it seems to intermittently
 exceed the 1Gi limit when starting up, so this was increased to 2Gi. 

New settings were tested and confirmed to work on `usdf-rsp-dev`.


stvoutsin (approved):
Those seem reasonable to me

PR #4418, opened 2025-03-25
DM-49704: Enable new TAP user in Sasquatch

stvoutsin:
PR to add a new "tap" user in Sasquatch. Plan is to use this for the event-based TAP service that is in development.


afausti (approved):
Looks great!

PR #4420, opened 2025-03-26
[DM-49723] Initial configuration for the TAP service in Sasquatch 

afausti:
Add the qserv KafkaUser and create initial KafkaTopics in the `lsst.tap` namespace.

### DIFF
--- a/applications/sasquatch/values-idfdev.yaml
+++ b/applications/sasquatch/values-idfdev.yaml
@@ -141,10 +141,14 @@ rest-proxy:
   kafka:
     topics:
       - test.next-visit
+      - lsst.tap.job-run
+      - lsst.tap.job-delete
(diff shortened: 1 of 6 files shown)


applications/sasquatch/values-idfdev.yaml
```diff
@@ -139,10 +141,13 @@ rest-proxy:
   kafka:
     topics:
       - test.next-visit
+      - lsst.tap.job-run
+      - lsst.tap.job-status
```
stvoutsin, line 145: Should we also run lsst.tap.job-delete now?

afausti: Sure, I can add the `lsst.tap.job-delete`  topic as well.

PR #4426, opened 2025-03-26
DM-48669: Deploy a dev prompt processing service for LSSTCam

hsinfang:


applications/prompt-keda-lsstcam/values-usdfdev-prompt-processing.yaml
```diff
@@ -0,0 +1,43 @@
+prompt-keda:
+
+  image:
+    pullPolicy: Always
+    # Overrides the image tag whose default is the chart appVersion.
+    tag: DM-48669-d_2025_03_24
```
eigerx, line 6: I think we'll want to change this back to latest before merge

applications/prompt-keda-lsstcam/values.yaml
```diff
@@ -0,0 +1,185 @@
...
+    # -- If set, get S3 credentials from this application's Vault secret.
+    auth_env: true
+    # -- Set this to disable validation of S3 bucket names, allowing Ceph multi-tenant colon-separated names to be used.
+    disableBucketValidation: 0
+    # -- If set, configure S3 checksum options.
+    checksum: WHEN_REQUIRED
+
+  # -- The URI to a microservice that maps image metadata to a file location.
+  # If empty, Prompt Processing does not use a microservice.
+  raw_microservice: ""
```
eigerx, line 68: Do we want to configure a microservice for LSSTCam?

eigerx (approved):
Looks good to me! Left a few quick questions on configuration choices. Thank you for setting this up!!

applications/prompt-keda-lsstcam/values-usdfdev-prompt-processing.yaml
hsinfang: Agree that we don't want the ticket tag merged.  Though it was convenient in testing so I had it in a commit marking "DO NOT MERGE".

applications/prompt-keda-lsstcam/values.yaml
hsinfang: Yes in `prod` but no in `dev`, so no in the shared `values.yaml`

PR #4431, opened 2025-03-27
DM-48978: Keda Support on Next Visit Fan Out

dspeck1:
Adds env variables and instrument config for Redis Steams to `next-visit-fan-out` and `next-visit-fan-out-keda` applications for co-existence.

### DIFF
--- a/applications/next-visit-fan-out-keda/values.yaml
+++ b/applications/next-visit-fan-out-keda/values.yaml
@@ -1,3 +1,6 @@
+# -- Platform to submit events to.  Either knative or keda.
+platform: "keda"
+
 knative:
   # -- A mapping of instrument to that instrument's Knative service.
   # @default -- See `values.yaml`.
@@ -18,6 +21,14 @@ keda:
   # -- Redis cluster host.
   # @default -- See `values.yaml`.
   redisHost: prompt-redis.prompt-redis
+  # -- Maximum delay time for Redis retries in seconds.
+  redisRetryDelayCap: 5
+  # -- Initial delay for first Redis retry in seconds.
+  redisRetryInitialDelay: 1
+  # -- Redis max retry count
+  redisRetryCount: 3
+  # -- Redis health check interval in seconds.
+  redisHealthCheckInterval: 3
   # -- A mapping of instrument to that instrument's Keda Scaled Job.
   # @default -- See `values.yaml`.
   redisStreams:
--- a/applications/next-visit-fan-out/values-usdfdev-prompt-processing.yaml
+++ b/applications/next-visit-fan-out/values-usdfdev-prompt-processing.yaml
@@ -1,4 +1,4 @@
-platform: knative
+platform: keda
 
 knative:
   maxMessages: 2000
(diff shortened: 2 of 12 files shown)


applications/next-visit-fan-out-keda/values.yaml
```diff
@@ -18,6 +21,14 @@ keda:
...
   # @default -- See `values.yaml`.
   redisHost: prompt-redis.prompt-redis
+  # -- Maximum delay time for Redis retries in seconds
+  redisRetryDelayCap: 5
+  # -- Initial delay for first Redis retry in seconds
+  redisRetryInitialDelay: 1
+  # -- Redis max retry count
+  redisRetryCount: 3
+  # -- Redis health check interval
+  redisHealthCheckInterval: 3
```
kfindeisen, line 31: What are the units? Seconds?

kfindeisen (commented):
I'm a bit confused by what's going on here, and I don't think I can finish the review until I understand.

After these three commits, we have two applications:
- `next-visit-fan-out-keda` handles Keda messages for both prod and dev, true to its name.
- `next-visit-fan-out` handles Keda messages for dev (redundantly with `next-visit-fan-out-keda`) and Knative messages for prod

The charts are duplicated rather than shared, I assume because we plan to drop Knative support soon.

Did I misunderstand anything? If not, then how will this work when deployed?

dspeck1:
I kept both because we are still running Knative in some of the tests.  If we want to use a merge to `main` on this PR as our way of turning off Knative the `next-visit-fan-out-keda` chart can be deleted.

kfindeisen:
I'm still confused. We have two Keda services and no Knative in dev because we're testing with Knative?

I assume this will cause `prompt-keda-<instrument>` to get duplicate fanned-out messages?

applications/next-visit-fan-out-keda/values.yaml
dspeck1: Yes.  I will update comment.

dspeck1:
I should have set `next-visit-fan-out-keda` in dev to disabled. I will add commit to do that.

applications/next-visit-fan-out-keda/values.yaml
```diff
@@ -18,6 +21,14 @@ keda:
   # -- Redis cluster host.
   # @default -- See `values.yaml`.
   redisHost: prompt-redis.prompt-redis
+  # -- Maximum delay time for Redis retries in seconds
+  redisRetryDelayCap: 5
+  # -- Initial delay for first Redis retry in seconds
+  redisRetryInitialDelay: 1
+  # -- Redis max retry count
+  redisRetryCount: 3
+  # -- Redis health check interval
```
kfindeisen, line 30: Please say this is seconds.

applications/next-visit-fan-out/values-usdfdev-prompt-processing.yaml
```diff
@@ -8,7 +8,7 @@ kafka:
   schemaRegistryUrl: http://10.103.126.51:8081
   sasquatchAddress: 10.100.226.209:9094
   consumerGroup: test-group-3
-  nextVisitTopic: test.next-visit
+  nextVisitTopic: test.next-visit-job
```
kfindeisen, line 11: Why is this change needed?

kfindeisen (approved):
Two minor comments.

Just to make sure I understand the desired state: `dev` has only `next-visit-fan-out`, which runs Keda, while `prod` keeps the dual-mode setup we've had for the last few months? That would mean that we'd still be using Knative in `prod` while being unable to test it in `dev`.

applications/next-visit-fan-out-keda/values.yaml
dspeck1: Thanks for reminder.  Fixed.

PR #4435, opened 2025-03-28
Add new obs_env, run_branch topic to the sasquatch connector.

tribeiro:


applications/sasquatch/values-base.yaml
```diff
@@ -285,6 +285,7 @@ rest-proxy:
     topics:
       - lsst.obsenv.action
       - lsst.obsenv.summary
+      - lsst.obsenv.run_branch
```
afausti, line 288: Adding topics here will create them using the KafkaTopic resource in Strimzi, unfortunately the topic operator does not accept "_" in the topic name, I suggest changing to `lsst.obsenv.run-branch` or similar.

afausti (changes requested):
Strimzi does not accept "_" in topic names unfortunately.

afausti (approved):
Perfect, thanks!

PR #4436, opened 2025-03-28
DM-45988: Support storing init-outputs in Prompt Processing central repo

kfindeisen:
This PR adds a Job and CronJob that create init-outputs and output collections for the Prompt Processing service. By design, these jobs share much of their config with the primary service and **must** be updated together in order to work correctly.

### DIFF
--- a/charts/prompt-keda/templates/_service-init.tpl
+++ b/charts/prompt-keda/templates/_service-init.tpl
@@ -41,6 +41,17 @@ spec:
           value: {{ .Values.s3.disableBucketValidation | toString | quote }}
         - name: CONFIG_APDB
           value: {{ .Values.apdb.config }}
+        - name: SASQUATCH_URL
+          value: {{ .Values.sasquatch.endpointUrl }}
+        {{- if and .Values.sasquatch.endpointUrl .Values.sasquatch.auth_env }}
+        - name: SASQUATCH_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: {{ template "prompt-keda.fullname" . }}-secret
+              key: sasquatch_token
+        {{- end }}
+        - name: DAF_BUTLER_SASQUATCH_NAMESPACE
+          value: {{ .Values.sasquatch.namespace }}
         - name: S3_ENDPOINT_URL
           value: {{ .Values.s3.endpointUrl }}
         {{- if .Values.s3.auth_env }}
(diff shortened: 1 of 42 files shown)


charts/prompt-keda/templates/service-init.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: init-{{ include "prompt-keda.fullname" . }}
+  annotations:
+    argocd.argoproj.io/hook: Sync
```
dspeck1, line 7: Make sense and will work.  Did you consider `PreSync` here and `Sync` for the Scaled Job?

charts/prompt-keda/templates/scaled-job.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: keda.sh/v1alpha1
 kind: ScaledJob
 metadata:
   name: {{ include "prompt-keda.fullname" . }}
+  annotations:
+    argocd.argoproj.io/hook: PostSync
```
dspeck1, line 6: Overall this makes sense.  Curious did you try to run the init jobs with a `PreSync` hook and this with a `Sync` hook?

charts/prompt-keda/templates/_service-init.tpl
```diff
@@ -0,0 +1,167 @@
...
+          value: {{ .Values.s3.disableBucketValidation | toString | quote }}
+        - name: CONFIG_APDB
+          value: {{ .Values.apdb.config }}
+        - name: SASQUATCH_URL
+          value: {{ .Values.sasquatch.endpointUrl }}
+        {{- if and .Values.sasquatch.endpointUrl .Values.sasquatch.auth_env }}
+        - name: SASQUATCH_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: {{ template "prompt-proto-service.fullname" . }}-secret
```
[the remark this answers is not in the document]
kfindeisen: I could swear I fixed that... 😕

dspeck1 (approved):
LGTM.  There a couple comments.

charts/prompt-keda/templates/service-init.yaml
kfindeisen: My understanding from the docs is that `PreSync` runs _before_ any config changes are applied. If that's correct, then this does need to be `Sync`.

dspeck1: Ok I understand now b/c of the config and successful sync aspect.  Thanks!

PR #4462, opened 2025-04-02
DM-49427: Filter out old fanned-out messages in Keda

kfindeisen:
This PR adds a message expiration config parameter to Prompt Processing. This parameter allows rejection of old nextVisit messages independently of the fan-out service, but the latter's functionality is not being deprecated at this time.


dspeck1 (approved):
LGTM

PR #4465, opened 2025-04-03
DM-47763: Extend love-conf configuration for love-nginx to support multiple entries.

sebastian-aranda:


mareuter (approved):
Looks good to me.

PR #4471, opened 2025-04-04
DM-49876: Upgrade mysql to version 8.0.41 for TAP_SCHEMA database

stvoutsin:
## Summary 

This addresses jira issue DM-49876:

## Jira issue description

The tap-schema pods of the tap application and other TAP services are crashing on USDF with OOM errors. Investigation suggests the MySQL image version needs to be updated.

## Cause

There was an issue on the USDF clusters caused by the version MySQL used in TAP_SCHEMA and the container runtime version used on some of the upgraded nodes.

## Fix 

Upgrade Mysql to version 8.0.41

Note: Depends on an upgrade to sdm_schemas and new TAP_SCHEMA images

### DIFF
--- a/applications/tap/values-idfdev.yaml
+++ b/applications/tap/values-idfdev.yaml
@@ -1,7 +1,8 @@
 cadc-tap:
   tapSchema:
     image:
-      repository: "lsstsqre/tap-schema-idfdev-tap"
+      repository: "stvoutsin/tap-schema-idfdev-tap"
+      tag: "latest"
 
   config:
     qserv:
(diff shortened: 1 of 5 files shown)


JeremyMcCormick (approved):
Looks good - updates for https://github.com/lsst/sdm_schemas/pull/331

stvoutsin:
@JeremyMcCormick I think the only thing remaining for this is to point to a new weekly release for schemas which contain the schema changes needed by the TAP upgrades here, so let me know when that is out and I'll then update it here

JeremyMcCormick:
@stvoutsin You can use this:

https://github.com/lsst/sdm_schemas/releases/tag/w.2025.15

Docker image tag is `w.2025.15`.

JeremyMcCormick:
I checked this branch today on `usdf-rsp-int` and the DP0.2 tab loads and displays correctly. A search on DP0.2 also works.

PR #4479, opened 2025-04-07
Resync sqre

davedavemckay:
Configuration YAMLs for `ukidacdev` (renamed from `rsp-test`) and `ukidacprod` (renamed from `roe`) have been added.


davedavemckay:
Will resubmit this. A commit was missing from the lsst-uk branch.

PR #4486, opened 2025-04-07
DM-49929: Load OP_CONNECT_TOKEN and VAULT_TOKEN from 1Password via env file

jonathansick:
This PR takes advantage of the 1Password CLI's ability to load secrets referenced by their path in a .env file. With this change, admins no longer need to manually load the VAULT_TOKEN and OP_CONNECT_TOKEN into their shell environments to run phalanx commands like `phalanx secrets sync`. An example usage is:

```sh
PHALANX_ENV="idfprod" op run --env-file="op/square.env" -- phalanx secrets audit
```

Note that I've had the user set PHALANX_ENV, which is used to specify secret paths in `square.env` specific to that Phalanx environment. I've also modified the Phalanx CLI to accept `PHALANX_ENV` as an environment variable instead of as a command-line argument.

Note an alternative approach could be to have a separate `.env` file for each Phalanx environment. This would let us set `PHALANX_ENV` in that file, and make the resulting CL invocations shorter, at the small cost fo maintaining more `.env` files. E.g.

```sh
op run --env-file="op/idfprod.env" -- phalanx secrets audit
```

What do you think? Perhaps this second approach would provide a better user experience.

### DIFF
--- a/docs/admin/audit-secrets.rst
+++ b/docs/admin/audit-secrets.rst
@@ -4,15 +4,27 @@ Audit secrets for an environment
 
 To check that all of the necessary secrets for an environment named ``<environment>`` are in Vault and appear to have the proper form, and to see exactly what a :doc:`syncing secrets for that environment <sync-secrets>` would do, run:
 
-.. prompt:: bash
+.. tab-set::
 
-   phalanx secrets audit <environment>
+   .. tab-item:: General
 
-Add the ``--secrets`` command-line option or set ``OP_CONNECT_TOKEN`` if needed for your choice of a :ref:`static secrets source <admin-static-secrets>`.
-For SQuaRE-managed deployments, the 1Password token for ``OP_CONNECT_TOKEN`` comes from the ``Phalanx 1Password tokens`` item in the SQuaRE 1Password vault.
+      .. prompt:: bash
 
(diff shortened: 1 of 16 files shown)


rra (commented):
The 1Password stuff is great, thank you!

`PHALANX_ENV` for the command-line tool makes me nervous because no matter how much you tell people to use `env` to set the environment variable for only that command, my experience is that some number of people don't listen to you and do `export PHALANX_ENV=idfprod` so that they can stop typing it each time. This then becomes a land mine and two days later they run `phalanx secrets sync --delete` in the same window and accidentally do production (for example).

I therefore feel like forcing people to explicitly name the environment on the command line is a bit of a safety feature, and am not sure getting rid of it is a good idea. It's still true that people can make aliases that accomplish the same thing and are dangerous in the same way, but that requires a bit more intentional effort that hopefully would be linked with a bit more caution.

jonathansick:
Ok that sounds reasonable. So I'll drop `PHALANX_ENV` and then set up an env file for each environment.

jonathansick:
@rra I've reworked the PR with individual env files for each Phalanx environment and also worked the method into our documentation (using tab sets) to show the op run methodology as an option.

rra (approved):
This looks great. Thank you!

PR #4493, opened 2025-04-09
DM-49993: Switch qserv Sasquatch user to TLS authentication

rra:
Now that the Qserv Kafka bridge will be running in the same cluster as Sasquatch, we can use TLS for authentication rather than passwords. Switch the corresponding Kafka user resource.


afausti (approved):
Looks good to me!

PR #4504, opened 2025-04-10
DM-49928: Add remote control user passwords to love-manager on summit.

sebastian-aranda:


applications/love/secrets-summit.yaml
```diff
@@ -0,0 +1,60 @@
+auth-ldap-bind-password:
```
mareuter, line 1: Only add to this file what is _different_ for the summit. Things that appear everywhere should remain in `secrets.yaml`.

mareuter (approved):
Please fixup the secrets, but otherwise this looks fine.

PR #4506, opened 2025-04-10
DM-50055: Revert DM-49767 and revert TAP image versions to 2.11.0

JeremyMcCormick:


rra:
I think this branch can now be deleted since everything has now been upgraded to versions that use the new schema?

PR #4508, opened 2025-04-11
DM-50065: Add `nvr-control` application

fritzm:
Adds a new `nvr-control` application, initially only enabled in the BTS environment.

This is a phalanx-ified version of the prototype described in [TSTN-047](tstn-047.lsst.io), including staging behind Gafaelfawr ingress.

When initially deployed, will be a generic Home Assistant deployment with a PV provisioned to hold configuration state data.  Initial Home Assistant "onboarding" (admin account creation, unifi and auth plugin configuration, dashboard create etc.) will be conducted via a one-time temporary `kubectl` port-forward.  After this, a PV snapshot can be taken to be used in future deployments without re-onboarding should the PV ever be lost.


aibsen (approved):
looks good

PR #4524, opened 2025-04-14
Transition consdb to the logical replica at USDF

bbrondel:
In consdb, the the database should be changed to point to the logical replica instead of the new physical replica.


aibsen (approved):
db host modified so it points to the logical replica instead of the new physical replica. lgtm.

PR #4527, opened 2025-04-15
DM-50054: Update love-manager pods to revision 1 (cycle 41) on summit.

sebastian-aranda:


aibsen (commented):
Change revision to 1. Looks good.

PR #4529, opened 2025-04-15
DM-49281: Update exposurelog to support S3 butler connections on summit and base.

sebastian-aranda:


Vebop (approved):
This was the only app left pulling from lsstsqre/ it looks like, nice updates 👍

PR #4538, opened 2025-04-16
tickets/DM-49953: Change passwd/group to reflect Python container base

athornton:


rra:
Can we delete the `lsst_lcl` stuff too now?

rra:
Why do we want a user account to install everything as? It's all installed on a read-only image that we don't allow the user to further modify, and they don't have that UID anyway, so why not install everything owned by root?

athornton:
The main reason is that I don't know what happens if I run lsstinstall as root.  Maybe it's perfectly fine, but it's meant to be run as an unprivileged user.  https://ls.st/lsstinstall is nontrivial.  If you feel strongly about it, though, now is the time to try building the container that way, BEFORE we've cut a recommended.

rra:
I guess I'm confused about what would misbehave, but if you want to not run it as root, you could run it as a non-root user and then chown all the files and delete the user.

athornton:
The worst it's gonna misbehave, I guess, is that it destroys the container I'm trying to build, but what I'm more worried about is that the container doesn't work because of permissions or something.  I guess though that if it's got the wrong permissions for UID 1000 then it's no different from wrong permissions for UID 0 from the user (who is neither of those) perspective.

athornton:
I take it we should also retool jupyterlab-base to install the UI Venv as root, then?

athornton:
It feels like a bad idea to me to have the container default to running as UID 0 and relying on K8s to set runAsUser correctly.  At least for the base container, sure, I'll install everything as root, but I still want to change to 1000:1000 to start the Lab.

athornton:
Meanwhile back in lsstinstall:

```
[1500](https://github.com/lsst-sqre/sciplat-lab/actions/runs/14624083296/job/41031247866#step:8:1501)
 > [base-stack-image 2/6] RUN ./install-dm-stack w_2025_16:
[1501](https://github.com/lsst-sqre/sciplat-lab/actions/runs/14624083296/job/41031247866#step:8:1502)
129.9 and feel free to ask any questions via the LSST Community forum:
[1502](https://github.com/lsst-sqre/sciplat-lab/actions/runs/14624083296/job/41031247866#step:8:1503)
129.9     https://community.lsst.org/c/support
[1503](https://github.com/lsst-sqre/sciplat-lab/actions/runs/14624083296/job/41031247866#step:8:1504)
129.9 + source /opt/lsst/software/stack/loadLSST.bash
[1504](https://github.com/lsst-sqre/sciplat-lab/actions/runs/14624083296/job/41031247866#step:8:1505)
129.9 ++ '[' -z /bin/sh ']'
[1505](https://github.com/lsst-sqre/sciplat-lab/actions/runs/14624083296/job/41031247866#step:8:1506)
129.9 ++++ basename /bin/sh
[1506](https://github.com/lsst-sqre/sciplat-lab/actions/runs/14624083296/job/41031247866#step:8:1507)
129.9 +++ /opt/lsst/software/stack/conda/bin/conda shell.sh hook
[1507](https://github.com/lsst-sqre/sciplat-lab/actions/runs/14624083296/job/41031247866#step:8:1508)
130.0 ++ __conda_setup=
[1508](https://github.com/lsst-sqre/sciplat-lab/actions/runs/14624083296/job/41031247866#step:8:1509)
130.0 ++ echo 'Unknown shell'
[1509](https://github.com/lsst-sqre/sciplat-lab/actions/runs/14624083296/job/41031247866#step:8:1510)
130.0 ++ exit 1
[1510](https://github.com/lsst-sqre/sciplat-lab/actions/runs/14624083296/job/41031247866#step:8:1511)
130.0 Unknown shell
```

Having /bin/sh as your shell?!?  MADNESS

rra:
> It feels like a bad idea to me to have the container default to running as UID 0 and relying on K8s to set runAsUser correctly. At least for the base container, sure, I'll install everything as root, but I still want to change to 1000:1000 to start the Lab.

I'm not sure I understand. It's equally bad to start the lab as 1000:1000; either way, we are relying on Kubernetes to start the lab as the correct user. I think it would have to be some non-Nublado use case for the container where this would matter, but isn't the container intended specifically for use with Nublado and nothing else? It relies on a bunch of environment variables that only Nublado sets, for example.

rra:
Also you could set the default user to be `nobody:nogroup` if you want to make sure the default is something unprivileged.

athornton:
Well, since we're removing /etc/passwd and friends, I don't know that there's any reason to use 65534:65534 vs 1000:1000.  Is there?

athornton:
OK, this dovetails with https://github.com/lsst-sqre/nublado/pull/519 and https://github.com/lsst-sqre/sciplat-lab/pull/193

athornton:
I merged those, everything seems fine, so I think this is good now too.

rra (approved):
Looks great, thanks. This will produce possibly strange results when people spawn old images, but I don't think that should cause any actual problems, just cosmetic issues with `ls` output.

athornton:
We've got that image as recommended.  I've rebased and will merge once the checks pass.

PR #4541, opened 2025-04-16
DM-50267: Prompt Processing SSL Kafka Consumer Fix

dspeck1:
Change to `SASL_SSL` and Kafka broker IPs after Sasquatch issues with connectivity to Summit through firewall.


ktlim (approved):
LGTM.

PR #4544, opened 2025-04-17
DM-50272: argo-cd - Slack notifications for unhealthy apps

fajpunk:


fajpunk:
This isn't going to work because ArgoCD bases app health on the health of the immediate child resources, so the workload resources like Deployment and StatefulSet, and not Pods... and ArgoCD will never say that a StatefulSet is unhealthy, the worst it gets is Progressing.

PR #4551, opened 2025-04-17
Update sdm_schemas to w.2025.16

JeremyMcCormick:


stvoutsin:
Validated that this works fine for tap and ssotap

PR #4558, opened 2025-04-17
Add ssotap config to portal for USDF environments

JeremyMcCormick:


gpdf (approved):
Looks reasonable to me as long as "ssotap" is actually in all three instances.

@rra Is there some reason "livetap" might have been left out of the "usdf-int" configuration that you can think of?

rra:
usdfint is fairly new. I'm guessing it just wasn't set up there for some reason. I don't know of any reason why we would have intentionally omitted it.

JeremyMcCormick:
Did something change on usdf-rsp-int that fixed this?

I have not deployed this and the DP0.3 tab is showing correctly there now.

PR #4578, opened 2025-04-21
Add tutorial notebook pull cronjob/idfdev

athornton:
This will keep our tutorial notebooks fresh, initially on IDF-dev so we don't have to do a clone with each lab start.

### DIFF
--- a/applications/nublado/templates/tutorial-cronjob.yaml
+++ b/applications/nublado/templates/tutorial-cronjob.yaml
@@ -1,0 +1,56 @@
+{{ if .Values.cronjob.enabled }}
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: tutorials
+spec:
+  schedule: {{ .Values.cronjob.schedule | quote }}
+  successfulJobsHistoryLimit: 1
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          restartPolicy: Never
+          automountServiceAccountToken: false
+          {{- with .Values.cronjob.tolerations }}
+          tolerations:
+{{- toYaml . | indent 12 }}
+          {{- end }}
+          {{- with .Values.cronjob.affinity }}
+          affinity:
+{{- toYaml . | indent 12 }}
+          {{- end }}
+          containers:
+          - name: tutorial
+            image: "{{ .Values.cronjob.image.repository }}:{{ .Values.cronjob.image.tag | default .Chart.AppVersion }}"
+            imagePullPolicy: "{{ .Values.cronjob.image.pullPolicy }}"
+            {{- with .Values.cronjob.resources }}
+            resources:
+{{ toYaml . | indent 14 }}
+            {{- end }}
+            command: [ "/entrypoint.sh" ]
+            securityContext:
+              allowPrivilegeEscalation: false
+              readOnlyRootFilesystem: true
+              runAsNonRoot: true
+              runAsUser: 1000
+              runAsGroup: 1000
+              capabilities:
+                drop:
+                - all
+            env:
+            - name: "GIT_SRC"
+              value: {{ .Values.cronjob.gitSource | quote }}
+            - name: "GIT_TARGET"
+              value: {{ .Values.cronjob.gitTarget | quote }}
+            - name: "GIT_BRANCH"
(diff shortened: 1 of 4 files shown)


athornton:
Eventually this should be rolled into the big service discovery service, but for now this is I think a reasonable way to avoid a clone-per-lab-start.

applications/nublado/templates/tutorial-cronjob.yaml
```diff
@@ -0,0 +1,56 @@
...
+                - all
+            env:
+            - name: "GIT_SRC"
+              value: {{ .Values.cronjob.gitSource | quote }}
+            - name: "GIT_TARGET"
+              value: {{ .Values.cronjob.gitTarget | quote }}
+            - name: "GIT_BRANCH"
+              value: {{ .Values.cronjob.gitBranch | quote }}
+            volumeMounts:
+            - mountPath: {{ .Values.cronjob.targetVolume.mountPath }}
```
stvoutsin, line 49: Did you mean:
`- mountPath: {{ .Values.cronjob.targetVolumePath }}
`

applications/nublado/templates/tutorial-cronjob.yaml
```diff
@@ -0,0 +1,56 @@
...
+              name: target
+          volumes:
+          - name: target
+            nfs:
+              server: {{ .Values.cronjob.targetVolume.server }}
+              path: {{ .Values.cronjob.targetVolume.path }}
```
stvoutsin, line 55: And here should it be:
```
      server: {{ .Values.cronjob.targetVolume.nfs.server }}
      path: {{ .Values.cronjob.targetVolume.nfs.path }}
```

stvoutsin (changes requested):

applications/nublado/templates/tutorial-cronjob.yaml
athornton: Nope, but I do need to correct values.yaml with the flattened version.

applications/nublado/values.yaml
```diff
@@ -675,6 +675,60 @@ secrets:
...
+    requests:
+      cpu: "50m"
+      memory: "50Mi"
+
+  # -- Tolerations for the tutorials cronjob.
+  tolerations: []
+  # -- Affinity rules for the tutorials cronjob.
+  affinity: {}
+  # -- Schedule for the tutorials cronjob.
+  schedule: "43 * * * *"
```
stvoutsin, line 729: I'm wondering if this is excessive, and if once or twice a day would be sufficient?

applications/nublado/values.yaml
```diff
@@ -675,6 +675,60 @@ secrets:
...
+  gitSource: "https://github.com/lsst/tutorial-notebooks"
+
+  # -- Target where repository should land
+  gitTarget: "/project/cst_repos/tutorial-notebooks"
+
+  # -- Branch of repository to clone
+  gitBranch: "main"
+
+  # -- Where repository volume should be mounted
+  targetVolumePath: "/project"
```
stvoutsin, line 694: It seems that this may be redundant since the `targetVolume` defines a `mountPath` which gets used?

applications/nublado/values.yaml
```diff
@@ -675,6 +675,60 @@ secrets:
...
+# their refresh.
+cronjob:
+  # -- Enable tutorials refresh?
+  enabled: false
+
+  # -- Source for repository to clone
+  gitSource: "https://github.com/lsst/tutorial-notebooks"
+
+  # -- Target where repository should land
+  gitTarget: "/project/cst_repos/tutorial-notebooks"
```
stvoutsin, line 688: If we're always cloning at the target mount path perhaps we can omit the` /project/ ` here and construct the env variable using the t`argetVolume.mountPath`? Unless we expect to ever use a different directory?

stvoutsin (approved):
Looks good, only some minor comments

applications/nublado/values.yaml
athornton: I'd like to see CST updates quickly, and one pull from a repo per hour per site is...not that excessive.  I think an hour is a reasonable default, but of course I do since that's why I chose the value.

PR #4585, opened 2025-04-23
DM-50462: Add MPSky to Prompt Processing configs

kfindeisen:
This PR adds configuration hooks for `MP_SKY_URL` to Prompt Processing. However, it leaves the variable undefined for everything except LSSTCam-prod.

### DIFF
--- a/applications/prompt-keda-hsc/values.yaml
+++ b/applications/prompt-keda-hsc/values.yaml
@@ -67,6 +67,10 @@ prompt-keda:
   # If empty, Prompt Processing does not use a microservice.
   raw_microservice: ""
 
+  # -- The URI to the MPSky ephemerides service.
+  # Empty value is allowed, but its handling is undefined.
+  mpSky_service: ""
+
   imageNotifications:
     # -- Hostname and port of the Kafka provider
     # @default -- None, must be set
--- a/charts/prompt-keda/templates/scaled-job.yaml
+++ b/charts/prompt-keda/templates/scaled-job.yaml
@@ -85,6 +85,11 @@ spec:
                 value: {{ .Values.imageNotifications.kafkaClusterAddress }}
               - name: RAW_MICROSERVICE
                 value: {{ .Values.raw_microservice }}
+              # MpSkyEphemerisQuery handles MP_SKY_URL being unset, but not empty
+              {{- with .Values.mpSky_service }}
+              - name: MP_SKY_URL
+                value: {{ . }}
+              {{- end }}
               - name: SASQUATCH_URL
                 value: {{ .Values.sasquatch.endpointUrl }}
               {{- if and .Values.sasquatch.endpointUrl .Values.sasquatch.auth_env }}
--- a/charts/prompt-proto-service/templates/prompt-proto-service.yaml
+++ b/charts/prompt-proto-service/templates/prompt-proto-service.yaml
@@ -75,6 +75,11 @@ spec:
           value: {{ .Values.imageNotifications.kafkaClusterAddress }}
         - name: RAW_MICROSERVICE
           value: {{ .Values.raw_microservice }}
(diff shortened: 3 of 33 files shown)


applications/prompt-keda-hsc/values.yaml
```diff
@@ -67,6 +67,10 @@ prompt-keda:
   # If empty, Prompt Processing does not use a microservice.
   raw_microservice: ""
 
+  # -- The URI to the MPSky ephemerides service.
+  # Empty value is allowed, but its handling is undefined.
+  mpSky_service: ""
```
kfindeisen, line 72: Copying @Gerenjie's comments from Jira:
> Wouldn’t the current setting, `mpSky_service: ""`, pass the empty string (rather than no value) to the environment variable, spoiling the intended effect of undefined value rather than empty value? 
>
> For example, when I set the MP_SKY_URL environment variable to '', I get the wrong effect of os.environ.get('MP_SKY_URL', fallback) returning '' rather than fallback.

charts/prompt-keda/templates/scaled-job.yaml
```diff
@@ -85,6 +85,11 @@ spec:
                 value: {{ .Values.imageNotifications.kafkaClusterAddress }}
               - name: RAW_MICROSERVICE
                 value: {{ .Values.raw_microservice }}
+              # MpSkyEphemerisQuery special-cases no variable but not empty variable
+              {{- with .Values.mpSky_service }}
+              - name: MP_SKY_URL
+                value: {{ . }}
+              {{- end }}
```
kfindeisen, line 92: Copying @Gerenjie's comments from Jira:
> The documentation “# MpSkyEphemerisQuery special-cases no variable but not empty variable” isn’t super clear to me – maybe “MpSkyEphemerisQuery handles MP_SKY_URL being unset, but not empty.”?

applications/prompt-keda-hsc/values.yaml
kfindeisen: The `with` action (I think that's the term?) in the template takes care of that: if `mpSky_service` is non-empty (bool conversion works the same way in Go templates as it does in Python), it defines `MP_SKY_URL` with the same value. If it _is_ empty, it doesn't define `MP_SKY_URL` at all. That keeps the user-facing YAML config from being hypersensitive to exactly whether it needs to be `""`, `null`, etc.

kfindeisen:
> I’m not in love with setting `mpSky_service: http://sdfiana014.sdf.slac.stanford.edu:3666/ephemerides/` in a second place (fallback + Phalanx) if there’s not an explicit reason to, since when things change, we’ll have to remember everywhere we need to change it.

I see it as exactly the opposite -- this is the place where we _should_ be setting the service, because it's part of Prompt Processing's runtime environment and not a property of the code. If and when we have a permanent service up, we need to set it here _whether or not_ it also becomes the dev environment default. If we did as you propose and made the "fallback" the primary URL, we'd have to create a new Prompt Processing release with the `ap_association` change and carefully coordinate it with deployment of the service.

Personally, I hope the fallback is a temporary measure because having environment stuff in configs is nothing but trouble (as we learned the hard way with the APDB, but it's too late to back out now). But that's none of my business.

charts/prompt-proto-service/templates/prompt-proto-service.yaml
```diff
@@ -75,6 +75,11 @@ spec:
           value: {{ .Values.imageNotifications.kafkaClusterAddress }}
         - name: RAW_MICROSERVICE
           value: {{ .Values.raw_microservice }}
```
kfindeisen, line 77: A counterexample to maybe explain the `with` block a bit better -- these two lines always define `RAW_MICROSERVICE`, but may define it as the empty string (which in this case means "there isn't one, don't try to connect").

Gerenjie (approved):

PR #4593, opened 2025-04-24
Update Prompt Processing alert stream.

kfindeisen:
This PR updates the Prompt Processing alert topic from `lsst-alerts` to `lsst-alerts-v7.4`. This is for forward-compatibility only; we are not yet running any pipelines that send alerts to this topic.

It also reorganizes the LSSTCam-prod config file so that the Prompt Processing version and alert stream version are easier to update together.


kfindeisen:
See #4595 for the alert stream itself.

hsinfang (approved):
Looks good. Thank you for updating it!

PR #4594, opened 2025-04-24
Fix conditional for telegraf-password

afausti:
There was a typo in the secret conditional for the `telegraf-password`  as revealed by `phalanx secrets audit`.
Now `phalanx secrets audit` looks correct for both IDF dev and int.
I also removed the unused static secrets from 1password for those environments.


rra (approved):
Great, thank you!

PR #4609, opened 2025-04-28
WIP:  mpsky: solar system ephemerides computation service for PP

mjuric:
This PR adds an `mpsky` application, deploying a containerized version of [MPSky](https://github.com/mjuric/mpsky) that can be accessed by PP pipelines at nighttime.

This service is meant to be used purely internally (by the PP pipeline), is read-only, and doesn't come with authentication.  On KTL's recommendation, to make it visible internally (and inaccessible externally) we've dropped the Ingress and changed the service type to a LoadBalancer instead, using an IP from sdf-services pool. (see discussion [here](https://rubin-obs.slack.com/archives/C07Q45N7KHV/p1742520202629099)).

This is an initial merge, with more development expected to occur over the next few days; that's why the image tag is currently set to 'latest'.


applications/mpsky/templates/service.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: "mpsky"
+  labels:
+    {{- include "mpsky.labels" . | nindent 4 }}
+  annotations:
+#    metallb.universe.tf/address-pool: sdf-services
+    loadBalancerIP: 172.24.5.245
```
jonathansick, line 9: This IP is being baked into the service template, and therefore will be the same for all Phalanx environments where mpsky is deployed (usdf-rsp-dev, usdf-rsp, etc.). I _think_ this may need to be different for each deployment. To do this, add a value to `values.yaml` that will get overriden for each environment (i.e., `values-usdfdev.yaml`). To handle that awkward initial deployment, you could do something like set the default loadBlancerIP to "none" and do a conditional check to use the sdf-services annotation instead, like this (I haven't checked this syntax, so try this first):

```suggestion
  annotations:
    {{- if .Values.loadBalancerIP == "none" }}
    metallb.universe.tf/address-pool: sdf-services
    {{- else }}
    loadBalancerIP: {{ .Values.loadBalancerIP | quote }}
    {{- end }}
```

Last, add a comment here that captures the discussion from Slack about how this IP address is determined.

applications/mpsky/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,21 @@
...
+    # Allow inbound access from pods (in any namespace) labeled
+    # gafaelfawr.lsst.io/ingress: true.
+    - from:
+        - namespaceSelector: {}
+          podSelector:
+            matchLabels:
+              gafaelfawr.lsst.io/ingress: "true"
+      ports:
+        - protocol: "TCP"
+          port: 8080
```
jonathansick, line 21: Is the NetworkPolicy here necessary, given that gafaelfawr and ingress are out of the picture?

athornton:
So when you say "used internally" you mean "accessible only to things running inside the USDF network, but not necessarily on this K8s cluster" ?

PR #4614, opened 2025-04-28
Changes to prompt processing prod for 2025-04-28

hsinfang:


kfindeisen (approved):
Looks good! Looking at Eli's post, do we have flats for the new detectors yet?

hsinfang:
Still only `pseudoFlats`, no new flats yet.

PR #4617, opened 2025-04-29
DM-50571: Enable prompt-processing Kafka User on Summit

dspeck1:
Enable prompt processing Kafka user on the Summit.

### DIFF
--- a/applications/sasquatch/charts/prompt-processing/templates/kafka-topics.yaml
+++ b/applications/sasquatch/charts/prompt-processing/templates/kafka-topics.yaml
@@ -1,21 +1,13 @@
-
+{{- $cluster := .Values.cluster.name }}
+{{- range $topic := .Values.topics }}
 ---
-apiVersion: kafka.strimzi.io/v1beta2
+apiVersion: kafka.strimzi.io/v1beta1
 kind: KafkaTopic
 metadata:
-  name: "test.next-visit"
+  name: {{ $topic }}
   labels:
-    strimzi.io/cluster: {{ $.Values.cluster.name }}
+      strimzi.io/cluster: {{ $cluster }}
 spec:
-  partitions: 1
   replicas: 3
----
-apiVersion: kafka.strimzi.io/v1beta2
-kind: KafkaTopic
-metadata:
-  name: "test.next-visit-job"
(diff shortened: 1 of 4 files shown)


afausti (commented):
This will install the `prompt-processing` subchart at the Summit, which has the Strimzi KafkaTopic and KafkaUser resources. Do you know if the `test.` topics are still needed? If they are still in use at the dev cluster I would suggest doing something similar to what we do in the [control-system subchart](https://github.com/lsst-sqre/phalanx/blob/main/applications/sasquatch/charts/control-system/values.yaml#L9) so that we can control which topics are created. The reason for doing that is because we dont want the `test.` topics created at the Summit. In case those topics are no longer used I would just remove them.

dspeck1:
The topics are in use in Dev.  Thanks!  I update the chart to use the same logic as the control-system sub-chart.  Ready for review again.

applications/sasquatch/charts/prompt-processing/templates/kafka-topics.yaml
```diff
@@ -1,21 +1,13 @@
-
+{{- $cluster := .Values.cluster.name }}
+{{- range $topic := .Values.topics }}
 ---
-apiVersion: kafka.strimzi.io/v1beta2
+apiVersion: kafka.strimzi.io/v1beta1
```
afausti, line 4: I think the correct API version for Strimzi is v1beta2, any reason for changing it?

afausti (approved):
Thanks for taking care of the test topics, only made a comment on. the Strimzi API version which I think should be v1beta2

applications/sasquatch/charts/prompt-processing/templates/kafka-topics.yaml
dspeck1: Thanks for catching.  It was copied from the [control-system](https://github.com/lsst-sqre/phalanx/blob/5966a529189424cca2fc9e42342677254ab413a5/applications/sasquatch/charts/control-system/templates/kafka-topics.yaml#L4) and didn't notice it.  Updated it to `v1beta2`

afausti (approved):

PR #4618, opened 2025-04-29
DM-50564: Change ingester read topic from alerts-simulated to lsst-alerts-v7.4

bsmartradio:
Switch the ingester from reading alerts-simulated to lsst-alerts-v7.4


erinleighh (approved):
Thanks for letting me see the process!

PR #4632, opened 2025-05-01
DM-50635: Add ukidacprod and ukidacdev environments

stvoutsin:


davedavemckay:
Looks good to me!
A couple of points:
* the changes make it look like `roe` values YAMLs were renamed to `ukidacdev`, then changed, while the `ukidacprod` YAMLs were created, and mostly matched the contents of the `roe` ones. Noting here to avoid future confusion: `roe` has been renamed to `ukidacprod` and `ukidacdev` is new.
* there are values YAMLs for `exposurelog`, which could be removed - `exposurelog` is not used in the UK IDAC RSPs and so is not listed in `environments` values YAMLs.

PR #4655, opened 2025-05-05
DM-50719: Convert summit Argo CD to Keycloak

rra:
Use Keycloak for summit Argo CD authentication instead of GitHub.


rra:
Do not merge before the next summit maintenance window. Requires a Phalanx secrets sync first.

PR #4665, opened 2025-05-07
Bump Wobbly limit to more than 1 CPU to try to silence warnings

athornton:


rra (approved):
I could have sworn I'd tried this already but I guess not. I suppose there's no real problem with the main wobbly process also being allowed 1.2 CPUs.

PR #4671, opened 2025-05-07
Tickets/dm 50766

tribeiro:


applications/sasquatch/values-summit.yaml
```diff
@@ -382,6 +390,15 @@ telegraf:
...
       offset: newest
+    scimma:
+      enabled: true
+      database: "lsst.scimma"
+      timestamp_format: "unix_ms"
+      timestamp_field: "timestamp"
+      topicRegexps: |
+        [ "lsst.scimma" ]
+      debug: true
+      offset: newest
```
afausti, line 401: In general is better to use the default here `offset: oldest`, which means that when the connector restarts it will continue consuming the earliest message available as opposed to the latest avoiding gaps in the data.

applications/sasquatch/charts/scimma/templates/kafka-topics.yaml
```diff
@@ -0,0 +1,20 @@
+---
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaTopic
+metadata:
+  name: "lsst.scimma.too_alert"
```
afausti, line 5: The use of  `_` or  `-` as separator is a little problematic. Strimzi does not accept `_` in the topic name, and if you use `-` you have to be careful when creating the Avro schema because the REST Proxy doesn't like `-` in the schema name. My preference here would be camel case for the topic and schema names `lsst.scimma.tooAlert`

applications/sasquatch/charts/scimma/templates/kafka-topics.yaml
```diff
@@ -0,0 +1,20 @@
...
+  labels:
+    strimzi.io/cluster: {{ $.Values.cluster.name }}
+spec:
+  partitions: 1
+  replicas: 3
+---
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaTopic
+metadata:
+  name: "lsst.scimma.too_alert_test"
```
afausti, line 15: Same comment as above.

applications/sasquatch/values-base.yaml
```diff
@@ -259,6 +259,15 @@ telegraf:
...
       debug: true
+    scimma:
+      enabled: true
+      database: "lsst.scimma"
+      timestamp_format: "unix_ms"
+      timestamp_field: "timestamp"
+      topicRegexps: |
+        [ "lsst.scimma" ]
+      debug: true
+      offset: newest
```
afausti, line 270: I would use "oldest" unless you have a strong reason for using "newest" as explained above.

applications/sasquatch/values-usdfdev.yaml
```diff
@@ -157,6 +157,11 @@ kafka-connect-manager:
         connectInfluxDb: "lsst.prompt"
         topicsRegex: "lsst.prompt.*"
         tags: dataset_tag,band,instrument,skymap,detector,physical_filter,tract,exposure,patch,visit,group
+      scimma:
+        enabled: true
+        timestamp: "timestamp"
+        connectInfluxDb: "lsst.scimma"
+        topicsRegex: "lsst.scimma.*"
```
afausti, line 164: I would remove this configuration from usdfdev in favor or base.

applications/sasquatch/values-usdfdev.yaml
```diff
@@ -179,6 +184,7 @@ rest-proxy:
       - lsst.verify
       - lsst.lf
       - lsst.prompt
+      - lsst.scimma
```
afausti, line 187: Same here.

afausti (changes requested):
I think you have everything you need, I made a suggestion for the topic names to avoid problems with "_" or "-" separators and I think in general we want `offset: "oldest"` in the connector  connector config to consume the earliest messages. There's still some configuration in usdfdev that I would remove in favor of base.

afausti (approved):
LGTM!

PR #4674, opened 2025-05-07
DM-50751: Remove prompt-proto-service from Phalanx

kfindeisen:
This PR removes the `prompt-proto-service` application and all direct references to it from the environment. Some vestigial knative configs are kept in `next-visit-fan-out`, because they can't be removed without code changes.


dspeck1 (approved):
LGTM

PR #4678, opened 2025-05-08
Shorten idle culling time for lab in IDF-dev

athornton:
Short for XSRF testing.


athornton:
Ah, OK.  If those settings are not used, we should remove them.

rra (approved):

PR #4701, opened 2025-05-13
DM-50858: Add global.environmentName to app template

fajpunk:
Currently this used by some apps to provide the environment name as the `SENTRY_ENVIRONMENT` variable, but it could be useful for other things.


rra (approved):
Thanks! Sorry for the tedious test case updating.

PR #4705, opened 2025-05-13
DM-50846: Support OCI repos in the Phalanx CLI

fajpunk:
[`helm repo` commands aren't supported for OCI repos](https://github.com/helm/helm/issues/10565#issuecomment-1015531100).

Without this, I can't `phalanx application template` an app with a helm dependency from an OCI registry, like:

```
- name: "grafana-operator"
  repository: "oci://ghcr.io/grafana/helm-charts"
  version: "v5.17.1"
```

I'm not sure if there's anywhere else in the Phalanx CLI code we need to modify to account for this.

### DIFF
--- a/src/phalanx/storage/config.py
+++ b/src/phalanx/storage/config.py
@@ -300,7 +300,7 @@ class ConfigStorage:
         for dependency in chart.get("dependencies", []):
             if "repository" in dependency:
                 repository = dependency["repository"]
-                if not repository.startswith("file:"):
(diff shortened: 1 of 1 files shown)


athornton (approved):
LGTM

rra (approved):
I think that should hopefully be it. I guess it downloads the Helm chart anyway without needing to index the repository first?

fajpunk:
Yep, it seems to grab the chart just fine.

PR #4721, opened 2025-05-16
Switch ukidacdev cert-manager to route53

gpfrancis:
Update the ukidacdev environment to use Let's Encrypt DNS and Route53 for TLS instead of using an external certificate.


stvoutsin:
There is a small yaml linting issue but looks ok other than that

PR #4742, opened 2025-05-21
Tickets/DM-49936: enable Eups-distributor

roceb:
Create a nginx app that allows for the sharing of eups tarballs.


ktlim (approved):
Looks reasonable. Kind of a shame we need a custom image, but at least we don’t have the big Persistent Disk.

roceb:
Understandable. I will look into a way to eliminate it if we can. Important part is that we have a working

PR #4750, opened 2025-05-22
tickets/DM-50924: move to finalized paths for NetApp in idf-dev/int

athornton:


applications/portal/values-idfint.yaml
```diff
@@ -5,8 +5,8 @@ config:
   volumes:
     sharedWorkarea:
       nfs:
-        path: "/share1/home/firefly/shared-workarea"
-        server: "10.22.240.130"
+        path: "/firefly-share/firefly/shared-workarea"
```
stvoutsin, line 8: Could you remind me why the path is slightly different than `idfdev` (`/firefly/`)

stvoutsin (approved):
Looks good as far as I can see, with one question

applications/portal/values-idfint.yaml
athornton: I need to check that it actually is....

athornton: Turns out it is int that doesn't match reality.

PR #4764, opened 2025-05-22
DM-51077: Make PP LSSTCam prod aware of the embargo AWS profile 

hsinfang:


kfindeisen (approved):
Looks good, assuming you're planning to make the dev changes on DM-51022.

PR #4791, opened 2025-05-30
tickets/DM-49959B: allow configuration of Lab landing page

athornton:
Adds a setting to override initial page the user gets when spawning a lab.


athornton:
cf https://github.com/lsst-sqre/lsst-rsp/pull/80

rra (approved):

PR #4807, opened 2025-06-03
tickets/SP-2213: Add an nginx webserver to serve static pages behind Rubin Obs. authentication

ehneilsen:
For discussion, see this [slack thread](https://rubin-obs.slack.com/archives/C07Q45N7KHV/p1746565895184519).

The webserver serves everything in `/sdf/data/rubin/shared/scheduler/reports` under `https://usdf-rsp-int.slac.stanford.edu/schedview-static-pages`. For example, `/sdf/data/rubin/shared/scheduler/reports/doc/planisphere.html` is served at `https://usdf-rsp-int.slac.stanford.edu/schedview-static-pages/doc/planisphere.html`.

I use the "unprivileged" variant of the official nginx docker container (which serves on port 8080) so that it can run without root.


rhiannonlynne (approved):
I am completely unqualified to do this review, but it looks fine to me. 
One question I suppose - the service limit is 8GB with a 'request' (which I think is the general expected-use?) of 4GB .. is this enough to serve the pages we're expecting in nginx?

PR #4814, opened 2025-06-04
DM-51022: Change rubin-pp-dev-users/central_repo_2 butler repo to ChainedDatastore like /repo/embargo

hsinfang:


applications/prompt-keda-lsstcomcamsim/values-usdfdev-prompt-processing.yaml
```diff
@@ -25,6 +25,7 @@ prompt-keda:
   s3:
     imageBucket: rubin-pp-dev
     endpointUrl: https://s3dfrgw.slac.stanford.edu
+    aws_profile: prompt-processing-dev
```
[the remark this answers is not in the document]
kfindeisen: I assume it's defined for `usdfdev` to emulate the use of an explicit profile in the production Butler config?

applications/prompt-keda-lsstcam/values-usdfprod-prompt-processing.yaml
```diff
@@ -72,7 +72,7 @@ prompt-keda:
   s3:
     imageBucket: rubin-summit
     endpointUrl: https://sdfembs3.sdf.slac.stanford.edu
-    cred_file_auth: True
+    aws_profile: embargo
```
kfindeisen, line 75: I wonder if it might be worth keeping `cred_file_auth` here just because we really do need it, whether or not it's the default. I assume `aws_profile` is only useful if `cred_file_auth` is set?

kfindeisen (changes requested):
No objections to the changes, but `aws_profile` isn't actually documented anywhere. Please add proper docs to all the `values.yaml` files!

applications/prompt-keda-lsstcomcamsim/values-usdfdev-prompt-processing.yaml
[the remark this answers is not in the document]
hsinfang: They are now documented and sorry for missing it too! 

I'd prefer the default being empty, in which case S3 utilities would take its default accordingly. Always setting it to `default` isn't quite right, because S3 environment variables can take precedence. This is now better documented too.

applications/prompt-keda-lsstcam/values-usdfprod-prompt-processing.yaml
hsinfang: > aws_profile is only useful if cred_file_auth is set
Yes. 

I'm adding a check for that.  The deployment would fail if the parameters aren't consistent.  

With that, maybe it's okay not keeping `cred_file_auth` there?

hsinfang:
@kfindeisen  may you please take another look?  
Please ignore the "[DO NOT MERGE]" commits; they will be removed.

charts/prompt-keda/templates/scaled-job.yaml
```diff
@@ -124,6 +124,9 @@ spec:
                 value: /app/s3/credentials
               {{- end }}
               {{- if .Values.s3.aws_profile }}
+                {{- if not $.Values.s3.cred_file_auth }}
+                  {{- fail "When s3.aws_profile is set, s3.cred_file_auth must be true." }}
+                {{- end }}
```
kfindeisen, line 129: This seems like this would make it harder to switch between file- and env-based authentication. Does any harm result from setting `AWS_PROFILE` when `AWS_SHARED_CREDENTIALS_FILE` is not defined?

charts/prompt-keda/values.yaml
```diff
@@ -66,9 +66,9 @@ s3:
   endpointUrl: ""
   # -- If set, define environment variables with S3 credentials from this application's Vault secret.
   auth_env: false
-  # -- If set, get a S3 credential file from this application's Vault secret.
+  # -- If set, get a S3 credential file from this application's Vault secret. Must be true if `aws_profile` is set.
   cred_file_auth: true
-  # -- If set, specify a S3 credential profile.
+  # -- If set, specify a S3 credential profile and `cred_file_auth` must be true. If empty and `auth_env`=`false`, the `default` profile is used.
```
kfindeisen, line 71: Should this be "If empty and `cred_file_auth = true`"? I wouldn't expect the profile to interact with `auth_env` either way.

kfindeisen: I suggest dividing the sentences one per line, like is done for e.g. `raw_microservice`.

kfindeisen (approved):
I'm not sure about forbidding the use of `aws_profile` without `cred_file_auth` -- what problem is this trying to solve? Otherwise looks good.

charts/prompt-keda/values.yaml
hsinfang: 1. `aws_profile` is set, `cred_file_auth` must be `true`. The profile is used no matter `auth_env` is `true` or `false`  (the two S3 cred env vars are set or not; they are ignored anyway). 
2. `aws_profile` is not set. `cred_file_auth` can be either `true` or `false`. If `auth_env` is `true`, the env vars are used. No profile is used.
3. `aws_profile` is not set and `auth_env` is `false`, the `default` profile is used and expected, otherwise failure.  
  
I was trying to cover the (2) scenario, so didn't want to say "the `default` profile is used when `aws_profile` is empty".  I guess the wording is still confusing?

PR #4821, opened 2025-06-06
tickets/DM-51117: Allow specification of config TOML for Kapacitor in sasquatch

athornton:
Allow us to add squadcast alerts to Kapacitor in sasquatch by letting us specify the squadcast webhook endpoint in config toml.


applications/sasquatch/charts/kapacitor/Chart.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: v1
+name: kapacitor
+version: 1.0.0
+appVersion: 1.6.4
+description: InfluxDB's native data processing engine. It can process both stream and batch data from InfluxDB.
+keywords:
```
afausti, line 6: This keywords field can be removed.

applications/sasquatch/charts/kapacitor/Chart.yaml
```diff
@@ -0,0 +1,25 @@
...
+- kapacitor
+- stream
+- etl
+- timeseries
+- influxdb
+- influxdata
+home: https://www.influxdata.com/time-series-platform/kapacitor/
```
afausti, line 13: The home, mantainers and engine fields can also be removed.

applications/sasquatch/charts/kapacitor/Chart.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: v1
+name: kapacitor
+version: 1.0.0
+appVersion: 1.6.4
```
afausti, line 4: We are using Kapacitor 1.7.7,  this can be specified here now that we are keeping the chart. See also the comment below on fixing the values.yaml with the defaults for sasquatch.

afausti (changes requested):
Main comment is that we should fix the chart values.yaml to keep the default values used in Sasquatch and to fix the documentation auto generated by helm-docs. 
Also, you can look at the Kafdrop chart as an example of external chart that was added to Sasquatch.

athornton:
I'm guessing we probably also want to use a more standard _helpers.tpl rather than importing the one from Kapacitor.

I was thinking about trying to upstream the changes to let us override the TOML, but given that your suggestions go in the opposite direction, and given that no one has merged anything to the chart in 2 years, I'm happy to let it go.

applications/sasquatch/values-idfdev.yaml
```diff
@@ -199,3 +199,8 @@ backup:
 
 tap:
   enabled: true
+
+# Send alerts to squadcast
+kapacitor:
+  enabled: true
+  squadcastURL: https://api.squadcast.com/v1/incidents/kapacitor/0352cecd87a2fe20cbeda558bf048320e6a0614b
```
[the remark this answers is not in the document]
rra: If this is like a Slack web hook, unfortunately I think it needs to be changed now that it's been made public. Otherwise, there is at least a trivial denial of service attack on whoever gets the Squadcast notifications. (If it requires additional authentication, then never mind.)

PR #4848, opened 2025-06-11
DM-51220: Update consdb to add can_see_sky and zernike coefficients

bbrondel:
* Updates hinfo in BTS, TTS, and summit
* Updates consdb tap to use ticket branch


ktlim (approved):
LGTM.  Tested at BTS and usdf-dev.

PR #4930, opened 2025-06-23
DM-51482: Hard-code HiPS URL - no templates in values files

gpdf:


gpdf:
I guess that's OK as long as we don't merge this branch back to `main` until then, OK.

PR #4942, opened 2025-06-23
DM-51494: Switch default DP1 HiPs coverage map to r band

dhirving:
We now have single-channel HiPS for DP1.  Frossie made an executive decision to switch to the r band for the default coverage map.  The r band is better than the colors we were using before, because it has fewer edge artifacts and is less confusing as a background for other things.


gpdf (approved):
LGTM, though I have not verified that the URL is actually functional

PR #4926, opened 2025-06-20
DM-51412: Enable dp1 hips data for datalinker on idfprod

stvoutsin:

PR #4970, opened 2025-06-26
DM-51571: Deploy DP1 HiPS to production

dhirving:
Configure datalinker and crawlspace to serve the DP1 HiPS maps at `idfprod`.


rra:
Does this subsume #4926? I'm guessing only one of the two PRs should be tracked. (cc @stvoutsin)

dhirving:
> Does this subsume #4926? I'm guessing only one of the two PRs should be tracked. (cc @stvoutsin)

The configuration in #4926 is incorrect so yeah, this would be a better one to go with.

PR #4982, opened 2025-06-27
[DM-51572] Increase interval between broker restarts during a Kafka rollout 

afausti:
The Strimzi operator uses the `readinessProbe` to check whether the brokers are ready to accept connections, and then proceed with restarting the brokers in sequence during a Kafka rollout. 
The interval between broker restarts is too short,  ~30s at BTS.  We found that it takes at least ~6  min for the Kafka brokers to finish rebalancing the leader partitions in the cluster after a broker is removed. That's because of the large number of partitions in these clusters (~1400 per broker) and the large number of consumer groups that need to stabilize.
~6min is longer than the Strimzi Operator default timeout of 5min, so we need to increase both the Strimzi Operator timeout and the `readinessProbe.initialDelaySeconds` to give more time for the brokers to initialize.
This PR is not to be merged yet but will help us to test the Control System stability at BTS during a Kafka rollout.
This was tested on idfdev with an initial delay of 4min which applies to both controllers and brokers because the readinessProbe configuration is done at the Kafka resource level and not at the KafkaNodepool level.

```
sasquatch-controller-3                            1/1     Running     0          9m54s
sasquatch-controller-4                            1/1     Running     0          18m
sasquatch-controller-5                            1/1     Running     0          14m
```

```
sasquatch-kafka-0                                 1/1     Running     0          9m49s
sasquatch-kafka-1                                 1/1     Running     0          5m35s
sasquatch-kafka-2                                 0/1     Running     0          77s
```

### DIFF
--- a/applications/sasquatch/README.md
+++ b/applications/sasquatch/README.md
@@ -603,6 +603,7 @@ Rubin Observatory's telemetry service
 | strimzi-kafka.mirrormaker2.resources | object | `{"limits":{"cpu":1,"memory":"4Gi"},"requests":{"cpu":"500m","memory":"2Gi"}}` | Kubernetes resources for MirrorMaker2 |
 | strimzi-kafka.mirrormaker2.source.bootstrapServer | string | None, must be set if enabled | Source (active) cluster to replicate from |
 | strimzi-kafka.mirrormaker2.source.topicsPattern | string | `"registry-schemas, lsst.sal.*"` | Topic replication from the source cluster defined as a comma-separated list or regular expression pattern |
+| strimzi-kafka.readinessProbe | object | `{}` | Readiness probe configuration for the Kafka brokers. |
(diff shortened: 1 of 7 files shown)


afausti:
Need to coordinate with Michael a time to test these changes on BTS.

afausti:
We have tested this today and delaying the broker rollouts helped with the control system stability. We still noticed missing heartbeats when the brokers terminate, though, caused by leader partition re-election. We are still deciding if we'll merge this PR or not and collected Kafka logs from the tests for further analysis.

afausti:
Adding an initial delay of ~10min to the readiness probe also delay the restart to the controllers. The total delay is thus 6*10min which becomes prohibitive. We decided to not adopt this solution. We won’t merge the PR but want to keep it around as a draft in case we need to revisit the code changes.

PR #4990, opened 2025-06-27
DM-51592: Change TAPQuerySetRunner to use async in mobu for idfdev

stvoutsin:
I think this should change `TAPQuerySetRunner` to use async. 
Targeting `idfdev` initially, if this works I think we should make this change on the other IDF environments as well now that we've changed TAP to have a 1 minute timeout for sync queries.

### DIFF
--- a/applications/mobu/values-idfint.yaml
+++ b/applications/mobu/values-idfint.yaml
@@ -94,6 +94,7 @@ config:
       business:
         type: "TAPQuerySetRunner"
         options:
+          sync: false
           query_set: "dp0.2"
         restart: true
     - name: "sia"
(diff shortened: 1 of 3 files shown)


applications/mobu/values-idfint.yaml
```diff
@@ -104,5 +104,6 @@ config:
       business:
         type: "SIAQuerySetRunner"
         options:
+          sync: false
```
fajpunk, line 107: Do we also want to do this for the `tap` business?

stvoutsin: Actually I pasted that at the wrong place, it should be in the TAP business not SIA

PR #4995, opened 2025-06-27
Remove filestore-backup

athornton:
There will be no mounted filestore anymore, hence nothing worth backing up.


fajpunk:
Do we need to manually delete any old backups now that this purger cron is going away?

PR #5022, opened 2025-06-30
DM-51654: Remove old direct Butler secrets from non-prod

rra:
Remove the old direct Butler secrets and the environment variable settings pointing to them from idfdemo, idfdev, and idfint Nublado.


athornton (approved):
Seems fine.  A cursory check of lsst.rsp suggests that everywhere we use these variables during startup, they are guarded by checks that they are in fact set, and I assume you've run nublado from this branch in dev enough to smoke test it and make sure it doesn't barf on startup.

PR #5023, opened 2025-06-30
Make the Kafka-based TAP (3.3.1) the default and pin other envs to 2.13.1

stvoutsin:
We've moved the idfs to use a Kafka (qserv-bridge) based approach so this PR sets the cadc-tap chart to use the latest version of that as the default. For other envs that do not support this yet, pin their version to the latest release of the legacy TAP service.


rra (approved):
Looks great, thank you!

PR #5027, opened 2025-07-01
DM-51665: Wind down influx-based infrastructure alerting in GKE envs

fajpunk:
Stop running telegraf-ds in any GKE-backed environment. Its purpose is now served by `google-cloud-observability` and Google-cloud-based infrastructure.


athornton (approved):
LGTM

PR #5036, opened 2025-07-02
Switch ukidacprod cert-manager to route53

gpfrancis:
Update the ukidacprod environment to use Let's Encrypt DNS and Route53 for TLS instead of using an external certificate.
Changes should be identical to commit 138db33 but for the ukidacprod environment.


stvoutsin (approved):
Looks good!

PR #5040, opened 2025-07-02
DM-51693: Change portal to only show 20 most recent queries

stvoutsin:


rra (approved):
This is fine for a quick fix, but maybe expose the setting in `values.yaml` and use a substitution variable here? That way we can tweak it easily (and per environment if needed).

PR #5099, opened 2025-07-14
tickets/DM-51655: move purger into nublado...

athornton:
...and also DRY out the nublado services filesystem config.

### DIFF
--- a/applications/nublado/templates/purger-cronjob.yaml
+++ b/applications/nublado/templates/purger-cronjob.yaml
@@ -46,14 +46,18 @@ spec:
               name: {{ .Values.purger.volumeName | quote }}
             - mountPath: /etc/purger
               name: config
+          # We need to extend this for PVCs too
           volumes:
           {{- range .Values.controller.config.lab.volumes }}
           - name: {{ .name | quote }}
-	  {{- if eq .source.type "nfs" }}
+          {{- if eq .source.type "nfs" }}
             nfs:
               server: {{ .source.server | quote }}
               path: {{ .source.serverPath | quote }}
-	  {{- end }}
+          {{- else if eq .source.type "hostPath" }}
+            hostPath:
+              path: {{ .source.path | quote }}
+          {{- end }}
           {{- end }}
           - name: config
             configMap:
--- a/applications/nublado/templates/purger-cronjob.yaml
+++ b/applications/nublado/templates/purger-cronjob.yaml
@@ -47,7 +47,12 @@ spec:
             - mountPath: /etc/purger
               name: config
           volumes:
-          {{- toYaml .Values.controller.config.lab.volumes | nindent 10 }}
+          {{- range .Values.controller.config.lab.volumes }}
+          - name: {{ .name | quote }}
+            nfs:
+              server: {{ .server | quote }}
+              path: {{ .serverPath | quote }}
+          {{- end }}
           - name: config
             configMap:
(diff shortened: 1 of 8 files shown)


applications/nublado/templates/tutorials-cronjobs.yaml
```diff
@@ -104,11 +111,19 @@ spec:
             - name: "GIT_BRANCH"
               value: {{ .Values.cronjob.artifacts.gitBranch | quote }}
             volumeMounts:
-            - mountPath: {{ .Values.cronjob.artifacts.targetVolume.mountPath }}
-              name: target
+            - mountPath: {{ .Values.cronjob.artifacts.targetVolume.mountPath | quote }}
+              name: {{ .Values.cronjob.tutorials.targetVolume.volumeName | quote }}
```
stvoutsin, line 115: Should this be:

```suggestion
              name: {{ .Values.cronjob.artifacts.targetVolume.volumeName | quote }}
```

applications/nublado/templates/purger-cronjob.yaml
```diff
@@ -0,0 +1,65 @@
...
+          {{- end }}
+          containers:
+          - name: purger
+            image: "{{ .Values.purger.image.repository }}:{{ .Values.purger.image.tag | default .Chart.AppVersion }}"
+            imagePullPolicy: "{{ .Values.purger.image.pullPolicy }}"
+            {{- with .Values.purger.resources }}
+            resources:
+              {{- toYaml . | nindent 14 }}
+            {{- end }}
+            {{- if .Values.controller.slackAlerts }}
```
stvoutsin, line 31: If we want to use the slackAlerts from the purger I think it would be:

```suggestion
            {{- if .Values.purger.slackAlerts }}
```

applications/nublado/values.yaml
```diff
@@ -725,15 +804,13 @@ cronjob:
     # -- Repository volume definition
     # @default -- See `values.yaml`
     targetVolume:
+      # -- Name of volume to mount (from controller.lab.config.volumes)
+      # @default -- None, must be set for each environment
+      volumeName: null
```
stvoutsin, line 809: Should the artifacts section match the tutorials?

applications/nublado/templates/tutorials-cronjobs.yaml
stvoutsin: Only relevant if the artifacts will also change to use a `volumeName`
I think this is a bug, but was the intention for tutorials and artifacts to share the same volume?

stvoutsin (approved):
I think this looks fine with just a couple of comments/questions.

applications/nublado/templates/tutorials-cronjobs.yaml
athornton: Oh.  They probably always will, but the config shouldn't assume that.

applications/nublado/templates/purger-cronjob.yaml
athornton: Let's just remove it from the purger as a separate config.  If the controller sends slack alerts so should the purger.

applications/nublado/values.yaml
athornton: In practice yes but not necessarily.  I think.

athornton: Yes, it should.

PR #5103, opened 2025-07-15
wobbly - Make uvicorn keep-alive timeout longer than ingress-nginx's

fajpunk:
`ingress-nginx` creates a pool of persistent connections to k8s services. These connections have the keep-alive timeout set to 60s. But default, the [Uvicorn keep-alive timeout](https://www.uvicorn.org/settings/#timeouts) is 5s.

This can cause a race condition where:
#. Uvicorn closes a TCP connection
# `ingress-nginx` hasn't received the TCP `FIN` yet and tries to send a data on that connection #. `ingress-nginx` gets an `RST` packet for a response because the connection is already closed #. `ingress-nginx` returns a 502: connection reset by peer response.


rra (approved):
Oh, you can just set this directly in the environment and uvicorn picks it up automatically? That's neat; I was thinking we were going to have to change the `Dockerfile`.

PR #5107, opened 2025-07-15
DM-51768: Support dax_apdb metrics in Prompt Processing

kfindeisen:
This PR adds a `dax_apdb` metrics configuration flag to Prompt Processing, and turns it on for the production instance.


kfindeisen:
The duplication is to work around a known issue with the documentation builder -- the files themselves inherit from `charts/*/values.yaml`, but the inherited values don't show up in the docs.

Mostly I wanted your feedback on the changes to `charts/prompt-keda/templates/scaled-job.yaml` and whether the environment patch is correct.

andy-slac:
> feedback on the changes to `charts/prompt-keda/templates/scaled-job.yaml` and whether the environment patch is correct

That look perfect, thanks!

PR #5114, opened 2025-07-17
[DM-51842] Grafana default role for new users

afausti:


afausti:
@fajpunk I've tested this [config ](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#auto_assign_org_role) on idfdev and new users have "Editor" role by default.

PR #5175, opened 2025-07-29
Add a PVC to hold memory profiles

fajpunk:


applications/qserv-kafka/values.yaml
```diff
@@ -126,9 +126,17 @@ frontend:
   # -- Affinity rules for the qserv-kafka frontend pod
   affinity: {}
 
-  # -- Whether to allow containers to run as root. Set to true to allow use of
-  # debug containers to diagnose issues such as memory leaks.
-  allowRootDebug: false
+  # -- Settings for debugging
```
rra, line 129: You only want documentation strings on the individual settings, not on the containers, or the helm-docs output and defaults will be weird.

applications/qserv-kafka/templates/frontend-deployment.yaml
```diff
@@ -61,6 +61,10 @@ spec:
                 secretKeyRef:
                   name: "qserv-kafka"
                   key: "redis-password"
+            {{- if .Values.frontend.debug.enabled | and .Values.frontend.debug.disablePymalloc }}
```
rra, line 64:
```suggestion
            {{- if (and .Values.frontend.debug.enabled .Values.frontend.debug.disablePymalloc) }}
```
seems a bit easier to read to me? Might be a matter of taste.

rra (approved):
Just a few minor nits.

PR #5201, opened 2025-08-05
DM-52053: qserv-kafka - periodic arq metrics

fajpunk:
Add a `CronJob` to periodically publish arq metrics using the script added in [this qserv-kafka PR](https://github.com/lsst-sqre/qserv-kafka/pull/151).

### DIFF
--- a/applications/qserv-kafka/templates/metrics-cronjob.yaml
+++ b/applications/qserv-kafka/templates/metrics-cronjob.yaml
@@ -1,0 +1,58 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: periodic-metrics
+  labels:
+    {{- include "qserv-kafka.labels" . | nindent 4 }}
+spec:
+  schedule: {{ .Values.periodicMetrics.schedule | quote }}
+  concurrencyPolicy: "Forbid"
+  jobTemplate:
+    spec:
+      template:
+        metadata:
+          labels:
+            {{- include "qserv-kafka.selectorLabels" . | nindent 12 }}
+            app.kubernetes.io/component: "periodic-metrics"
+            qserv-kafka-redis-client: "true"
+        spec:
+          restartPolicy: "Never"
+          containers:
+            - name: periodic-metrics
+              command: ["qserv-kafka", "publish-periodic-metrics"]
+              env:
+              {{- include "qserv-kafka.envVars" (dict "Values" .Values) | nindent 14 }}
+              envFrom:
+              - configMapRef:
+                  name: "qserv-kafka"
+              image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+              imagePullPolicy: {{ .Values.image.pullPolicy }}
+              resources:
+                {{- toYaml .Values.periodicMetrics.resources | nindent 16 }}
+              securityContext:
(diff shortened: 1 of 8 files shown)


applications/qserv-kafka/templates/metrics-cronjob.yaml
```diff
@@ -0,0 +1,95 @@
...
+                {{- if .Values.config.qservRestUsername }}
+                - name: "QSERV_KAFKA_QSERV_REST_PASSWORD"
+                  valueFrom:
+                    secretKeyRef:
+                      name: "qserv-kafka"
+                      key: "qserv-password"
+                {{- end }}
+              envFrom:
+                - configMapRef:
+                    name: "qserv-kafka"
```
rra, line 64: I'm sorry to make you suffer for my laziness, but we're now repeating this stuff three times, so it really should be moved into a helper macro in `_helpers.tpl` similar to `gafaelfawr.envVars` in the Gafaelfawr chart.

applications/qserv-kafka/values.yaml
```diff
@@ -240,6 +240,20 @@ resultWorker:
...
+
+  # -- Resource limits and requests for the qserv-kafka periodic metrics pods
+  # @default -- See `values.yaml`
+  resources:
+    limits:
+      cpu: "1"
+      memory: "300Mi"
+    requests:
+      cpu: "1"
+      memory: "145Mi"
```
rra, line 255: This is surely excessive, probably because it's a copy of the result worker metrics. I think it's fine to either tune the metrics for this particular job or to just blindly use whatever the resources for the result workers are, but I think we should probably pick one or the other, since right now it looks like we've measured and decided this job really needs that many resources.

Presumably we're going to have a bunch of these for different applications that use arq, so maybe it would be worth trying to arrive at a fairly standard set of resources, like we did with the Cloud SQL Proxy sidecar. Unfortunately, we don't have a great way of sharing those across services right now unless we figure out how to write a shared subchart, which is probably overkill and poses a bunch of other complexity issues.

applications/qserv-kafka/templates/metrics-cronjob.yaml
fajpunk: The periodic worker doesn't need most of this, it only needs redis connection config, metrics config, and a queue name to monitor. I can split the config out here (and in the app code) in a subsequent PR if that's OK with you. For now, I'll pull this config into a template and use it for all three components (frontend, backend, periodic metrics).

applications/qserv-kafka/values.yaml
fajpunk: Yeah, ideally, this periodic metrics worker is somehow a totally separate thing that gets deployed as a totally separate workload, and thus truly does have very consistent resource reqs across all apps that provision it.

For now, agreed it should be able to run with less resources. I tried to see the resources that the noteburst instance of this was using, and I can't find any Google metrics for it. I don't think it lives long enough to get sample.

applications/qserv-kafka/templates/metrics-cronjob.yaml
rra: Oh, in that case, whatever is easier for you! I thought you'd need all of it because you're loading the application config so it has to pass validation.

rra (approved):
Thanks, this looks great.

PR #5203, opened 2025-08-06
DM-49670: Deploy Butler writer service for prompt processing (standalone variant)

kfindeisen:
This PR adds a Butler writer service for Prompt Processing in an alternative architecture to #5079 . The service is enabled in `dev` but used only by LSSTCam-dev.


dhirving (commented):
The Butler writer service application setup and the environment setup for PP to use it look good to me.

kfindeisen:
The CI / docs failure doesn't seem to be transient after all; I've asked for help on `#square-docs-support`.

PR #5277, opened 2025-08-22
OSW-922: Update TTS config to move from NFS to buckets.

aibsen:


applications/auxtel/secrets-tucson-teststand.yaml
```diff
@@ -1,3 +1,11 @@
+oods-latiss-secret-key:
```
mareuter, line 1: Don't need this entry.

applications/simonyitel/secrets-tucson-teststand.yaml
```diff
@@ -1,3 +1,20 @@
+oods-comcam-secret-key:
```
mareuter, line 1: Don't need this entry.

applications/simonyitel/secrets-tucson-teststand.yaml
```diff
@@ -1,3 +1,20 @@
...
+  description: >-
+    S3 secret key for CCOODS oods-comcam user.
+butler-secret-key:
```
mareuter, line 4: Don't need this entry.

applications/uws/values-tucson-teststand.yaml
```diff
@@ -38,27 +41,13 @@ uws-api-server:
     exportPath: "/auxtel/repo/LATISS"
     subPath: ""
     readOnly: false
-  - name: repo-comcam
```
mareuter, line 41: The LATISS related entries can also go too.

mareuter (approved):
A few things to fix up, but otherwise I think we can proceed with this. Don't sync any apps after merging as we need to fixup the secrets.

PR #5286, opened 2025-08-25
DM-52298: Add schema v6.0 to registry and add new schema kafka topic

bsmartradio:
Additionally, add give brokers access to new topic.


erinleighh (approved):
Looks good!

PR #5287, opened 2025-08-25
Update Prompt Processing to 9.0.0 and Alert stream topic to lsst-alerts-v9.0.

erinleighh:


bsmartradio (approved):
Looks good, though I am surprised there is no docs updated with the new topic or image tag? I checked and it doesn't look like it auto generates, but that's maybe something to look into in the future.

PR #5325, opened 2025-09-03
Update the IP to the MPSky ephemerides service

hsinfang:


Gerenjie:
Looks good to me and I double-checked that the new URL is correct.

PR #5358, opened 2025-09-09
DM-52452: nublado - Specify explicit resource requests for user labs

fajpunk:
Set explicit resource requests for Nublado user lab sizes. Initially set them to 1/4 of their respective limits, which is what the Nublado controller did.

Goes with [this Nublado PR](https://github.com/lsst-sqre/nublado/pull/579/files)


rra (approved):
Looks great.

I think at some point we should change the size specification to not be a list and instead be a dict, with a separate list of enabled sizes, but we can do that separately.

PR #5399, opened 2025-09-18
Update s3 file notifications kafka address in PP-prod

hsinfang:


hsinfang:
This worked last night in PP-prod. 

Maybe it can be folded into https://github.com/lsst-sqre/phalanx/pull/5401 , or we just merge after that.

dspeck1 (approved):
LGTM.  I am merging [#5401] now.

PR #5416, opened 2025-09-24
add obsloctap to sasquatch

womullan:


womullan:
@afausti ok 0 but it has false at root for all others - you say set just this one true ?

afausti (approved):
Looks great, this should deploy the the obsloctap resources now.

PR #5418, opened 2025-09-24
DM-52590: gafaelfawr - update for new Safir sentry helpers

fajpunk:
Move all of the Sentry config into environment variables.

Goes with [this Gafaelfawr PR](https://github.com/lsst-sqre/gafaelfawr/pull/1354/files)


rra (approved):
Looks good, with the caveat about configuration path (`config.sentry` vs. `sentry`) and using `/health` as noted in the other review.

PR #5430, opened 2025-09-25
DM-52647: gafaelfawr - fix update-schema job

fajpunk:
Add the necessary env vars and volume mounts to the Gafaelfawr update schema job so that it works in environments with metrics enabled.


rra (approved):
Thanks for catching this!

PR #5443, opened 2025-09-29
SP-2630: Bump plot-navigator, adjust resources.

ctslater:
I will switch from the branch to the tagged version once the plot-navigator changes are merged.


ctslater:
Yes as intended, not sure it's ideal but just going on it empirically being better than before.

PR #5464, opened 2025-10-01
Add kafka users to the obsenv sasquatch application.

tribeiro:


afausti (approved):
Looks good to me. To add the password, someone needs to [sync the secrets for the environment](https://phalanx.lsst.io/admin/sync-secrets.html) and then sync Sasquatch in that environment to create the Kafka User.
In your application `secrets.yaml`  you can then [copy the password from Sasquatch](https://phalanx.lsst.io/developers/helm-chart/define-secrets.html).

PR #5465, opened 2025-10-01
OSW-1162: Change resource limits for eas-sim on BTS

bbrondel:
Changed to match summit resource limits.


rbovill (approved):
Looks good to me.

PR #5489, opened 2025-10-06
DM-52712: Add ArgoCD Styling

Fireye04:


fajpunk:
Ahh, we're not going to be able to do this with a simple switch like I had originally suggested. 

## Why can't we make a nice switch?

We're using an [off-the-shelf Helm chart](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/values.yaml) to install ArgoCD, which renders the Kubernetes resource templates for us based on the [Helm values](https://helm.sh/docs/chart_template_guide/values_files/) that we provide. We need to make any changes from [the style customization docs](https://argo-cd.readthedocs.io/en/stable/operator-manual/custom-styles) by overriding the [default chart values](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/values.yaml) with our own values, in our own values files. We cannot use any templating functionality in values files, which means that we can't conditionally set some values based on other values, or dynamically set a value to another value.

## What can we do?

The off-the-shelf chart should provide ways to make all of the changes that we need by setting values in our values files that will be merged with and override the default off-the-shelf chart values.. It won't be as nice as just setting a single value to `true` or `false`, but it'll work.

There are three values files that we need to think about:
* The [default chart values](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/values.yaml)
* The [values that we're setting in our Phalanx `argocd` chart](https://github.com/lsst-sqre/phalanx/blob/main/applications/argocd/values.yaml) that will override the off-the shelf-values. Anything under the `argo-cd:` key in that values file will be merged into the off-the-shelf default values.
* The [environment-specific values files in Phalanx for the `argocd` values](https://github.com/lsst-sqre/phalanx/blob/main/applications/argocd/values-idfdev.yaml) (`idfdev` in this link). These will be merged last into the combination of the Phalanx chart values and the off-the-shelf chart values.

The merging happens in this order, with values from the right overriding values from the left:
```
off-the-shelf defaults <-- Phalanx values.yaml <-- Phalanx values-idfdev.yaml
```
For any key that already exists, lists and simple values are replaced, and dicts are merged.
* If it is a simple value, then the value on left will be replaced with the value on the right
* If it is a list, the the value on the left will be replaced with the value on the right
* If it is a dict, then the value from each key on the left will be merged with the value at the same key on the right according to these rules (it's a deep merge)

## How do we apply the changes in the docs using these values?

(Almost) everywhere the docs show us updating a template with something, the off-the-shelf chart should provide us a way to set a Helm value that will result in those changes to the template when the chart gets rendered.

### [Custom styles](https://argo-cd.readthedocs.io/en/stable/operator-manual/custom-styles/#custom-styles)

It looks like the off-the-shelf chart comes with an abstraction of its own to get custom styles: [the config.styles key](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/values.yaml#L172). I think you would put the styles you want in that key in the environment-specific Phalanx values files of the environments you want to change. I think you'll have to paste the same CSS in each one, I don't think there is a way to abstract it :frowning_face: 

### [Banner content](https://argo-cd.readthedocs.io/en/stable/operator-manual/custom-styles/#banners)

For this, I think you'll need to get the relevant values into the `argocd-cm` configmap. The off-the-shelf chart provides a way to do this through the [configs.cm](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/values.yaml#L3738) value. It should work to put those `ui.banner*` values in there in the Phalanx environment-specific values files. It looks like we're already putting some stuff in there.

Fireye04:
Alright I've got the banner applied everywhere, Let me know which environments I should apply the stylesheet to :)

fajpunk (approved):
This looks great. Before merging, move the prod styles to the values files from the prod environments, which are, at least:

* idfprod
* roundtable
* usdfprod
* summit

PR #5500, opened 2025-10-07
DM-52580: Make the logging level for the Butler Writer Service configurable

kfindeisen:
This PR configures a single, global logging level for the Butler Writer service. The implementation of the logging level is in lsst-dm/prompt_processing_butler_writer#6.


srp3rd (approved):
Yes, this looks fine

PR #5995, opened 2026-01-30
Command to rebuild GKE clusters

fajpunk:
Add a command to rebuild GKE clusters. Remove other recover commands that do things that we don't need to do outside of a full cluster rebuild.

`phalanx recover restore` will restore data and workloads from a source GKE cluster to a destination GKE cluster. No data will be lost. Specifically, it will:

* Run  checks against the source cluster using `kubectl` and `argocd` to make sure it is in a good state for rebuilding
* Run checks against Google Cloud to make sure certain resources are provisioned correctly for the destination cluster to work
* Scale down all workloads and suspend all crons in the source cluster
* Release all of the public IPs for all LoadBalancer services in the
  source cluster
* Take a backup of the source cluster using Google Cloud's Backup for GKE product
* Restore any `PersistentVolumes` from the source cluster, along with their data, to the destination cluster
* Freshly install all workloads in the destination cluster using ArgoCD.

The ArgoCD install looks very similar to the existing new cluster install, but it is different enough that it gets it's own logic. Strimzi Kafka in particular needs certain resources across all applications installed in a certain order so that we don't lose any Kafka data.

### Review suggestions

This is a nasty PR, but not quite as nasty as it initially looks because much of the diff is data fixtures. I think the best way to review would be to start with the model files, then the storages, then the services. Then if you dare, look at the tests. I still don't like these tests.


fajpunk:
Here's an Anthropic Opus 4.7 review with the stock `/review` command in Claude Code:

### Overview

Large PR (87 files, +19,791 / −787; ~75% of additions are JSON test fixtures) that:

- Adds two new `phalanx recover` subcommands: `restore` (orchestrates a full backup-and-restore of a GKE cluster using Backup-for-GKE) and `preflight-check`.
- Adds new services (`GKERecoveryService`, `RecoveryEnvironmentService`, `GoogleCloudService`), new storage (`GoogleCloudAPIStorage` wrapping `google-cloud-gke-backup`, `google-cloud-compute-v1beta`,
`google-cloud-container`), a new ArgoCD Pydantic model module, and a `retryer` helper.
- Removes finer-grained `recover` subcommands (`suspend-crons`, `release-service-ips`, `pause-sasquatch-kafka-reconciliation`, etc.) — these become internal-only behaviours rolled into `scale-down` / `scale-up`
/ `restore`.
- Refactors `NamespacedResource` into a generic `Resource` base + `NamespacedResource` subclass; adds `PersistentVolume`, `Condition`/`Status`, and `is_ready()`.

Architecturally this looks good and the recovery sequencing reflects real operational knowledge (Strimzi cluster ID, sasquatch ordering, cert-manager retry). The bugs are localized and easy to fix; the two test
bugs are the most important since they hide future regressions in the new CLI commands.

---

### 🐛 Bugs (please fix)

- **`src/phalanx/services/environment.py` — `_argocd_login` calls `get_vault_client` twice.** Copy-paste: `vault = self._vault_storage.get_vault_client(...)` is duplicated; the second call shadows the first with
 an identical result. Delete one.

- **`src/phalanx/exceptions.py` — `GoogleCloudGKERestoreFailedError` says "Backup":**

    ```python
    msg = f"Backup {restore} failed."
    ```

    Should be `f"Restore {restore} failed."` — otherwise restore failures are reported as backup failures.

- **`tests/cli/recover_restore_test.py` — missing `test_` prefix:**

    ```python
    def restore_requires_both_contexts() -> None:
    ```

    Pytest will silently skip this; the "requires both contexts" cases for `restore` are never exercised. Rename to `test_restore_requires_both_contexts`.

- **`tests/cli/recover_preflight_check_test.py` — typo `prefilght-check`:**

    ```python
    args = ["recover", "prefilght-check", "--new-context", "fake-new-context"]
    ```

    Click exits 2 because the command doesn't exist, not because `--old-context` is missing — the assertion `exit_code == 2` passes for the wrong reason.

- **`src/phalanx/services/recovery_environment.py` — double sleep in `_wait_for_app_statuses`.** The `for _ in retryer(...)` already sleeps between iterations, but inside the loop you also call
`time.sleep(interval.total_seconds())` in the `except ValueError` branch. Each failed attempt waits `2 * interval`. Drop the explicit `time.sleep` and just `continue`.

- **`src/phalanx/services/cluster.py` — broad `except Exception: traceback.print_exc(); continue`** in `scale_down_workloads`, `scale_up_workloads`, `restore_service_ips`. This swallows programming errors
(`KeyError`, `AttributeError`, etc.) silently and continues scaling during a destructive recovery. Narrow to the specific exceptions you expect (`InvalidScaleStateError`, `InvalidLoadBalancerServiceStateError`,
`CommandFailedError`).

- **`src/phalanx/storage/command.py` — common_args order changed from prefix to suffix:**

    ```python
    args = args + tuple(self._common_args)
    ```

    Old: `kubectl --context X get pods`; new: `kubectl get pods --context X`. For `kubectl`/`helm`/`argocd` global flags this happens to work, but it's a non-obvious behaviour change — please add a comment
explaining why, or revert if there isn't a concrete reason.

### ⚠️ Likely bugs / worth a closer look

- **`src/phalanx/models/argocd.py` — magic literal `"sasquatch"`.** `ApplicationList.get_unsynced()` hardcodes `app.metadata.name == "sasquatch"` to apply special PVC-sync rules. You added a
`SASQUATCH_NAMESPACE` constant; please use a separate `SASQUATCH_APP_NAME` constant here (the app name and namespace are conceptually different).

- **`_wait_for_app_statuses` catches `ValueError`.** `list_applications_except_nublado_users` doesn't itself raise `ValueError`; `ApplicationList.model_validate(raw)` raises Pydantic's `ValidationError` (a
`ValueError` subclass) — so this *works* but is fragile. Catch `ValidationError` explicitly or have the storage layer raise a domain-specific exception.

- **`src/phalanx/storage/kubernetes.py` — `get_namespaced_resource` returns base `NamespacedResource`,** but `_wait_for_kafka` calls `.is_ready()` on the result. `is_ready()` lives on `Resource` so it works, but
 the model only validates base fields. Consider making this generic (`get_namespaced_resource[T](kind, name, ns, model: type[T]) -> T`).

- **`get_sasquatch_kafka_cluster_id` uses `kubectl run … --image foo --overrides …`.** Using `--image foo` and overriding via `--overrides` is a documented kubectl trick but very surprising — please add a
comment. Also `pvc_name = "data-0-sasquatch-kafka-0"` assumes a specific replica numbering; note in the docstring that this fails if replica 0's PVC isn't recoverable.

- **`gke_recovery._check_static_ips`** ignores `rubin-rag` in two places (`cluster.py` and `gke_recovery.py`). Pull into a constant like:

    ```python
    IGNORED_LOAD_BALANCER_NAMESPACES = frozenset({"rubin-rag"})
    ```

- **`recover_preflight_check` requires `--vault-role-id`, `--vault-secret-id`, `--run-id`** — but `run_id` isn't used in preflight at all. Drop `--run-id` from preflight.

### 🎨 Style / polish

#### Docstring copy/paste errors

- `models/kubernetes.py` `ConditionStatus`: "Valid values for the spec.type field of a Service." — wrong; this is about Condition status.
- `models/kubernetes.py` `PersistentVolume`: "kuberentes" → "kubernetes".
- `storage/google_cloud_api.py`: "comands" → "commands" (twice).
- `factory.py:create_environment_service` docstring has a `Parameters: context` section but the method takes no `context` arg.
- `factory.py` says label is `phalanx-run` — actually `phalanx-run-id` per constants.
- `cli.py`: "beging" → "begin".
- `storage/argocd.py`: "expect" → "expects" (`most functionality that uses this method expect`).
- `services/retryer.py` & `recovery_environment.py`: "The number of time to try" → "The number of times".
- `recovery_environment._wait_for_app_statuses` docstring: "statues" → "statuses".
- `storage/argocd.py` `set_helm_value`/`unset_helm_value`: "Set a helm value for an ArgoCD **project**" — should be **application**.

#### Other polish

- **`gke_recovery._check_static_ips`** — naming `_addresses` then `addresses = [a.address for a in _addresses]` is awkward; inline it. Several surrounding f-strings have double-spaces and a missing space after
`loadBalancerIP:`.

- **`cluster.scale_down_workloads` / `scale_up_workloads`** — `for w in sasquatch_workloads: workloads.remove(w)` is O(n²). Use:

    ```python
    sasquatch_workloads = [w for w in workloads if w.namespace == SASQUATCH_NAMESPACE]
    workloads = [w for w in workloads if w.namespace != SASQUATCH_NAMESPACE]
    ```

- **`print()` used heavily for user-facing output** in services, while the rest of the codebase mixes `click.echo` (CLI) and `print` (services). Not blocking; consider piping through Click for consistent
stream/encoding handling.

- **`recover restore` and `recover preflight-check` Click options are duplicated verbatim.** Extract a shared `@_recover_options` decorator so the two can't drift.

- **`storage/argocd.py:sync_all`** — the diff turns an inline `self._argocd.run(...)` into `args = [...]; self._argocd.run(*args)` with no behavioural change. Either revert or use `args` for branching (which it
doesn't).

### 🔐 Security

- Vault role-id and secret-id accepted from `VAULT_ROLE_ID` / `VAULT_SECRET_ID` env vars — good. Worth double-checking no current call sites log them.
- `gcloud auth application-default login` is implicitly required — please call this out in `recover restore --help` alongside the existing `argocd`/`helm`/`kubectl` checks.

### ✅ Test coverage

- Coverage of the new orchestration is solid (happy-path restore + each preflight failure variant), but assertions verify *call sequences against frozen JSON fixtures* — minor ordering tweaks force fixture
regeneration. Worth a future investment in higher-level intent-based assertions.
- `tests/support/recover.py:assert_run_calls` always looks under `recover/preflight-check/...` even for `restore` tests — `mock_happy_restore`'s fixtures end up under
`recover/preflight-check/restore-succeeds-*`. Confusing; move restore fixtures under `recover/restore/`.
- The two test bugs above (`prefilght-check` typo; missing `test_` prefix) materially reduce coverage of negative paths for both new commands.

### ⚡ Performance / runtime

- `wait_for_backup` defaults: 720 attempts × 5s = 60 min. Reasonable.
- `_wait_for_app_statuses` defaults: 120 × 5s = 10 min — but with the double-sleep bug it's actually 20 min worst case.
- `retain_pvs` issues one `kubectl patch` per PV serially; for clusters with many PVs this could be slow. Not blocking.

---

### Summary

Request changes for the bug list, then merge. The architecture is sound; the issues are localized.

src/phalanx/models/argocd.py
```diff
@@ -0,0 +1,187 @@
...
+    "Status",
+    "SyncStatus",
+    "YAMLModel",
+]
+
+
+class YAMLModel(BaseModel):
+    """A model that accepts camelCase keys."""
+
+    model_config = ConfigDict(alias_generator=to_camel, populate_by_name=True)
```
rra: You can use `safir.pydantic.CamelCaseModel` here, I think.

src/phalanx/models/kubernetes.py
```diff
@@ -68,6 +100,34 @@ class NamespacedResource(BaseModel):
...
+        return False
+
+
+class NamespacedResource(Resource):
+    """Properties common to all namespaced Kubernetes resources."""
+
+    namespace: str = Field(validation_alias=AliasPath("metadata", "namespace"))
+    """The namespace the resource is in."""
+
     def get_kind_name(self) -> str:
```
rra: This method doesn't use namespace so should probably stay as a method on the parent model.

src/phalanx/services/cluster.py
```diff
@@ -77,15 +80,50 @@ def scale_down_workloads(self) -> None:
...
+        for workload in sasquatch_workloads:
+            workloads.remove(workload)
+
+        # Scale down other workloads before sasquatch so we don't get errors
+        # about metrics.
         for workload in workloads:
-            self._check_workload_state(workload)
+            try:
+                self._check_workload_state(workload)
+            except Exception:
```
rra: It's safe to continue after any error checking workload state? If so, maybe capture this pattern in the called method or make a new method or add a flag if it's not always safe. This pattern repeats several times so it's a mild bit of duplicate code.

src/phalanx/services/cluster.py
```diff
@@ -136,9 +228,20 @@ def release_service_ips(self) -> list[Service]:
         new_services = []
         services = self._kubernetes.get_phalanx_load_balancer_services()
         for service in services:
+            if service.namespace == "rubin-rag":
```
rra: We'll never find this again when it's deep in the code and will get surprised by it later. Can we lift this up into a constant or a configuration setting in the environment values.yaml or something? Maybe a list of services to exclude from backup and restore or some component thereof?

src/phalanx/services/cluster.py
```diff
@@ -283,3 +384,11 @@ def _check_workload_state(self, workload: Workload) -> None:
...
             if workload.replicas != 0:
                 raise InvalidScaleStateError(workload)
+
+    def kube_version(self) -> None:
+        """Print the version of the kubectl client and kubernetes server.
+
+        Useful for checking that kubectl can connect to a cluster using a given
+        context.
+        """
+        self._kubernetes.version()
```
rra: Nit: This feels awkward -- it's running a thing down inside the stack and just assuming it writes to stdout. Maybe capture the output instead and return it up the stack to whatever is calling this so that the caller can make a decision about whether to print it, compare it to something, or whatever?

src/phalanx/services/gke_recovery.py
```diff
@@ -0,0 +1,311 @@
...
+    def _check_source_synced(self) -> list[str]:
+        """Make sure all ArgoCD apps in source cluster are in sync.
+
+        Returns
+        -------
+        list[str]
+            Any errors that this check found.
+        """
+        errors: list[str] = []
+        apps = self._source_environment.list_argocd_apps_except_nublado_users(
```
rra: As much as I appreciate the long method name :) I think this should probably take an exclude list that we get from a constant, because it's not just `nublado-users`, it's also `nublado-fileservers` (and in theory `ocps-uws-job`, although that will probably never run on GKE).

src/phalanx/storage/kubernetes.py
```diff
@@ -535,3 +575,126 @@ def set_service_to_load_balancer(self, service: Service) -> None:
             "--patch",
             '{"spec": {"type": "LoadBalancer"}}',
         )
+
+    def pause_sasquatch_kafka_reconciliation(self) -> None:
```
rra: This is a minor and kind of annoying nit, but in the name of keeping the storage layer as generic Kubernetes operations, I would call this `pause_kafka_reconciliation` and take the namespace in which to find the `Kafka` resources as an argument. That makes the decision to apply this specifically to Sasquatch a decision at the policy (service) layer instead of encoding it in the storage layer. Similarly below with the other operations. `get_sasquatch_kafka_cluster_id` is a bit awkward because the PVC name encodes Phalanx policy, but you could potentially pass that in from the service at least (and maybe move it into a constant so that it's more findable).

tests/data/input/applications/vault-secrets-operator/charts/vault-secrets-operator-2.7.0.tgz
rra, line 1: This looks like an accident that probably shouldn't be committed?

rra (dismissed):
Wow, this is an absolute ton of work! I left a few comments here and there, but overall I am all in favor of merging this. We can always fiddle with things like the test suite later, but this is great stuff. Thank you for doing this!

It would be great to add documentation for administrators about how to do the whole process (probably in a separate PR).

PR #6194, opened 2026-03-11
muster - New cached auth routes

fajpunk:
Add a new `GafaelfawrIngress` to muster with auth caching enabled. The initial usecase for this is Gafaelfawr performance testing.

Routes to the Muster path added in https://github.com/lsst-sqre/muster/pull/14/


fajpunk:
This was used for auth cache performance testing:
https://github.com/lsst-sqre/sqr-111/pull/4

rra:
Bringing this back to life as a draft PR because I don't want to lose the Varnish configuration, since I suspect we may need to take this approach to fix our ingress issues.

PR #6326, opened 2026-04-08
DM-54604: Enable prompt dataset in repertoire for idfev and idfint

stvoutsin:


rra (commented):
Unfortunately this can't be merged until all relevant Nublado containers are rebuilt with the fixed lsst.rsp.

PR #6342, opened 2026-04-10
DM-54621: Update RA resources and make charts consistent

mfisherlevine:


sebastian-aranda (approved):
Looks good! we need to check the redis memory request and limit, but let's keep until we have the upcoming changes and we can verify it again 👍 .

csilva-cl:
The intention of my redis comment was the opposite, the limits are too high. The pod has never used more than 1G, so reserving 25G is way too much.  I think 1G/5G is plenty.

mfisherlevine:
I'm in the middle of a big refactor where I might significantly change my Redis usage, so I don't think now is a great time to do the changing of the limits in that direction, but I certainly take your point that 50G is likely too high. On the other hand, we're much more CPU constrained than memory constrained, and this is a single pod not a set, so it's probably not that big of a deal. If it crashes, that brings all of Summit processing down completely, so I'm definitely erring on the side of caution for now. Once I've done the refactor, I will look at memory usage and take a wide safety margin and then reduce my usage.

PR #6360, opened 2026-04-15
DM-54003: Add Prompt Pub application

dspeck1:
Add prompt pub application and enable in production.  Development will be deployed with different PR.

### DIFF
--- a/applications/prompt-pub/Chart.yaml
+++ b/applications/prompt-pub/Chart.yaml
@@ -1,0 +1,8 @@
+apiVersion: v2
+appVersion: 0.1.0
+description: Manages the lifecycle of Butler Prompt Data Products as they move from the embargo repo to publication for end users.
+name: prompt-pub
+sources:
+- https://github.com/lsst-dm/prompt_publication_service
+type: application
+version: 1.0.0
--- a/applications/prompt-pub/README.md
+++ b/applications/prompt-pub/README.md
@@ -1,0 +1,40 @@
+# prompt-pub
+
+Manages the lifecycle of Butler Prompt Data Products as they move from the embargo repo to publication for end users.
+
+## Source Code
+
+* <https://github.com/lsst-dm/prompt_publication_service>
+
+## Values
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| affinity | object | `{}` | Affinity rules for the prompt-pub statefulset pod |
+| alloyDbProxy.config.instanceUri | string | `""` | Uri for Allow DB instance |
+| alloyDbProxy.image.pullPolicy | string | `"IfNotPresent"` | Pull policy for the allow db proxy image |
+| alloyDbProxy.image.repository | string | `"gcr.io/alloydb-connectors/alloydb-auth-proxy"` | Image to use for alloy db proxy |
+| alloyDbProxy.image.tag | string | `"1.14.2"` | Tag of image to use |
+| alloyDbProxy.resources | object | See `values.yaml` | Resource limits and requests for the alloy db proxy pod |
+| global.host | string | Set by Argo CD | Host name for ingress |
+| global.repertoireUrl | string | Set by Argo CD | Base URL for Repertoire discovery API |
+| global.vaultSecretsPath | string | Set by Argo CD | Base path for Vault secrets |
+| image.pullPolicy | string | `"IfNotPresent"` | Pull policy for the prompt-pub image |
+| image.repository | string | `"ghcr.io/lsst-dm/prompt_publication_service"` | Image to use in the prompt-pub deployment |
+| image.tag | string | The appVersion of the chart | Tag of image to use |
+| middleware.lsstResources.numWorkers | int | `4` | Sets the number of concurrent Butler file transfers per process.  Setting it explicitly here because this value is nondeterministic if you don’t assign something, but the exact value isn’t that critical. |
+| middleware.lsstResources.s3ProfileEmbargo | string | `"https://sdfembs3.sdf.slac.stanford.edu"` | S3 Profile for Embargo |
+| nodeSelector | object | `{}` | Node selection rules for the prompt-pub statefulset pod |
+| podAnnotations | object | `{}` | Annotations for the prompt-pub statefulset pod |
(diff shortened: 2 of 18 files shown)


applications/prompt-pub/templates/configmap.yaml
```diff
@@ -0,0 +1,36 @@
...
+              table: file_datastore_records
+        - datastore:
+            # Raws are served from a separate bucket.
+            name: raw
+            cls: lsst.daf.butler.datastores.fileDatastore.FileDatastore
+            root: s3://slac@rubin-lsstcam-raw/
+            records:
+              table: raw_datastore_records
+    registry:
+      db: postgresql://prompt_pub@172.24.10.74/prompt
```
dhirving: This URI should point to the Alloydb proxy sidecar, since that is how we're reaching Google.

applications/prompt-pub/templates/statefulset.yaml
```diff
@@ -0,0 +1,152 @@
...
+          - -c
+          - cp -L /app/db-auth-mount/db-auth.yaml /app/dbauth/ && chown 48045:4085 /app/dbauth/db-auth.yaml
+            && chmod u=r,go-rwx /app/dbauth/db-auth.yaml
+          resources: {}
+          volumeMounts:
+          - mountPath: /app/db-auth-mount
+            name: db-auth-mount
+            readOnly: true
+          - mountPath: /app/dbauth
+            name: db-auth-credentials-file
```
dhirving: I don't think we need this init container (or the db-auth file mounts) since you're using `LSST_DB_AUTH_CREDENTIALS` to provide the dbauth secret.

applications/prompt-pub/templates/statefulset.yaml
```diff
@@ -0,0 +1,152 @@
...
+                secretKeyRef:
+                  key: lsst_db_auth_credentials
+                  name: prompt-pub-secret
+            - name: LSST_RESOURCES_S3_PROFILE_embargo
+              value: {{ .Values.middleware.lsstResources.s3ProfileEmbargo }}
+        - name: alloydb-proxy
+          image: "{{ .Values.alloyDbProxy.image.repository }}:{{ .Values.alloyDbProxy.image.tag }}"
+          command:
+            - "/alloydb-auth-proxy"
+            - "--auto-iam-authn"
```
dhirving: Using `--auto-iam-authn` will need some corresponding terraform work to create the AlloyDB account and associate it with the service account.  In the meantime we can use the username/password I sent you yesterday.

applications/prompt-pub/templates/statefulset.yaml
```diff
@@ -0,0 +1,155 @@
...
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 48045 # rubinppb
+        runAsGroup: 4085 # rubin-users
+        fsGroup: 4085
+      containers:
+        - name: {{ .Chart.Name }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          command: ["python3", "-m"]
+          args: ["lsst.prompt_publication_service.service"]
```
dhirving: I'd rather use the defaults from the container for command/args, in case this changes.

applications/prompt-pub/templates/statefulset.yaml
```diff
@@ -0,0 +1,155 @@
...
+              subPath: credentials
+              readOnly: true
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
+        - name: alloydb-proxy
+          image: "{{ .Values.alloyDbProxy.image.repository }}:{{ .Values.alloyDbProxy.image.tag }}"
+          command:
+            - "/alloydb-auth-proxy"
+            - "--public-ip"
+            - "--address=0.0.0.0"
```
dhirving: I think we want to leave out --address here so that only localhost can connect.

applications/prompt-pub/README.md
```diff
@@ -0,0 +1,41 @@
+# prompt-pub
+
+Replicates prompt data products from the USDF to the US DAC.
```
dhirving:
```suggestion
Manages the lifecycle of Butler Prompt Data Products as they move from the embargo repo to publication for end users.
```

applications/prompt-pub/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Source Code
+
+* <https://github.com/lsst-sqre/prompt-pub>
+* <https://github.com/lsst-dm/prompt_publication_service>
```
dhirving:
```suggestion
* <https://github.com/lsst-dm/prompt_publication_service>
```

applications/prompt-pub/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+appVersion: 0.1.0
+description: Replicates prompt data products from the USDF to the US DAC.
+name: prompt-pub
+sources:
+- https://github.com/lsst-sqre/prompt-pub
+- https://github.com/lsst-dm/prompt_publication_service
```
dhirving:
```suggestion
- https://github.com/lsst-dm/prompt_publication_service
```

dhirving (dismissed):
Looks good, just a few minor nitpicks.

applications/prompt-pub/templates/statefulset.yaml
dspeck1: Agree we want to limit it.  An address is required, though, to connection outside of the container.  `127.0.0.1` did not work with the TCP based health checks since the health check comes from the Pod IP.  I discovered there is a `--health-check` flag in alloydb to create a health check endpoint.  I enabled the health check flag, restricted the alloydb connections to `127.0.0.1` and the health checks to the Pod IP.  During testing also discovered that the prompt pub container can start even if the alloydb container does not.  Added startup probe to both containers and a readiness probe to prompt pub that requires alloydb be ready.

applications/prompt-pub/templates/statefulset.yaml
dspeck1: Removed.

PR #6386, opened 2026-04-22
DM-54754: Enable Herald metrics in sasquatch application for idfdev and idfint

stvoutsin:
Changes to sasquatch config for idfdev and idfint to enable us to start sending metrics from Herald


afausti (approved):
Looks good! I think that's all you need to create the Kafka resources and configure Telegraf in Sasquatch.

PR #6395, opened 2026-04-24
DM-54773: Fix weka mount points for Prompt Publication

dhirving:
Fix an issue where the prompt publication service was mounting its Weka data volume at a different mount point than the "canonical" path used by interactive nodes.  This was causing it to write paths into the Butler database that were not reachable by end users.


dspeck1 (approved):
LGTM.  The correct locations are updated.

PR #6415, opened 2026-04-29
DM-54614: Add DP2 to TAP_SCHEMA

JeremyMcCormick:
These changes were based on the existing configuration for `dp1`.

### DIFF
--- a/applications/tap/values-idfint.yaml
+++ b/applications/tap/values-idfint.yaml
@@ -16,3 +16,13 @@ cadc-tap:
     instanceConnectionName: "science-platform-int-dc5d:us-central1:science-platform-int-8f439af2"
     serviceAccount: "tap-service@science-platform-int-dc5d.iam.gserviceaccount.com"
     database: "tap"
+
+  tapSchema:
+    type: "containerized"
+    image:
+      repository: "lsstsqre/tap-schema-idfint-tap"
+      tag: "tickets-DM-54614"
+      pullPolicy: "Always"
+    database: "TAP_SCHEMA"
+    username: "TAP_SCHEMA"
+    useVaultPassword: false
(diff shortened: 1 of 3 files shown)


stvoutsin (commented):
Changes look correct to me, although you may want to rebase now that we updated the schema version + template in the base values, so you should only need to update the schemaVersion.

Other than that, do we want to verify that this is the order we want the schemas to appear in the portal (dp2 after dp1)?

rra (dismissed):
Looks okay to me as well, with the note about schema ordering.

stvoutsin (approved):

applications/tap/values-idfint.yaml
```diff
@@ -16,3 +16,13 @@ cadc-tap:
     instanceConnectionName: "science-platform-int-dc5d:us-central1:science-platform-int-8f439af2"
     serviceAccount: "tap-service@science-platform-int-dc5d.iam.gserviceaccount.com"
     database: "tap"
+
```
stvoutsin: This looks good for using the old containerized approach as discussed rather than the repertoire-managed TAP_SCHEMA. We can work through what is required to support the new sdm_schemas changes and then switch back to the new approach once that is resolved.

stvoutsin (approved):

applications/tap/values-idfint.yaml
stvoutsin, line 12: If we have ObsCore under dp2 we'll also need to add dp2.ObsCore:access_url to this rule here for rewriting the URL of `access_url`

applications/tap/values-idfint.yaml
```diff
@@ -16,3 +16,13 @@ cadc-tap:
...
+  # tapSchema:
```
stvoutsin: Should we remove these comments before merging in?

JeremyMcCormick: I removed them.

PR #6430, opened 2026-05-04
nublado - Update mem reqs/limits and idle culler timeouts

fajpunk:
*DO NOT MERGE* this until after the idfprod cluster rebuild.

* The new cluster will have the highmem nodes, with 2x more memory per CPU.
* New idle timeouts in idfprod and idfint are 25 hrs

### DIFF
--- a/applications/nublado/values-idfprod.yaml
+++ b/applications/nublado/values-idfprod.yaml
@@ -105,6 +105,9 @@ hub:
   landingPage: "/lab/tree/notebooks/tutorials/welcome.md"
 
 jupyterhub:
+  cull:
+    timeout: 90000  # 25hrs
+
   hub:
     config:
       JupyterHub:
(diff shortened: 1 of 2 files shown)


applications/nublado/values-idfprod.yaml
```diff
@@ -50,6 +50,25 @@ controller:
...
+        # Double the memory because these are running on nodes with a 2x ratio
+        # of mem to CPU compared to standard nodes
+        - size: "small"
+          resources:
+            limits:
+              cpu: 1.0
+              memory: "8Gi"
+            requests:
+              cpu: 0.25
+              memory: "1.5Gi"
```
athornton, line 64: If we're keeping the ratio constant shouldn't this be 2Gi?

athornton (dismissed):
Fine, although I don't understand why 1.5Gi rather than 2Gi (but either one would work).

applications/nublado/values-idfprod.yaml
fajpunk: It's double the [default of `0.75Gi` in values.yaml](https://github.com/lsst-sqre/phalanx/blob/main/applications/nublado/values.yaml#L428)... but maybe the default should be `1Gi` and this should be `2Gi` like you're suggesting here?

PR #6439, opened 2026-05-07
DM-54783: Add Kafbat Kafka Management app to Sasquatch

dspeck1:
Add Kafbat Kafka management application to Sasquatch as a subchart.  Enable in Prompt Procesing dev and prod.  Brianna confirmed functionality meets needs for the Alert Stream Broker.

### DIFF
--- a/applications/sasquatch/charts/kafbat/templates/deployment.yaml
+++ b/applications/sasquatch/charts/kafbat/templates/deployment.yaml
@@ -1,0 +1,124 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kafbat
+  labels:
+    {{- include "kafbat.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      {{- include "kafbat.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "kafbat.selectorLabels" . | nindent 8 }}
+      annotations:
+        {{- with .Values.podAnnotations }}
+        {{ toYaml . | indent 8 }}
+        {{- end }}
+    spec:
+      automountServiceAccountToken: false
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 1000
+        runAsGroup: 1000
+      containers:
+        - name: kafbat
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
+          env:
+            - name:  AUTH_TYPE
+              value: "DISABLED"
+            - name: SERVER_SERVLET_CONTEXT_PATH
+              value: {{ .Values.server.servlet.contextPath | quote }}
+            - name: KAFKA_CLUSTERS_0_NAME
+              value: {{ .Values.cluster.name | quote }}
+            - name: KAFKA_CLUSTERS_0_BOOTSTRAPSERVERS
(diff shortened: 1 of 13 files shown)


applications/sasquatch/charts/kafbat/templates/deployment.yaml
```diff
@@ -0,0 +1,126 @@
...
+              value: {{ .Values.server.servlet.contextPath | quote }}
+            - name: KAFKA_CLUSTERS_0_NAME
+              value: {{ .Values.cluster.name | quote }}
+            - name: KAFKA_CLUSTERS_0_BOOTSTRAPSERVERS
+              value: {{ .Values.kafka.bootstrap | quote }}
+            - name: KAFKA_CLUSTERS_0_READONLY
+              value: "true"
+            - name: KAFKA_CLUSTERS_0_PROPERTIES_SECURITY_PROTOCOL
+              value: {{ .Values.kafbatUser.securityProtocol | quote }}
+            {{- if .Values.kafbatUser.kafkaSSL }}
```
afausti: I think this toggle is not necessary. The Kafka user is created with  TLS authentication so providing the truststore/keystore settings is the only option here.

applications/sasquatch/charts/kafbat/templates/deployment.yaml
```diff
@@ -0,0 +1,126 @@
...
+          volumeMounts:
+            - name: cluster-ca
+              mountPath: /etc/kafka/secrets
+            - name: user-certs
+              mountPath: /etc/kafka/user-secrets
+              readOnly: true
+      volumes:
+      - name: cluster-ca
+        secret:
+          secretName: sasquatch-cluster-ca-cert
```
afausti: This gets the CA cert from the Strimzi cluster, I would suggest using the `{{ .Values.cluster.name }}` setting instead of hardcoding sasquatch.

applications/sasquatch/charts/kafbat/values.yaml
```diff
@@ -0,0 +1,99 @@
...
+  # -- Kafka topic prefixes to filter topics by
+  # @default -- ["lsst"]
+  topicPrefixes:
+    - "lsst"
+
+kafbatUser:
+  # -- Kafka Security Protocol for user
+  securityProtocol: "SSL"
+  # -- Enables trust and key store SSL settings for with Kafka User
+  kafkaSSL: true
```
afausti: I think this toggle can be removed, see comment above.

applications/sasquatch/charts/kafbat/values.yaml
```diff
@@ -0,0 +1,99 @@
...
+  servlet:
+    # -- The context path to serve requests on
+    contextPath: /kafbat
+
+# -- The endpoint of Schema Registry
+schemaRegistry: "http://sasquatch-schema-registry.sasquatch:8081"
+
+service:
+  # -- Additional annotations to add to the service
+  annotations: {}
```
afausti, line 62: I notice that service annotations are not used in service.yaml so this setting doesn't take effect.

afausti (dismissed):
This is great, thanks for adding the Kafbat UI! 
Spot some minor inconsistencies.

applications/sasquatch/charts/kafbat/templates/deployment.yaml
dspeck1: I agree.  Removed.

applications/sasquatch/charts/kafbat/templates/deployment.yaml
dspeck1: Thanks for catching.  Updated.

applications/sasquatch/charts/kafbat/values.yaml
dspeck1: Updated.

PR #6442, opened 2026-05-07
DM-54177: Update schema to version v11 and remove v10.0

bsmartradio:
Remove the v10.0 topic from prod

Topic is unused and needs to be removed

Update schema tag


dspeck1 (approved):
LGTM

PR #6462, opened 2026-05-13
DM-54874: Add Cronjobs to scale up and down Prompt Pub Replicas

dspeck1:
Add cronjobs, service account, role, and role bindings to scale up and down the Prompt Publication statefulset on a schedule  to avoid running during observing.


dhirving (approved):
Looks good to me

PR #6469, opened 2026-05-15
DM-54950: Convert to safir.testing.data from syrupy

rra:
Use the new safir.testing.data class for loading test data and comparing expected results, replacing the experiment with syrupy. This makes Phalanx match our other applications and replaces the various local functions for loading test data, and will unblock some other improved sophistication for tests later.


fajpunk (approved):
Nice, this is a straightforward conversion!

PR #6472, opened 2026-05-15
DM-54968: Assign some Renovate PRs to owners

rra:
Assign the Renovate PRs for some packages that SQuaRE doesn't handle to owners who should be processing them.


rra:
Merged with a different branch name since the Renovate syntax checking requires a magical branch name.

PR #6482, opened 2026-05-20
fov-quicklook: update image tag to support data query

michitaro:
Update image tag to support data query.

### DIFF
--- a/applications/fov-quicklook/values.yaml
+++ b/applications/fov-quicklook/values.yaml
@@ -12,7 +12,7 @@ image:
   pullPolicy: Always
 
   # -- Tag of image to use
-  tag: main
+  tag: fov-quicklook-local-issue-10-differnce-image
 
 config:
   # -- URL path prefix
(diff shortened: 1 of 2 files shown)


applications/fov-quicklook/values.yaml
```diff
@@ -12,7 +12,7 @@ image:
   pullPolicy: Always
 
   # -- Tag of image to use
-  tag: main
+  tag: fov-quicklook-local-issue-10-differnce-image
```
gpdf, line 15: I understand that you are planning to update this PR and are not ready to merge it.  Just for reference when you do that: “differnce” was misspelled in the tag.  This isn’t a functional issue.

More importantly, I don’t think have a non-“main”, non-version-number tag as the tag in the `values.yaml` file is not really standard practice for the way we use Phalanx.  If you’re going to use a “temporary” tag, that would be as an override, for example, in `values-idfdev.yaml`, but not in the main `values` file.

gpdf (changes requested):

PR #6532, opened 2026-06-03
DM-55137: Configure USDF dev environment to use external PG for all TAP utility databases

stvoutsin:
Configures all tap applications for `usdf-dev` to use external postgres for  the UWS and TAP_SCHEMA databases.
In the process updates repertoire configuration which now manages TAP_SCHEMA updates for those apps.

In the changes there is also a minor correction to repertoire's job update helm hook to remove a user env var that is obsolete.

### DIFF
--- a/applications/repertoire/values.yaml
+++ b/applications/repertoire/values.yaml
@@ -524,6 +524,13 @@ config:
         databaseUrl: "postgresql://obsforgetap@127.0.0.1:5432/obsforgetap"
         databasePasswordKey: "obsforgetap-database-password"
 
+      consdbtap:
+        enabled: false
+        schemas:
+          - "efd"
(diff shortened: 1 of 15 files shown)


applications/repertoire/values.yaml
```diff
@@ -524,6 +524,13 @@ config:
         databaseUrl: "postgresql://obsforgetap@127.0.0.1:5432/obsforgetap"
         databasePasswordKey: "obsforgetap-database-password"
 
+      consdbtap:
+        enabled: false
+        schemas:
+          - "efd"
```
JeremyMcCormick: If this corresponds to a YAML in `sdm_schemas`, I don't think this is valid. It would be the full list of schemas from `cdb_*.yaml` and `efd_*.yaml`.

PR #6543, opened 2026-06-04
DM-55158: Add tables API versions for TAP servers

rra:
Add the tables endpoints to the API versions list for TAP servers in the Repertoire service discovery information. This should fix sending credentials to those URLs with the current version of PyVO and a corresponding fix to lsst.rsp.

### DIFF
--- a/applications/repertoire/values.yaml
+++ b/applications/repertoire/values.yaml
@@ -195,6 +195,10 @@ config:
         datasets:
           - "efd"
         template: "https://{{base_hostname}}/api/consdbtap"
+        versions:
+          tables:
+            template: "https://{{base_hostname}}/api/consdbtap/tables"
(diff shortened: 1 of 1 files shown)


applications/repertoire/values.yaml
```diff
@@ -195,6 +195,10 @@ config:
         datasets:
           - "efd"
         template: "https://{{base_hostname}}/api/consdbtap"
+        versions:
+          tables:
+            template: "https://{base_hostname}}/api/consdbtap/tables"
```
stvoutsin: Looks like there is a missing opening brace here and in the entries below

stvoutsin (changes requested):

applications/repertoire/values.yaml
rra: Whoops thanks

stvoutsin (approved):
Changes all look good to me

PR #6544, opened 2026-06-04
Disable Herald on idfprod

stvoutsin:
Temporarily disable Herald from prod until prompt dataset is available

### DIFF
--- a/environments/values-idfprod.yaml
+++ b/environments/values-idfprod.yaml
@@ -14,7 +14,7 @@ onepassword:
 vaultPathPrefix: "secret/phalanx/idfprod"
 
 applications:
-  herald: true
+  herald: false
   butler: true
   datalinker: true
   ghostwriter: true


rra (approved):

gpdf:
> disable Herald from prod until prompt dataset is available

Is this really the reason?  I think there's no harm in supporting alert retrieval without either the Prompt images or the PPDB catalogs, though of course it will become much more useful later on.

I understood from Frossie that the concern was that it would be preferable to release the service in conjunction with a Portal that can use it successfully.  Did I have that right?

In any event, I'm not disagreeing with the outcome.

stvoutsin:
> > disable Herald from prod until prompt dataset is available
> 
> Is this really the reason? I think there's no harm in supporting alert retrieval without either the Prompt images or the PPDB catalogs, though of course it will become much more useful later on.
> 
> I understood from Frossie that the concern was that it would be preferable to release the service in conjunction with a Portal that can use it successfully. Did I have that right?
> 
> In any event, I'm not disagreeing with the outcome.

Just to clarify as that description wasn't very clear, we have to enable a `prompt` flag in our service discovery app (repertoire) so this works from a notebook `get_service_url("alerts", "prompt")`, otherwise that would fail and we'd get mobu notebook failures. 
Once we are able to release the new Portal then we'll release Herald along with it with that dataset enabled in repertoire.

PR #6559, opened 2026-06-09
DM-55187: Change IVOA registry default pathPrefix to /api

stvoutsin:
Target of this PR is to allow moving the publishing registry path to `/api/registry` instead of  `/discovery/ivoa`.

### DIFF
--- a/applications/repertoire/templates/ingress.yaml
+++ b/applications/repertoire/templates/ingress.yaml
@@ -27,7 +27,7 @@ template:
                   port:
                     number: 8080
             {{- if .Values.config.ivoaRegistry }}
-            - path: {{ .Values.config.ivoaRegistry.pathPrefix | quote }}
+            - path: "{{ .Values.config.ivoaRegistry.pathPrefix }}/registry"
               pathType: "Prefix"
               backend:
                 service:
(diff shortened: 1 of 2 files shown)


stvoutsin:
One thing I'm not sure about is if we need an ingress rule for the pathPrefix or not. Should I add it to the ingress as a conditional? Is there any harm if that gets added for /api/ which is already created elsewhere?
If we exclude, anyone who uses a different pathPrefix than /api/ would not get an ingress rule added for it

rra:
We don't create `/api` elsewhere, just paths underneath it, so we do need an ingress rule, but it can just be a path on the existing ingress. So I think the one we already have that you added yesterday will do it.

rra (dismissed):

stvoutsin:
> We don't create `/api` elsewhere, just paths underneath it, so we do need an ingress rule, but it can just be a path on the existing ingress. So I think the one we already have that you added yesterday will do it.

I think I need to make the optional path include the hardcoded /registry path though so that we're not claiming the base /api path? So something like:

`- path: "{{ .Values.config.ivoaRegistry.pathPrefix }}/registry"`

rra:
Oh, now I see what you mean. Yes, that's completely correct.

rra:
This was handled in #6572.

PR #6603, opened 2026-06-18
Enable Herald of idfprod and add prompt dataset to repertoire config for prod

stvoutsin:
Changes
- Set `herald: true` in environments/values-idfprod.yaml
- Add "prompt" to datasets list for `repertoire/values-idfprod.yaml`


rra (approved):
Looks good to me!

PR #6604, opened 2026-06-18
DM-54614: Enable dp2 only for idfint

JeremyMcCormick:


applications/repertoire/values-idfint.yaml
```diff
@@ -18,6 +18,8 @@ config:
         enabled: true
       ssotap:
         enabled: true
+    schemas:
```
stvoutsin: I think you need to include all schemas here, and also the nesting needs to be:

```
servers:
  tap:
    enabled: true
    schemas:
      - dp2
      - dp1
      - ivoa_obscore
      - dp02_dc2
```

JeremyMcCormick: @stvoutsin Should be fixed now.

stvoutsin (approved):
Look good to me!

rra (approved):

PR #6623, opened 2026-06-24
DM-55274: Update IVOA Publishing Registry content with Catalog Resource records for TAP entries

stvoutsin:
## Changes
- Add one Catalog Resource record per dataset served by our TAP service that we want to advertise.
- Remove GMS ivoaRegistry block (Don't think we need to have a record for the GMS endpoint)

### DIFF
--- a/applications/repertoire/values.yaml
+++ b/applications/repertoire/values.yaml
@@ -142,12 +142,17 @@ config:
     creator: "Vera C. Rubin Observatory"
     ivoid: "ivo://org.rubinobs/registry"
     path: "/api/registry"
-    repositoryName: "Rubin Science Platform IVOA Publishing Registry"
+    repositoryName: "Rubin IVOA Publishing Registry"
     rights: >-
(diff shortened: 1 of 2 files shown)


applications/repertoire/values.yaml
```diff
@@ -374,15 +387,27 @@ config:
             ivoaStandardId: "ivo://ivoa.net/std/VOSI#tables"
         ivoaRegistry:
           ivoaServiceType: "tap"
-          ivoid: "ivo://org.rubinobs/tap"
-          title: "Rubin Science Platform TAP Service"
+          ivoid: "ivo://org.rubinobs/qserv-tap"
```
stvoutsin, line 399: This name is pending a decision from Gregory.

stvoutsin:
Merging this as it is, though it will require one more round of discussion on the right IVOID to use for the TAP service. (We do not publish the VO Registry yet anyway so this is still all internal)

PR #6627, opened 2026-06-25
DM-55317: Configure SIA on data-int for early DP2

dhirving:
Add an SIA endpoint for the "early" DP2 release on data-int.


stvoutsin (approved):
Change looks good to me!

PR #6647, opened 2026-06-29
Remove CST group writeability from tutorials/data cronjobs

athornton:


fajpunk:
We talked offline about if we need to do anying chown-ing manually after this, you said maybe and you'd do it :)

fajpunk (approved):

PR #6653, opened 2026-07-01
Add couger01 to developer role list at usdf-dev.

couger01:


villarrealas:
Eric is dev for weatherbroadcaster, so approving this from the USDF end.

PR #6660, opened 2026-07-02
Add (soft)pod anti-affinity to love, envsys and simonyitel summit and BTS deployment.

tribeiro:
These updates aim to improve the Kubernetes scheduling of these systems pods and try to prevent them from landing all on the same node as we were experiencing at the summit and BTS.


aibsen (approved):
looks good to me :)

PR #6678, opened 2026-07-08
DM-55425: Add auto scaling to ingester

bsmartradio:
The ingester currently cannot read the alert stream fast enough for when we are at full capacity. We are adding auto scaling to help read out faster.

### DIFF
--- a/applications/sasquatch/charts/alert-database/values.yaml
+++ b/applications/sasquatch/charts/alert-database/values.yaml
@@ -48,6 +48,46 @@ ingester:
   # anything else to suppress logging.
   logLevel: verbose
 
+  # -- Seconds Kubernetes waits after SIGTERM before sending SIGKILL. Should
+  # be long enough for the ingester to finish processing its current batch,
+  # commit offsets, and leave the Kafka consumer group cleanly.
+  terminationGracePeriodSeconds: 60
+
+  # -- Number of static replicas when autoscaling is disabled.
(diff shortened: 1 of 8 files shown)


applications/sasquatch/charts/alert-database/values.yaml
```diff
@@ -48,6 +48,41 @@ ingester:
...
+    # -- Period to wait after the last trigger fires before scaling down (seconds).
+    cooldownPeriod: 300
+
+    # -- Kafka consumer group to monitor for lag. Must match the consumer
+    # group used by the ingester application.
+    consumerGroup: "alertdb-ingester"
+
+    # -- Kafka offset reset policy for the KEDA scaler if the consumer group
+    # has no committed offset.
+    offsetResetPolicy: "latest"
```
dspeck1, line 89: Would `earliest` be better scaling behavior?  It would look at all messages currently not processed vs the ones added after it starts.

dspeck1:
How long does each process take?  With a scaled object when the queue is empty KEDA will scale down pods. Since Kubernetes/KEDA does not know that a task is processing it picks a pod at random.  `terminationGracePeriodSeconds` can be adjusted to run as long as the last processing or to support long running jobs move to a Scaled Job.

dspeck1 (dismissed):
LGTM.  One question on setting of offset.  Please cleanup git history before merging.

PR #6680, opened 2026-07-08
DM-55301: PPDB Replication Cloud SQL Proxy

dspeck1:
Add cloud sql proxy to PPDB replication and values for dev.

### DIFF
--- a/applications/ppdb-replication/templates/exporter-deployment.yaml
+++ b/applications/ppdb-replication/templates/exporter-deployment.yaml
@@ -21,6 +21,12 @@ spec:
         {{- end }}
     spec:
       volumes:
+        - name: gcp-credentials-file
+          secret:
+            secretName: ppdb-replication
+            items:
+              - key: gcp-credentials.json
+                path: gcp-credentials.json
         - name: "ppdb-replication-secrets-raw"
           secret:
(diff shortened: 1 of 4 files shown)


JeremyMcCormick:
Notes from meeting:

- You may use the existing Vault key with a new version for credentials rather than making a new one.
- The existing credentials path can be changed from `gcs-credentials.json` to `gcp-credentials.json` to make it clearer.

applications/ppdb-replication/templates/exporter-deployment.yaml
```diff
@@ -21,6 +21,12 @@ spec:
         {{- end }}
     spec:
       volumes:
+        - name: cloudsql-credentials-file
```
JeremyMcCormick: I think we need to use the existing credentials key in Vault or it will be duplicated, unless you think there is a good reason to keep them separate.

JeremyMcCormick (approved):
This looks good to me. I have not tested it, but I have to make some other changes on top of this in order to test effectively. So if there are issues with it, we can fix in an additional patch PR.

applications/ppdb-replication/templates/exporter-deployment.yaml
dspeck1: Updated credentials to use the same credentials for both.

PR #6737, opened 2026-07-21
DM-55570: Add a new Gafaelfawr scope write:obsforge

afausti:
The Prompt publication service runs outside the RSP and needs to send notifications to ObsForge.
To grant access to the ObsForge external API we’ll use a new scope `write:obsforge`.
ObsForge is deployed on IDF so group mapping is needed only on those environments.

### DIFF
--- a/applications/gafaelfawr/values-idfprod.yaml
+++ b/applications/gafaelfawr/values-idfprod.yaml
@@ -100,6 +100,8 @@ config:
       - "g_users"
     "write:sasquatch":
       - "g_admins"
+    "write:obsforge":
+      - "g_admins"
 
   initialAdmins:
     - "adam"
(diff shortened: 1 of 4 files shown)


applications/gafaelfawr/values-idfprod.yaml
```diff
@@ -100,6 +100,9 @@ config:
       - "g_users"
     "write:sasquatch":
       - "g_admins"
+    "write:obsforge":
+      - "g_admins"
+      - "g_developers"
```
rra: We usually don't grant access to `g_developers` on prod, but no problem if it's needed for your specific use case.

rra (dismissed):

applications/gafaelfawr/values-idfprod.yaml
afausti: I think I can remove `g_developers`. 
An admin can create a service token and hand it over to Prompt publication developers so that should be fine.

PR #6749, opened 2026-07-23
DM-55593: Update USDAC/IDF configs for DP2 HiPS/MOC datasets

gpdf:
Main tasks:

- Use DP2 color HiPS for coverage
- Make HiPS list phalanx-configurable
- Use DP2 HiPS list for idfprod and idfint

Also, clarify that the phalanx property involved does in fact control the coverage HiPS

### DIFF
--- a/applications/portal/README.md
+++ b/applications/portal/README.md
@@ -15,6 +15,7 @@ Rubin Science Platform Portal Aspect
 | config.alertMessage | string | `nil` | If not null, display this alert to all users in a banner |
 | config.cleanupInterval | string | `"36h"` | How long results should be retained before being deleted |
 | config.debug | string | `"FALSE"` | Set to `TRUE` to enable service debugging |
+| config.hipsListUrl | string | `/api/hips/list` in the local Science Platform | URL for master list of Rubin-featured HiPS datasets |
 | config.hipsUrl | string | `/api/hips/images/color_gri` in the local Science Platform | URL for default HiPS map to be used for coverage displays |
 | config.livetap | string | `""` | Endpoint under `/api/` for the live TAP service on the instance, if present |
 | config.showUserInfo | string | `"true"` | Whether to show information about the logged-in user |
--- a/applications/portal/README.md
+++ b/applications/portal/README.md
@@ -15,7 +15,7 @@ Rubin Science Platform Portal Aspect
 | config.alertMessage | string | `nil` | If not null, display this alert to all users in a banner |
 | config.cleanupInterval | string | `"36h"` | How long results should be retained before being deleted |
 | config.debug | string | `"FALSE"` | Set to `TRUE` to enable service debugging |
-| config.hipsUrl | string | `/api/hips/images/color_gri` in the local Science Platform | URL for default HiPS service |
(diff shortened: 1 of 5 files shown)


applications/portal/templates/statefulset.yaml
```diff
@@ -48,7 +48,12 @@ spec:
             - name: "PROPS_alert__viewer__fits__service__url"
               value: "{{ .Values.global.baseUrl }}/api/alerts"
             - name: "PROPS_lsst__hips__masterUrl"
-              value: "{{ .Values.global.baseUrl }}/api/hips/list"
+              value: >-
+                {{- if .Values.config.hipsListUrl }}
+                       {{ .Values.config.hipsListUrl }}
+                {{- else }}
+                       {{ .Values.global.baseUrl }}/api/hips/list"
+                {{- end }}
```
robyww, line 63: you might also want to do this for `PROPS_lsst__hips__masterUrl`. I just added it so you will have to rebase to main

applications/portal/README.md
```diff
@@ -15,7 +15,8 @@ Rubin Science Platform Portal Aspect
 | config.alertMessage | string | `nil` | If not null, display this alert to all users in a banner |
 | config.cleanupInterval | string | `"36h"` | How long results should be retained before being deleted |
 | config.debug | string | `"FALSE"` | Set to `TRUE` to enable service debugging |
-| config.hipsUrl | string | `/api/hips/images/color_gri` in the local Science Platform | URL for default HiPS service |
+| config.hipsListUrl | string | `/api/hips/list` in the local Science Platform | URL for master list of Rubin-featured HiPS datasets |
+| config.hipsUrl | string | `/api/hips/images/color_gri` in the local Science Platform | URL for default HiPS map to be used for coverage displays |
```
robyww, line 19: same here.

applications/portal/values-idfint.yaml
```diff
@@ -12,7 +12,8 @@ config:
         size: "20Gi"
         storageClass: "standard-rwo"
   ssotap: "ssotap"
-  hipsUrl: "https://data-int.lsst.cloud/api/hips/v2/dp1/deep_coadd/band_r"
+  hipsListUrl: "https://data-int.lsst.cloud/api/hips/v2/dp2/list"
+  hipsUrl: "https://data-int.lsst.cloud/api/hips/v2/dp2/color_gri"
```
robyww, line 16: and here, and the other places

robyww (dismissed):
Looks fine. I am not sure why it is necessary to expand it beyond the other way but if it is then you should probably make a `mocListUrl`

applications/portal/templates/statefulset.yaml
gpdf: You mean, for the `moc__masterUrl`, I think?  Yes, I'll do that now.

applications/portal/README.md
gpdf: I'm not changing the global default at this time.  I'm not sure how that would affect other RSP instances.

robyww (approved):

gpdf:
> Looks fine. I am not sure why it is necessary to expand it beyond the other way but if it is then you should probably make a `mocListUrl`

It's for controllability at other RSP sites, and to allow a new list to be deployed only on, e.g., data-int, for an extended period without having to use a branch deploy.

PR #6783, opened 2026-07-27
DM-55623: Add idfprod mobu users to g_rubin group

rra:
Allow them to bypass the user blocks while testing the DP2 release.


rra:
Was only needed during the testing period, does not need to be merged.

PR #6794, opened 2026-07-28
Temporarily grant access to lsst.sal namespace

afausti:
- For testing at BTS the camera kafka user requires access to write to lsst.sal namespace


aibsen (commented):
looks correct to me.

PR #6799, opened 2026-07-29
USDFSM-112: Add local ZFS /lscratch for USDF RSP notebooks

atanikan:
Mount zfs-lscratch-storage at /lscratch alongside existing NFS /scratch so labs get fast node-local scratch without breaking shared scratch users. Sized PVCs at 10Gi.


dspeck1 (dismissed):
LGTM.  Please let me know when the issues are resolved on the node/ZFS side and I can merge to prod.

PR #6810, opened 2026-07-30
DM-55637: Update to EDP2 v2 schema: improve ObsCore column ordering

gpdf:


stvoutsin (dismissed):
Change looks good to me, this will indeed update TAP_SCHEMA to version of `EDP2-deploy-v2` for all envs that are synced and don't override this.

stvoutsin:
As discussed via Zoom probably a good idea to also update the datalink snippets URLs for datalinker and tap for consistency in this PR.

stvoutsin (dismissed):
Looks good!

PR #6814, opened 2026-08-03
DM-55687: Change PPDB Replication to CloudSQL Python Connector

dspeck1:
Remove Cloud SQL Proxy and change to using CloudSQL Python Connector.


applications/ppdb-replication/values-usdfdev.yaml
```diff
@@ -1,9 +1,6 @@
-cloudSqlProxy:
-  image:
-    tag: "2.23.0"
+image:
 
-  config:
-    instanceUri: "ppdb-dev-5c07:us-central1:ppdb-dev"
+  tag: tickets-DM-55588
```
JeremyMcCormick: You could change this to `main` as that ticket branch on `dax_ppdb` has been merged.

dspeck1: Updated.

PR #6815, opened 2026-08-04
DM-55679: Add upload arq worker to qserv-kafka and upload configuration

stvoutsin:
This PR is part of the work to enable moving the part of the job upload request to the Qserv API in qserv-kafka to a separate path so that subsequent queries are not blocked.
 
## Changes
- Adds templates for an upload worker in qserv-kafka
- Add new configuration specific to uploads in qserv-kafka values (`uploadWorkerMaxJobs` and `uploadWorkerMaxJobs`)

### DIFF
--- a/applications/qserv-kafka/Chart.yaml
+++ b/applications/qserv-kafka/Chart.yaml
@@ -5,7 +5,7 @@ version: 1.0.0
 description: "Qserv Kafka bridge"
 sources:
   - "https://github.com/lsst-sqre/qserv-kafka"
-appVersion: 5.0.0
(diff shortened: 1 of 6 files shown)


rra (dismissed):
This looks good. I would be somewhat inclined to delete the autoscaling stuff since I don't think there's any way this worker is ever going to hit 75% of the CPU given that it's just waiting for network calls, and that makes everything a bit simpler rather than carrying templates for something that's off by default and that we'll probably never turn on and test.

rra:
Moving this PR to the new test image so that we can use it to test the new release.

PR #6832, opened 2026-08-05
DM-55722: Stand up SkyViewer for DM internal HiPS viewing

mfisherlevine:


applications/skyviewer/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v2
+appVersion: 0.1.0
+description: Sky imagery browser for private HiPS surveys
+name: skyviewer
+sources:
+- https://github.com/lsst-sqre/skyviewer
```
ctslater: Not a real url.

docs/applications/skyviewer/values.md
```diff
@@ -0,0 +1,37 @@
+# skyviewer
+
+Sky imagery browser for private HiPS surveys
+
+## Source Code
+
+* <https://github.com/lsst-sqre/skyviewer>
```
ctslater: doesn't exist.

applications/skyviewer/secrets.yaml
```diff
@@ -0,0 +1,23 @@
...
+    not use CMS preview, so the value only has to be present, but generating
+    it rather than fixing it keeps those routes closed.
+  generate:
+    type: password
+"astro-objects-api-token":
+  description: >-
+    Token for the astro objects API used by the search panel. The value the
+    browser actually sends is compiled into the client bundle when the image
+    is built, because it is a NEXT_PUBLIC_ variable; this copy exists only to
+    satisfy the boot-time environment check, so it need not match.
```
ctslater: > this copy exists only to satisfy the boot-time environment check, so it need not match

Why is there a check then?

applications/skyviewer/secrets.yaml
```diff
@@ -0,0 +1,23 @@
...
+    unauthenticated queries, and the value is only required because the
+    application validates its whole environment when the server boots, so a
+    generated value is sufficient.
+  generate:
+    type: password
+"craft-revalidate-secret-token":
+  description: >-
+    Token gating the CMS preview and revalidate routes. This deployment does
+    not use CMS preview, so the value only has to be present, but generating
+    it rather than fixing it keeps those routes closed.
```
ctslater: > keeps those routes closed

I don't understand what that means.

applications/skyviewer/Chart.yaml
mfisherlevine: Fixed — now `https://github.com/lsst-dm/skyviewer-client`, which is where the image is built from. This was the `phalanx application create` starter default and I missed it.

docs/applications/skyviewer/values.md
mfisherlevine: Same cause as the Chart.yaml one: this file is generated by helm-docs from `sources`, so it repeated the bad URL. Corrected and regenerated.

applications/skyviewer/secrets.yaml
mfisherlevine: Agreed, that was opaque — I described the effect rather than the thing. Reworded to say what the token is actually for.

The app exposes `/api/preview` and `/api/revalidate`, which let a CMS editor view unpublished content. Both compare a token supplied by the caller against this value and reject the request if it does not match. This deployment does not do CMS previewing, so nothing needs to call them, and generating the value means nobody knows it and so nobody can.

applications/skyviewer/secrets.yaml
mfisherlevine: Fair question, and the description was doing a bad job of explaining it. The check is in the application, not in this chart, and it is arguably a wart.

`NEXT_PUBLIC_*` variables in Next.js are substituted into the JavaScript bundle when the image is built, not read at runtime — so the token the browser actually sends to the astro objects API is baked into the image. But the app validates its *entire* environment with `@t3-oss/env-nextjs` when the server boots, and that schema covers client variables too, so the server refuses to start unless the variable is set even though it never reads it for anything.

Hence: a value must be present, and its content is irrelevant. I chose to generate one rather than store a copy of the real token, since a stored copy would look authoritative while being able to silently drift from whatever is compiled into the image.

The real fix is app-side — exempt client variables from the boot-time check, since their values were fixed at build time. I own that fork ([lsst-dm/skyviewer-client](https://github.com/lsst-dm/skyviewer-client)) so I can do it; happy to, if you would rather this secret not exist at all.

applications/skyviewer/templates/deployment.yaml
```diff
@@ -0,0 +1,147 @@
...
+kind: Deployment
+metadata:
+  name: "skyviewer"
+  labels:
+    {{- include "skyviewer.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  # at one replica, RollingUpdate's maxUnavailable floors to zero, so the old
+  # pod is kept until the new one passes readiness — and a rollout whose new
+  # pod never comes up wedges with two unready pods that Argo CD cannot
```
ctslater: This doesn't make sense to me; why does this create "two unready pods"? The old one is still working fine.

applications/skyviewer/templates/deployment.yaml
```diff
@@ -0,0 +1,147 @@
...
+      containers:
+        - name: {{ .Chart.Name }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          ports:
+            - name: "http"
+              containerPort: 8080
+              protocol: "TCP"
+          env:
+            # the app validates its environment when the server boots, not
```
ctslater: Isn't this just saying that the correct env vars must be set, like any other app?

applications/skyviewer/values.yaml
```diff
@@ -0,0 +1,126 @@
...
+  # so without the right group every read is denied even with the volumes
+  # mounted.
+  supplementalGroups:
+    - 4085
+
+  # -- Claims to create and mount. Each entry needs `name`,
+  # `storageClassName`, `capacity` and `mountPath`. `config.hipsDataDir` must
+  # resolve below one of them, following symlinks.
+  mounts:
+    - name: "sdf-group-rubin"
```
ctslater: This is all deployment-specific and doesn't belong in the default values.yaml.

applications/skyviewer/values.yaml
```diff
@@ -0,0 +1,126 @@
...
+  # from this tree, so without this the viewer offers surveys whose every tile
+  # 404s.
+  hipsSurvey: ""
+
+  # -- Craft CMS GraphQL API, used for page content and tours. Imagery comes
+  # from `hipsSurvey`, not from here.
+  apiUrl: "https://api.skyviewer.app/api"
+
+  # -- Astro objects API, used by the search panel.
+  astroApiUrl: "https://us-central1-edc-prod-eef0.cloudfunctions.net/astro-objects-api"
```
ctslater: Deployment-specific, should be blank in default yaml.

applications/skyviewer/values.yaml
```diff
@@ -0,0 +1,126 @@
...
+  annotations: {}
+
+  # -- Path the app is served under. Must match the NEXT_PUBLIC_BASE_PATH the
+  # image was built with: Next.js compiles its base path into the client
+  # bundle, so the two cannot be changed independently.
+  path: "/skyviewer"
+
+config:
+  # -- Directory holding the HiPS surveys. Must sit below `hipsData.mountPath`.
+  hipsDataDir: "/sdf/group/rubin/shared/hips_views"
```
ctslater: Deployment-specific.

applications/skyviewer/values.yaml
```diff
@@ -0,0 +1,126 @@
...
+  # -- Whether to mount the HiPS volumes. Without them the viewer has no
+  # imagery.
+  enabled: true
+
+  # -- Supplemental group IDs the pod runs with. The data exports are
+  # group-readable only (/sdf/data/rubin/shared is drwxr-s---, group 4085),
+  # so without the right group every read is denied even with the volumes
+  # mounted.
+  supplementalGroups:
+    - 4085
```
ctslater: Deployment specific.

ctslater (dismissed):
Few things to fix first.

mfisherlevine:
The replies above are written by Claude acting on my, behalf and posted from my account - I asked it to handle this round because it actually understands this stuff, and I don't, whatsoever (as you know, Colin!). The commits are pushed and the reasoning comes from Claude - let me know if any of it doesn't look right.

Claude says: All six comments addressed in `b783372` and `e00b58d`. Rendering `usdfdev` is unchanged by both commits apart from the two comments themselves — I diffed `helm template` before and after to be sure.

applications/skyviewer/templates/deployment.yaml
mfisherlevine: Thanks — fixed in e00b58d.

applications/skyviewer/templates/deployment.yaml
mfisherlevine: Agreed, that was true of any app. Narrowed in e00b58d to the part that is not: the `NEXT_PUBLIC_` values were substituted into the client bundle when the image was built, so what is set here cannot change what the browser does — but the server still validates them at boot, so they have to be present and have to match what the image was built with. That is why `ingress.path` and the image's `NEXT_PUBLIC_BASE_PATH` cannot be changed independently.

applications/skyviewer/values.yaml
mfisherlevine: Moved to `values-usdfdev.yaml` in b783372, along with the other three you flagged.

Blanking it moved the failure mode: an unset value used to reach the app as an empty environment variable and fail its boot-time check, i.e. a crash-loop rather than a render error. So the template now wraps this one and the two API URLs in `required`, which fails at render and names the missing setting.

applications/skyviewer/values.yaml
mfisherlevine: Moved in b783372.

For the record, these two are the only ones of the four I would have argued about — they are the same public endpoints in any environment, which is what I took a chart default to be for. But they are also not something a new deployment should inherit silently, so no objection; they are now set in `values-usdfdev.yaml` and `required` in the template.

applications/skyviewer/values.yaml
mfisherlevine: Moved in b783372 — `supplementalGroups` now defaults to empty, with the 4085 and the reason for it in `values-usdfdev.yaml`.

applications/skyviewer/values.yaml
mfisherlevine: Moved in b783372, and this one was pure duplication — the same two claims were already in `values-usdfdev.yaml`, so the default is now an empty list.

`hipsData.enabled` also defaults to `false` to go with it: with blank mounts, defaulting to `true` would have produced a pod that sets `HIPS_DATA_DIR` with nothing mounted under it. An environment that wants imagery now opts in and supplies the directory, the mounts and the group.

PR #6833, opened 2026-08-06
OSW-2773: Update nightlydigest dev deployments to use new CI develop image.

sebastian-aranda:


Vebop (approved):
Will you add a note about rubincr & the security/uid to the ticket? Thanks for doing this!

sebastian-aranda:
> Will you add a note about rubincr & the security/uid to the ticket? Thanks for doing this!

Good catch!! yes I will :ok_hand: .

PR #6834, opened 2026-08-06
DM-55728: Create log explorer app

mfisherlevine:


environments/templates/applications/rubin/log-explorer.yaml
```diff
@@ -0,0 +1,42 @@
...
+        - name: "global.host"
+          value: {{ .Values.fqdn | quote }}
+        - name: "global.repertoireUrl"
+          value: "https://{{ .Values.fqdn }}/repertoire"
+        - name: "global.vaultSecretsPath"
+          value: {{ .Values.vaultPathPrefix | quote }}
+      valueFiles:
+        - "values.yaml"
+        - "values-{{ .Values.name }}.yaml"
+{{- end -}}
```
sebastian-aranda, line 42:
```suggestion
{{- end -}}

```

applications/log-explorer/values.yaml
```diff
@@ -0,0 +1,162 @@
...
+
+  # -- Loki password in the clear, for bringing an environment up before
+  # its Vault secret exists. **Do not commit a real value**: this repository
+  # is public, and anything set here also shows in plain text in `kubectl
+  # describe pod` and the Argo CD UI. Pass it as a Helm parameter override
+  # on the Argo CD Application instead, and clear it once the `loki-password`
+  # Vault secret is populated. Empty (the normal state) reads the password
+  # from that secret, and the pod starts either way — with neither, log
+  # fetches fail with a legible error and the rest of the app works.
+  lokiPassword: ""
```
sebastian-aranda, line 53: This should be removed as we already added the loki-password to vault.

applications/log-explorer/values.yaml
```diff
@@ -0,0 +1,162 @@
...
+  # fetching.
+  # 300 keeps the Loki query load modest; an exposure becomes viewable
+  # at most ~5 minutes after its shutter+5min ideal.
+  livePollS: 300
+
+  # -- How far behind "now", in seconds, each live poll stops. Covers Loki
+  # ingestion lag: a log line stamped t can be ingested seconds after t,
+  # and a poll that raced it would miss it. Raise this if the end-of-night
+  # verification pass reports pods needing refetches.
+  liveLagS: 60
```
sebastian-aranda, line 86: these could be moved into an object, e.g.

```
config:
    shutter:
        windowBefore: 5
        windowAfter: 300
        livePoll: 300
        liveLag: 60
```

applications/log-explorer/templates/deployment.yaml
```diff
@@ -0,0 +1,157 @@
...
+              value: {{ .Values.config.liveLagS | quote }}
+            - name: "LOKI_USERNAME"
+              value: {{ .Values.config.lokiUsername | quote }}
+            # The application takes no configuration from the browser at
+            # all, so this is the only way it can authenticate: there is no
+            # credentials panel to fall back on.
+            - name: "LOKI_PASSWORD"
+              {{- if .Values.config.lokiPassword }}
+              value: {{ .Values.config.lokiPassword | quote }}
+              {{- else }}
```
sebastian-aranda, line 86: I think this needs to be removed as we added the loki-password to the vault 💡 .

applications/log-explorer/tests/loki_credentials_test.yaml
sebastian-aranda, line 1: Does this require updates if we remove the "interim password"?

sebastian-aranda (dismissed):
Very very nice 👏 this is great work @mfisherlevine ! I learned very nice tricks by reviewing it 😁.

I left some minor comments and doubts. But I have an additional question that I'll state in this review comment: I see your tests `configuration_test.yaml`, `deployment_shape.yaml`, `ingress_test.yaml` and `loki_credentials.yaml` use `values-base.yaml`, is that on purpose? Don't we require to also test `values-summit.yaml`?

environments/templates/applications/rubin/log-explorer.yaml
mfisherlevine: Thanks, fixed.

applications/log-explorer/values.yaml
mfisherlevine: Fixed. Worth noting it wasn't a stand-in for the Vault secret, but a way to bring a new environment up before its secret exists - except `optional: true` on the `secretKeyRef` already covers that: the pod starts, and the app answers the first fetch with `LOKI_PASSWORD is not set.` So the literal only bought fetching before `secrets sync`, at the price of a credential rendering into kubectl describe pod and the Argo CD UI from a public repo. Gone now 👍

applications/log-explorer/templates/deployment.yaml
mfisherlevine: Thanks, fixed.

applications/log-explorer/values.yaml
mfisherlevine: Pushing back slightly: `windowBeforeS`/`windowAfterS` are shutter-relative, but `livePollS`/`liveLagS` are the Loki poll cadence and its ingestion lag, which are nothing to do with the shutter, so `config.shutter` would misname two of the four. Happy to do `config.window.{beforeS,afterS}` + `config.live.{pollS,lagS}` if you'd still like them nested though. What do you think?

applications/log-explorer/tests/loki_credentials_test.yaml
mfisherlevine: Yes, done. Dropped the literal-password case, and added one asserting `LOKI_PASSWORD` carries no `value` at all, so a plaintext path can't come back unnoticed.

mfisherlevine:
> Very very nice 👏 this is great work @mfisherlevine ! I learned very nice tricks by reviewing it 😁.
> 
> I left some minor comments and doubts. But I have an additional question that I'll state in this review comment: I see your tests `configuration_test.yaml`, `deployment_shape.yaml`, `ingress_test.yaml` and `loki_credentials.yaml` use `values-base.yaml`, is that on purpose? Don't we require to also test `values-summit.yaml`?

Yes, deliberate. The two files only differ in the `config.site block`, and `site_catalog_test.yaml` already covers that from both ends - base in one case, summit in the next. The other suites don't assert anything that depends on the site, so pointing them at `values-summit` would just render the same manifests and check the same things twice.

It did matter for credentials though, so I've added a summit case there. And if you'd want a guard against the two files drifting apart in anything other than the site block, that'd beat duplicating the suites - happy to add it.

applications/log-explorer/values.yaml
sebastian-aranda: got it, I think when looking at this the "S" suffix on each parameter made me wonder they were inside the same "Shutter" scope so made sense to group them in that case. I think you can leave it as it is, the docstrings should be enough :+1: .

sebastian-aranda:
> > Very very nice 👏 this is great work @mfisherlevine ! I learned very nice tricks by reviewing it 😁.
> > I left some minor comments and doubts. But I have an additional question that I'll state in this review comment: I see your tests `configuration_test.yaml`, `deployment_shape.yaml`, `ingress_test.yaml` and `loki_credentials.yaml` use `values-base.yaml`, is that on purpose? Don't we require to also test `values-summit.yaml`?
> 
> Yes, deliberate. The two files only differ in the `config.site block`, and `site_catalog_test.yaml` already covers that from both ends - base in one case, summit in the next. The other suites don't assert anything that depends on the site, so pointing them at `values-summit` would just render the same manifests and check the same things twice.
> 
> It did matter for credentials though, so I've added a summit case there. And if you'd want a guard against the two files drifting apart in anything other than the site block, that'd beat duplicating the suites - happy to add it.

Noted! thanks for the clear answer. I guess you don't expect the two files to drift apart really, so probably no need to duplicate the suits. Thank you @mfisherlevine !

sebastian-aranda (approved):

PR #6856, opened 2026-08-11
DM-55778: Change BigQuery TAP to new PPDB Dev GCP Project

dspeck1:
Set Big Query project to new PPDB project deployed via Terraform.


stvoutsin (approved):
Looks correct to me!

PR #6864, opened 2026-08-13
OSW-2819: Update nginx image for nightlydigest and love due to CVE-2026-42533.

sebastian-aranda:
This PR also removes pod-antiaffinity from the love nginx deployment configurations as it was preventing the pod from being synced up.


aibsen (approved):
looks good to me

PR #6866, opened 2026-08-13
Update tap schemaVersion to releases/EDP2-deploy-v3b

JeremyMcCormick:


gpdf (approved):
This is what's intended for deployment, subject to review in-place on data-int.

PR #6867, opened 2026-08-13
Update datalink URLs to EDP2-deploy-v3b

JeremyMcCormick:
Restores consistency after #6866 but should have no effect (DataLink config was not changed between -v2 and -v3b).


gpdf (approved):
Thank you, this looks correct.

PR #6885, opened 2026-08-18
DM-55823: ppdb-replication - Deploy USDF int services

fajpunk:
**Changes**

- Update `ppdbConfig` URLs for `usdfdev` and `usdfint` environments
- Add `cloudSql` config section to `usdfint`
- Change `dbName` in `values.yaml` to `ppdb` to match new naming convention

### DIFF
--- a/applications/ppdb-replication/values-usdfdev.yaml
+++ b/applications/ppdb-replication/values-usdfdev.yaml
@@ -11,7 +11,7 @@ config:
   # -- APDB config file resource
   apdbConfig: "/sdf/data/rubin/user/jeremym/ppdb-config/apdb_dev.yaml"
 
-    # -- CloudSQL Python Connector configuration
+  # -- CloudSQL Python Connector configuration
   cloudSql:
 
     # -- CloudSQL Connection Name
@@ -19,6 +19,9 @@ config:
 
     # -- Username in sa_name@project.iam format
     user: "usdf-replication@ppdb-dev-5c07.iam"
+
+    # -- Database name to connect to
+    dbName: "ppdb-chunk-tracking"
 
   # -- PPDB config file resource
   ppdbConfig: "gs://ppdb-dev-config/ppdb-dev-5c07_config.yaml"
--- a/applications/ppdb-replication/values-usdfdev.yaml
+++ b/applications/ppdb-replication/values-usdfdev.yaml
@@ -20,9 +20,6 @@ config:
     # -- Username in sa_name@project.iam format
     user: "usdf-replication@ppdb-dev-5c07.iam"
 
-    # -- Database name to connect to
-    dbName: "ppdb-chunk-tracking"
-
   # -- PPDB config file resource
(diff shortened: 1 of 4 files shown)


applications/ppdb-replication/values.yaml
```diff
@@ -123,4 +123,4 @@ config:
     user: ""
 
     # -- Database name to connect to
-    dbName: "ppdb-chunk-tracking"
+    dbName: "ppdb"
```
JeremyMcCormick, line 126: As discussed, this is a good change across all environments, but we should keep in mind that it will effect `dev` once merged, which hasn't been changed yet. So we'll need to update that database name and and the configuration file or the replication will break. (Blocked behind changing in `idf_deploy` and then propagating to `dev`.)

applications/ppdb-replication/values-usdfint.yaml
```diff
@@ -7,6 +7,15 @@ config:
...
+
+    # -- CloudSQL Connection Name
+    connectionName: "ppdb-int-6c62:us-central1:ppdb-int"
+
+    # -- Username in sa_name@project.iam format
+    user: "usdf-replication@ppdb-int-6c62.iam"
+
   # -- PPDB config file resource
   # TODO: Use a group rather than user path for these configurations
   ppdbConfig: "/sdf/data/rubin/user/jeremym/dev/DM-53599/ppdb-config/ppdb_lsstcam.yaml"
```
JeremyMcCormick: This is a new config for the int environment:

```suggestion
  ppdbConfig: "gs://ppdb-int-config/ppdb-int_config.yaml"
```

JeremyMcCormick (dismissed):
This shouldn't be merged until the config is updated as the existing one will not work.

```
ppdbConfig: "/sdf/data/rubin/user/jeremym/dev/DM-53599/ppdb-config/ppdb_lsstcam.yaml"
```

I will let you know when I have done this.

We also need to deal with db renaming of `ppdb-chunk-tracking` -> `ppdb`, which hasn't been done for `dev` yet.

applications/ppdb-replication/values.yaml
fajpunk: Good point, I initially thought that we would have re-provisioned the dev CloudSQL DB by the time we merged this PR, but we will not have done that. I've overridden the db name in dev so that it will stay as `ppdb-chunk-tracking`

applications/ppdb-replication/values-usdfdev.yaml
```diff
@@ -20,6 +20,9 @@ config:
     # -- Username in sa_name@project.iam format
     user: "usdf-replication@ppdb-dev-5c07.iam"
 
+    # -- Database name to connect to
+    dbName: "ppdb-chunk-tracking"
```
JeremyMcCormick: This `dbName` can be deleted.

fajpunk:
LGTM!

JeremyMcCormick (approved):

PR #6903, opened 2026-08-21
DM-55895: Enable ppdbtap in repertoire for idfint to enable move to CloudSQL tap_schema

stvoutsin:
Changes:
- Make `cloudsql` the default tap_schema type for `ppdbtap` and set the appropriate configuration for it
- Enable `ppdbtap` in repertoire for `idfint`
- Re-order tap.servers for idfint & idfdev
- Fix cadc-tap chart bug in deployment with duplicate `tapuser` `maxActive` setting

### DIFF
--- a/charts/cadc-tap/templates/tap-deployment.yaml
+++ b/charts/cadc-tap/templates/tap-deployment.yaml
@@ -125,7 +125,6 @@ spec:
                 -Dtapschemauser.url=jdbc:postgresql://localhost:5432/{{ .Values.tapSchema.database }}
                 {{- end }}
                 -Dtapschemauser.maxActive={{ .Values.tapSchema.tapuser.maxActive }}
-                -Dtapuser.maxActive={{ .Values.tapSchema.tapuser.maxActive }}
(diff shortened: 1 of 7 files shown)


JeremyMcCormick:
@stvoutsin We should pin `int` to a weekly like `w.2026.33` for `ppdbtap` in this PR, so that it is not getting the base schema version from the DP2 tag. It needs to sit on an older version of the PPDB schema until there are some updates to the `int` cloud infrastructure.

charts/cadc-tap/templates/tap-deployment.yaml
```diff
@@ -125,7 +125,6 @@ spec:
                 -Dtapschemauser.url=jdbc:postgresql://localhost:5432/{{ .Values.tapSchema.database }}
                 {{- end }}
                 -Dtapschemauser.maxActive={{ .Values.tapSchema.tapuser.maxActive }}
-                -Dtapuser.maxActive={{ .Values.tapSchema.tapuser.maxActive }}
```
JeremyMcCormick, line 128: This looks like a change unrelated to the ticket. Maybe note in the description if you want to include it on this PR?

PR #6908, opened 2026-08-21
Move USDF qservs from IPs to FQDNs

fritzm:


athornton (approved):
Confirmed that the names do indeed map to the IPv4 addresses they replace.

I'll merge.

PR #6913, opened 2026-08-24
DM-55906: Stand up SkyViewer on prod RSP

mfisherlevine:


applications/skyviewer/templates/deployment.yaml
```diff
@@ -42,12 +42,14 @@ spec:
               containerPort: 8080
               protocol: "TCP"
           env:
+            # read at runtime, which is what lets one image serve more than
```
ctslater: More claude comments that make sense speaking to you, but aren't relevant to someone reading this file.

ctslater (dismissed):

PR #6928, opened 2026-08-26
DM-55960: Allow kafka brokers and controllers to share the same kubernetes nodes in Sasquatch

afausti:
Allow brokers and controllers to share the same Kubernetes node reducing the minimum number of nodes required to run Sasquatch from 6 to 3.

This is the configuration we had for the Roundtable environments and this PR makes it the default in Sasquatch.

Tested on idfdev and seems to be working fine.


fajpunk (approved):
I thought we only wanted to change this in `idfdev`? If we discussed and we're OK with doing this everywhere, then this LGTM

afausti:
In case of a single node failure we lose one broker and one controller and the Kafka cluster can still run. So I think it is fine to adopt this as the default configuration everywhere.

afausti:
@rra in case you want to review this before I merge - thanks!

afausti:
@aibsen this also changes the Summit Kafka deployment but I think it is fine.

rra (approved):
Looks good to me!

PR #6944, opened 2026-08-31
OSW-2868: Remove anti-affinity rules for CSCs in envsys, love and simonyitel and add global topology spread constraints.

aibsen:


rbovill (approved):
LGtM

PR #6945, opened 2026-08-31
tickets/OSW 2839

couger01:
- **Add couger01 to developer role at usdf prod.**
- **Setup ingress authentication config for weatherbroadcaster.**


aibsen (approved):
looks good to me

PR #6953, opened 2026-09-02
DM-55024: Add Sasquatch GafaelfawrServiceToken

afausti:
ObsForge needs to connect to the Sasquatch Schema Registry at USDF RSP clusters. 
The Sasquatch Schema Registry is protected by a Gafaelfawr ingress and requires a Gafaelfawr token with `write:sasquatch` scope.
Deploy a `GafaelfawrServiceToken` resource in Sasquatch to request a service token with the right scope.


dspeck1 (approved):
LGTM

PR #6976, opened 2026-09-09
DM-56046: Add HPA best-practices

fajpunk:
TIL from the robots about `.. admonition`. If there is a better way to do this, please let me know!

### DIFF
--- a/docs/developers/helm-chart/autoscaling.rst
+++ b/docs/developers/helm-chart/autoscaling.rst
@@ -1,0 +1,92 @@
+###########
+Autoscaling
+###########
+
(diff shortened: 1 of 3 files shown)


docs/developers/helm-chart/autoscaling.rst
```diff
@@ -0,0 +1,89 @@
...
+
+      apiVersion: autoscaling/v2
+      kind: HorizontalPodAutoscaler
+      spec:
+        behavior:
+        metrics:
+        - resource:
+            name: cpu
+            target:
+              averageValue: 75
```
rra: You said `750m` above but here it's 75. If this is correct, could you add a note about why they're different?

rra (dismissed):
One minor note, otherwise looks good!

PR #6978, opened 2026-09-09
DM-56050: *-kafka -  Fix bridge worker resource and autoscaling config

fajpunk:
We want scale-up at 750m average CPU usage among all of the pods. We’re currently configured to scale up at 75% of 50m CPU usage.

There are a lot of other places in Phalanx where we should probably make this change (including the starter template), but we can do that in future PRs.

### DIFF
--- a/applications/bigquery-kafka/values.yaml
+++ b/applications/bigquery-kafka/values.yaml
@@ -170,7 +170,7 @@ frontend:
   # @default -- See `values.yaml`
   resources:
     limits:
-      cpu: "2"
+      cpu: "1"
       memory: "1Gi"
     requests:
       cpu: "100m"
@@ -205,8 +205,8 @@ fastWorker:
     # limit.
     maxReplicas: 10
 
-    # -- Target CPU utilization of bigquery-kafka fast worker pods.
-    targetCPUUtilizationPercentage: 75
+    # -- Target absolute CPU usage value of bigquery-kafka fast worker pods.
+    targetCPUQuantity: 750m
 
   # -- Node selection rules for the bigquery-kafka fast worker pods
   nodeSelector: {}
@@ -221,10 +221,10 @@ fastWorker:
   # @default -- See `values.yaml`
   resources:
     limits:
-      cpu: "250m"
+      cpu: "1"
       memory: "300Mi"
     requests:
-      cpu: "50m"
+      cpu: "1"
       memory: "145Mi"
 
   # -- Tolerations for the bigquery-kafka fast worker pods
@@ -256,8 +256,8 @@ slowWorker:
     # connection limit.
     maxReplicas: 10
 
-    # -- Target CPU utilization of bigquery-kafka slow worker pods.
-    targetCPUUtilizationPercentage: 75
+    # -- Target absolute CPU usage value of bigquery-kafka slow worker pods.
+    targetCPUQuantity: 750m
 
   # -- Node selection rules for the bigquery-kafka slow worker pods
   nodeSelector: {}
@@ -272,7 +272,7 @@ slowWorker:
   # @default -- See `values.yaml`
   resources:
     limits:
-      cpu: "2"
+      cpu: "1"
       memory: "1Gi"
     requests:
       cpu: "1"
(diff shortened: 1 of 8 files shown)


applications/qserv-kafka/values.yaml
```diff
@@ -271,8 +271,8 @@ slowWorker:
     # limit.
     maxReplicas: 10
 
-    # -- Target CPU utilization of qserv-kafka slow worker pods.
-    targetCPUUtilizationPercentage: 50
```
stvoutsin, line 275: Don't remember exactly why we had 50 here instead of 75, but just flagging in case that was deliberate as we are now changing it to match the others at 750m.

applications/bigquery-kafka/values.yaml
```diff
@@ -221,10 +221,10 @@ fastWorker:
   # @default -- See `values.yaml`
   resources:
     limits:
-      cpu: "250m"
+      cpu: "1"
       memory: "300Mi"
     requests:
-      cpu: "50m"
```
stvoutsin, line 227: Since we are now switching to average value my understanding is that the request value isn't part of the calculation right? In which case are we sure that increasing here to guarantee reservation of 1 core, and do we know if this changes scale-up if for example the node-pool can't supply the cores (is this even likely?)

applications/qserv-kafka/templates/fast-worker-hpa.yaml
```diff
@@ -14,14 +14,12 @@ spec:
   maxReplicas: {{ .Values.fastWorker.autoscaling.maxReplicas }}
   minReplicas: {{ .Values.fastWorker.autoscaling.minReplicas }}
   metrics:
-    {{- if .Values.fastWorker.autoscaling.targetCPUUtilizationPercentage }}
     - type: Resource
       resource:
         name: "cpu"
         target:
-          type: Utilization
-          averageUtilization: {{ .Values.fastWorker.autoscaling.targetCPUUtilizationPercentage }}
```
stvoutsin, line 23: I think doing this is correct for the fast worker, but because a lot of what it does is I/O bound I'm also not certain that we won't see memory blowing up before we hit the CPU trigger now that we're concentrating the load to fewer fast pods. Perhaps this is worth further testing during our scale tests just to confirm that a different scale-up strategy isn't needed alongside.

stvoutsin (commented):
This all looks good to me as far as I can see, just a couple of general questions/comments.

applications/qserv-kafka/values.yaml
fajpunk: Good catch, this should explicitly be `500m` per https://github.com/lsst-sqre/phalanx/commit/f03c725ec19c3a5a68f545a62a935a39

applications/bigquery-kafka/values.yaml
fajpunk: This is a good question. I don't think we have to worry about the node pool not being able to supply the cores. The node pool will scale up to satisfy the requests, until we hit the max number of nodes at least. We may end up paying for more CPU than we need though.

If we don't set the request to `1`, we could be risking never actually get to `750m` of usage, and thus never scaling up. I don't know how likely this would be in practice. When we were scaling based on a fraction of the request, it was guaranteed that it would be possible to reach that threshold, because Kubernetes is guaranteed to give us all of our request.

applications/qserv-kafka/templates/fast-worker-hpa.yaml
fajpunk: Yeah, definitely worth a scale test. If we need to, we can scale on both CPU _and_ memory. I think if we do the streaming memory usage calculations we were talking about though, we could confidently set the memory request so that we can't blow it.

applications/bigquery-kafka/values.yaml
rra: This logic works for me. I think we have tons of unallocated CPU in general in the cluster right now. We can always revisit if we see it blocking scaledown.

applications/qserv-kafka/templates/fast-worker-hpa.yaml
rra: I'm hoping that the new version with streaming uploads won't have a problem with memory usage in the fast worker, but we should scale test with a bunch of user table uploads and make sure. We should probably increase the request to 200MiB and the limit to 400MiB based on what we're currently seeing in prod, though.

rra (dismissed):

PR #6979, opened 2026-09-09
DM-56052: Fix Sasquatch prompt processing topic names

afausti:
All Sasquatch topics should start with the `lsst` prefix.


dspeck1 (approved):
LGTM.  Added commit to remove `lsst.prompt.next-visit-job` since we only need one test topic now.

PR #6988, opened 2026-09-11
DM-56075: Better isolation of Sasquatch REST Proxy and remote Schema Registry

afausti:
We notice that requests to the local Schema Registry in Sasquatch registered schemas into the remote Schema Registry instead. Configuration looks correct and a restart of the Schema Registryand the REST Proxy deployments fixed the problem. This PR tries to better isolate the REST Proxy from the remote  Schema Registry.


dspeck1 (dismissed):
LGTM...nice catch on enableServiceLinks.

PR #7009, opened 2026-09-16
RSO-902: Create phalanx application support running evening tailgate bot on BTS.

tribeiro:


applications/so-evening-tailgate/values-base.yaml
aibsen: the variables under envSecrets and env aren't referenced in deployment.yaml template, so I think  they won't be in the pod

applications/so-evening-tailgate/templates/deployment.yaml
aibsen: I think you need to add here under env:
{{- range .Values.env }}
            - name: {{ .name }}
              value: {{ .value | quote }}
{{- end }}
same for envFrom, so the values in values-base are picked up

applications/so-evening-tailgate/templates/vault-secrets.yaml
aibsen: I think you should remove the 'if' if you want only slack-webhook to be conditional. in the secrets.yaml that seems to be the only conditional secret

aibsen (dismissed):
looks good to me

aibsen (approved):
lgtm

PR #7012, opened 2026-09-16
DM-56085: Update Repertoire to 2.2.0

rra:
The new release adds support for more service discovery attributes, which will be used to improve integration with Squareone, RSP documentation, and the SIA service.


tests/data/input/environments/values-usdfdev-prompt-processing.yaml
```diff
@@ -1,5 +1,6 @@
 name: usdfdev-prompt-processing
 fqdn: usdf-prompt-processing-dev.slac.stanford.edu
+description: "Test environment"
```
athornton: This looks like an errant paste; I think it should be "Development environment for prompt processing at USDF" or some such?

pyproject.toml
```diff
@@ -22,7 +22,7 @@ classifiers = [
 requires-python = ">=3.13"
 dependencies = [
     "bcrypt",
-    "click!=8.2.0",
+    "click>8.2.0",
```
athornton, line 25: Sure, since they're now past the (somehow-broken) 8.2.0, but there's not a way to tell pip to use the minimum version of each package that will solve the constraints...

...and https://github.com/pypa/pip/issues/8085 omg it's full of tentacles.

athornton (dismissed):
One thing (usdf prompt processing dev description) that I think is an actual error, other than that LGTM.

tests/data/input/environments/values-usdfdev-prompt-processing.yaml
rra: Easy to miss because you have to look at the path but this is in the test suite, so I just added anything to satisfy the schema (it's not actually used by the tests).

athornton (approved):

PR #7014, opened 2026-09-17
DM-56064: Update dp2 Butler on data-int to "full" DP2

dhirving:
Update Butler server and SIA on data-int to point to a close approximation of the "full" DP2 release.  See https://rubinobs.atlassian.net/wiki/x/CgCveg for more info.


dhirving:
@stvoutsin I'd like to test the SIA change on data-int tomorrow when you're around.  I think with all the data loaded it is like 5x as many queries over 10-20x as much data, so we'll have to see if things work OK.  The ObsCore export at USDF (which I think uses the same queries?) is not moving very fast so there may be some issues.

stvoutsin (approved):
This looks good to me!
I've done some manual testing to check for the new types and it all seems to be working fine as far as I can see, so I'd say this is good to merge.

PR #7016, opened 2026-09-17
DM-55950: Change usdfint TAP services to Postgres backend for UWS & TAP_SCHEMA

stvoutsin:
Configures all tap applications for `usdfint` to use external postgres for the UWS and TAP_SCHEMA databases.


JeremyMcCormick:
Can you also update `schemaVersion` for `consdbtap` to `"w.2026.36"` in `values-usdfdev.yaml` so it is similar to `int`?

